PR #2, opened 2019-07-16
prime repo

deads2k:
adds makefile, glide.yaml, and bump to get started.


squeed:
cc @danwinship

deads2k:
We plan to scale back CI to unit tests, verifies, and building since no images from here are used yet. Then we'll add images and CI for them back as we go.

sttts:
/hold

until we have disable publishing, and CI + image is near to be in place

images/network-controller/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM registry.svc.ci.openshift.org/openshift/builder:golang-1.12 AS builder
+WORKDIR /go/src/github.com/openshift/sdn
+COPY . .
+RUN make build --warn-undefined-variables WHAT=cmd/network-controller
```
tnozicka: WHAT isn't supported
I think there is a separate target for every binary hopefully, try `make help`

images/network-controller/Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+
+FROM centos:7
+COPY --from=builder /go/src/github.com/openshift/sdn/network-controller /usr/bin/
+ENTRYPOINT ["/usr/bin/network-controller"]
+LABEL io.k8s.display-name="OpenShift Network Controller" \
+      io.k8s.description="This is a component of OpenShift, it controls the network" \
+      com.redhat.component="network-controller" \
+      io.openshift.tags="openshift,network-controller,sdn"
```
tnozicka: line end

tnozicka:
I'll fix the CI to take correct paths for those Dockerfiles

images/network-controller/Dockerfile
sttts: k

images/network-controller/Dockerfile
sttts: k

Makefile
```diff
@@ -14,7 +14,8 @@ include $(addprefix ./vendor/github.com/openshift/library-go/alpha-build-machine
 # $2 - Dockerfile path
 # $3 - context directory for image build
 # It will generate target "image-$(1)" for builing the image an binding it as a prerequisite to target "images".
-$(call build-image,origin-$(GO_PACKAGE),./images/sdn/Dockerfile,.)
+$(call build-image,origin-sdn,./images/sdn/Dockerfile,.)
```
tnozicka: reflect the real names here pls so we avoid confusion

sttts: done

sttts:
Waiting for network-controller to be published.

Makefile
```diff
@@ -0,0 +1,26 @@
...
+)
+
+# This will call a macro called "build-image" which will generate image specific targets based on the parameters:
+# $0 - macro name
+# $1 - target suffix
+# $2 - Dockerfile path
+# $3 - context directory for image build
+# It will generate target "image-$(1)" for builing the image an binding it as a prerequisite to target "images".
+$(call build-image,origin-node,./images/sdn/Dockerfile,.)
+$(call build-image,origin-network-controller,./images/network-controller/Dockerfile,.)
```
danwinship: Neither of these image names is really right; the sdn node image is currently called "origin-node", but that's a historical accident. (The SDN binary is the only thing left in the old `openshift start node` image.) It should be renamed to "origin-sdn" or "origin-openshift-sdn" while we're moving it.

For the controller side, "origin-network-controller" makes it sound generically network-related, but it's not; it's very specifically the controller for just the openshift-sdn network plugin. (It doesn't get run if you have any other network plugin configured.) So it should be "origin-sdn-controller" or "origin-openshift-sdn-controller".

danwinship: (comments about "network controller" vs "sdn controller" apply to the labels inside the image too, though those are easier to fix later without needing to coordinate multiple repos).

Also, you could just merge them into a single image if that made sense...

deads2k: -node is legacy, we aren't planning to change that as part of this push.


I suspect as a post-step, you'll want a single unified sdn image, but I think that can happen just as easily afterwards.  This is about seeing if we can get fast enough responses on images to make your CI a little more effective.

sttts: I agree with the rename of the node image. But let's leave that for a follow-up. We cannot safe the world in one PR. Better step by step.

@stlaz see network-controller name.

@danwinship is that a definitive answer for the network-controller image name?

images/node/Dockerfile
```diff
@@ -0,0 +1,39 @@
+FROM registry.svc.ci.openshift.org/openshift/builder:golang-1.12 AS builder
+WORKDIR /go/src/github.com/openshift/sdn
+COPY . .
+RUN go build "vendor/github.com/containernetworking/plugins/plugins/ipam/host-local"
+RUN go build "vendor/github.com/containernetworking/plugins/plugins/main/loopback"
+RUN mkdir -p /tmp/build; \
+    cp /go/src/github.com/openshift/origin/_output/local/bin/linux/$(go env GOARCH)/openshift-sdn /tmp/build/openshift-sdn; \
```
tnozicka: this won't match

images/node/Dockerfile
```diff
@@ -0,0 +1,39 @@
+FROM registry.svc.ci.openshift.org/openshift/builder:golang-1.12 AS builder
+WORKDIR /go/src/github.com/openshift/sdn
+COPY . .
+RUN go build "vendor/github.com/containernetworking/plugins/plugins/ipam/host-local"
```
tnozicka: do `make build` here

Makefile
tnozicka: we need to change it in CI as that's where  the definitive name for CI resides. "origin-sdn-controller" or "origin-openshift-sdn-controller"?

sttts: from Slack discussion with @danwinship: sdn-controller (with automatic ose/origin prefix)

images/node/Dockerfile
sttts: fixed.

images/node/Dockerfile
tnozicka: @sttts this break the build

tnozicka:
```
COPY --from=builder /go/src/github.com/openshift/sdn/openshift-sdn /usr/bin/
```

tnozicka: or pick this file https://github.com/openshift/origin/pull/23273/files#diff-29391d0364ab2ca3cb45701b081dcf3aR1

PR #4, opened 2019-07-18
Fix Makefile and .gitignore

tnozicka:
and most importantly this gives as our first promoted image on merge

/cc @sttts @deads2k


tnozicka:
/hold cancel
relevant CI already passed

PR #10, opened 2019-07-22
switch to SDN only fork of k/k

deads2k:
The rest of the kube deps use the real upstream values.  Removed one outdated patch that we no longer need because it's not part of a union hypershift binary.


deads2k:
And a link to your branch here; https://github.com/openshift/kubernetes/commits/sdn-4.2-kubernetes-1.14.0  follow the same pattern for other branches.

PR #11, opened 2019-07-22
Untaint node on startup if node is tainted and MTU is ok

rcarrillocruz:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1728518

### DIFF
--- a/pkg/network/node/node.go
+++ b/pkg/network/node/node.go
@@ -311,38 +311,48 @@ func (node *OsdnNode) validateMTU() error {
 		return fmt.Errorf("unable to determine MTU while performing validation")
 	}
 
-	if int(node.networkInfo.MTU) > mtu+50 {
+	needsTaint := mtu < int(node.networkInfo.MTU)+50
+	const MTUTaintKey string = "network.openshift.io/mtu-too-small"
+	mtuTooSmallTaint := &corev1.Taint{Key: MTUTaintKey, Value: "value", Effect: "NoSchedule"}
+	nodeObj, err := node.kClient.CoreV1().Nodes().Get(node.hostName, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("could not get Kubernetes Node object by hostname: %v", err)
+	}
+	tainted := taints.TaintExists(nodeObj.Spec.Taints, mtuTooSmallTaint)
+	if needsTaint != tainted {
 		resultErr := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
-			klog.V(2).Infof("Configured node MTU is more than default interface MTU plus VXLAN overhead, tainting node...")
-			n, err := node.kClient.CoreV1().Nodes().Get(node.hostName, metav1.GetOptions{})
-
+			nodeObj, err = node.kClient.CoreV1().Nodes().Get(node.hostName, metav1.GetOptions{})
 			if err != nil {
 				return fmt.Errorf("could not get Kubernetes Node object by hostname: %v", err)
 			}
+			nodeObjCopy := nodeObj.DeepCopy()
+			var nodeWithTaintMods *corev1.Node
 
-			oldNode, err := json.Marshal(n)
+			if needsTaint && !tainted {
+				klog.V(2).Infof("Default interface MTU is less than VXLAN overhead, tainting node...")
+				nodeWithTaintMods, _, err = taints.AddOrUpdateTaint(nodeObjCopy, mtuTooSmallTaint)
+				if err != nil {
+					return fmt.Errorf("could not taint the node with key %s: %v", MTUTaintKey, err)
+				}
+			} else if !needsTaint && tainted {
+				klog.V(2).Infof("Node has too small MTU taint but default interface MTU is big enough, untainting node...")
+				nodeWithTaintMods, _, err = taints.RemoveTaint(nodeObjCopy, mtuTooSmallTaint)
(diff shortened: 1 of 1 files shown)


pkg/network/node/node.go
```diff
@@ -355,6 +355,44 @@ func (node *OsdnNode) validateMTU() error {
...
+			nodeClone := n.DeepCopy()
+
+			untaintedNode, _, err := taints.RemoveTaint(nodeClone, &corev1.Taint{Key: MTUTaintKey, Value: "value", Effect: "NoSchedule"})
+
+			if err != nil {
+				return fmt.Errorf("could not untaint the node with key %s: %v", MTUTaintKey, err)
+			}
+
+			newNode, err := json.Marshal(untaintedNode)
+
```
danwinship: As a random style note, golang code (or maybe just kubernetes code?) tends to not use so many blank lines. In particular, you should never put a blank line between setting `err` and checking it.

pkg/network/node/node.go
```diff
@@ -311,38 +311,36 @@ func (node *OsdnNode) validateMTU() error {
...
-	if int(node.networkInfo.MTU) > mtu+50 {
+	const MTUTaintKey string = "network.openshift.io/mtu-too-small"
+	n, err := node.kClient.CoreV1().Nodes().Get(node.hostName, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("could not get Kubernetes Node object by hostname: %v", err)
+	}
+	mtuTooSmallTaint := &corev1.Taint{Key: MTUTaintKey, Value: "value", Effect: "NoSchedule"}
+	tainted := taints.TaintExists(n.Spec.Taints, mtuTooSmallTaint)
+	mtuWithOverhead := mtu + 50
+	if !tainted && int(node.networkInfo.MTU) > mtuWithOverhead {
```
danwinship: Sigh. This check was wrong to begin with. Ricky, you're excused because you're new to openshift-sdn, but @squeed and I both suck for not catching it in the original PR.

`node.networkInfo.MTU` is the VXLAN MTU. `mtu` is the actual MTU of the node's eth0. In order for the VXLAN to work, the node's MTU must be at least 50 bytes larger than the VXLAN MTU. So the correct check is:

    needsTaint := mtu < int(node.NetworkInfo.MTU) + 50

(As opposed to the current code which adds 50 to the wrong side, and mis-handles the `==` case.)

QE didn't catch this because their tests were focusing more on the 9001-vs-1500 case, and as it happens, 8951 is greater than either 1450 or 1550.

pkg/network/node/node.go
```diff
@@ -311,38 +311,36 @@ func (node *OsdnNode) validateMTU() error {
...
 		resultErr := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
 			klog.V(2).Infof("Configured node MTU is more than default interface MTU plus VXLAN overhead, tainting node...")
```
danwinship: (This correctly describes what the old code was checking, but will need to be updated now.)

pkg/network/node/node.go
```diff
@@ -311,38 +311,36 @@ func (node *OsdnNode) validateMTU() error {
...
-			n, err := node.kClient.CoreV1().Nodes().Get(node.hostName, metav1.GetOptions{})
```
danwinship: Oh, actually you need to keep the `Get` in the loop even if you already did it outside the loop. The point of `RetryOnConflict` is that if two people try to update at once, the first one will succeed, and that will change the `resourceVersion` on the object, and then when the the apiserver tries to process the second update, it will see that the `resourceVersion` you passed in is old and so return `409 Conflict` and force you to try again so that you don't accidentally undo the changes that the first person made. So anyway, this means that if you get a conflict, you need to call `Get` again to get the newly-updated version of the resource before trying to update it again.

tl;dr: keep the `Get`. And add it to the "remove" case.

Actually, you could merge a lot of the "add" and "remove" cases together by just testing if `tainted != needsTaint` first, and if so, then do the `RetryOnConflict` loop, and then inside that, figure out whether you need to be calling `AddOrUpdateTaint` or `RemoveTaint`, and everything else would be the same.

pkg/network/node/node.go
rcarrillocruz: sigh, gotcha...

squeed: damn damn damn. I stared at that for a while and convinced myself it was right, too.

pkg/network/node/node.go
```diff
@@ -311,38 +311,44 @@ func (node *OsdnNode) validateMTU() error {
...
+			var nodeWithTaintMods *corev1.Node
+
+			if tainted {
+				klog.V(2).Infof("Default interface MTU is less than VXLAN overhead, tainting node...")
+				nodeWithTaintMods, _, err = taints.AddOrUpdateTaint(nodeObjCopy, mtuTooSmallTaint)
+				if err != nil {
+					return fmt.Errorf("could not taint the node with key %s: %v", MTUTaintKey, err)
+				}
+			} else {
+				klog.V(2).Infof("Node has too small MTU taint but default interface MTU is big enough, untainting node...")
```
squeed: This is behind the `if needsTaint {` conditional, so I don't think it will actually untaint the node.. The logic is more like
```
if needsTaint && !tainted {
    AddTaint()
} else if !needsTaint && tainted {
    RemoveTaint()
}
```

PR #12, opened 2019-07-23
Allow DNS port when performing iptables filtering on cloud provider metadata IP

alexanderConstantinescu:
This addresses the bug described [here](https://jira.coreos.com/browse/SDN-493)

We want to allow DNS packets to be routed through the metadata IP as this is sometimes used by internal mechanism to these providers when performing certain networking tasks.

### DIFF
--- a/cmd/sdn-cni-plugin/openshift-sdn_linux.go
+++ b/cmd/sdn-cni-plugin/openshift-sdn_linux.go
@@ -126,9 +126,11 @@ var iptablesCommands = [][]string{
 	{"-A", "FORWARD", "-p", "tcp", "-m", "tcp", "--dport", "22623", "-j", "REJECT"},
 	{"-A", "FORWARD", "-p", "tcp", "-m", "tcp", "--dport", "22624", "-j", "REJECT"},
 
-	// Block cloud metadata IP
-	{"-A", "OUTPUT", "-d", "169.254.169.254", "-j", "REJECT"},
-	{"-A", "FORWARD", "-d", "169.254.169.254", "-j", "REJECT"},
+	// Block cloud provider metadata IP except DNS
+	{"-A", "OUTPUT", "-p", "tcp", "-m", "tcp", "-d", "169.254.169.254", "!", "--dport", "53", "-j", "REJECT"},
+	{"-A", "OUTPUT", "-p", "udp", "-m", "udp", "-d", "169.254.169.254", "!", "--dport", "53", "-j", "REJECT"},
+	{"-A", "FORWARD", "-p", "tcp", "-m", "tcp", "-d", "169.254.169.254", "!", "--dport", "53", "-j", "REJECT"},
+	{"-A", "FORWARD", "-p", "udp", "-m", "udp", "-d", "169.254.169.254", "!", "--dport", "53", "-j", "REJECT"},
 }
 
 func (p *cniPlugin) CmdAdd(args *skel.CmdArgs) error {
--- a/pkg/network/node/ovscontroller.go
+++ b/pkg/network/node/ovscontroller.go
@@ -160,9 +160,6 @@ func (oc *ovsController) SetupOVS(clusterNetworkCIDR []string, serviceNetworkCID
 		otx.AddFlow("table=30, priority=100, ip, nw_dst=%s, actions=goto_table:90", clusterCIDR)
 	}
 
-	// Link-local traffic
-	otx.AddFlow("table=30, priority=75, ip, nw_dst=169.254.0.0/16, actions=drop")
-
 	// Multicast coming from the VXLAN
 	otx.AddFlow("table=30, priority=50, in_port=1, ip, nw_dst=224.0.0.0/4, actions=goto_table:120")
 	// Multicast coming from local pods


cmd/sdn-cni-plugin/openshift-sdn_linux.go
```diff
@@ -126,9 +126,11 @@ var iptablesCommands = [][]string{
 	{"-A", "FORWARD", "-p", "tcp", "-m", "tcp", "--dport", "22623", "-j", "REJECT"},
 	{"-A", "FORWARD", "-p", "tcp", "-m", "tcp", "--dport", "22624", "-j", "REJECT"},
 
-	// Block cloud metadata IP
-	{"-A", "OUTPUT", "-d", "169.254.169.254", "-j", "REJECT"},
-	{"-A", "FORWARD", "-d", "169.254.169.254", "-j", "REJECT"},
+	// Block cloud metadata IP on all ports except 53 for tcp/udp
+	{"-p", "tcp", "-A", "OUTPUT", "-d", "169.254.169.254", "!", "--dport", "53", "-j", "REJECT"},
```
squeed: It's convention to put the "-A OUTPUT" first.

alexanderConstantinescu: Thanks, just updated the PR

squeed:
cc @danwinship

danwinship:
I think maybe it would be better to only block port 80/443 rather than to block everything except 53.

Also, we're blocking it in two places: see `pkg/network/node/ovscontroller.go` as well.

However, do we actually know that the metadata service on GCP contains privileged information that we need to prevent pods from accessing? Maybe the fix is "only block on AWS", not "only block http".

squeed:
> I think maybe it would be better to only block port 80/443 rather than to block everything except 53.
I would prefer whitelisting over blacklisting.

> However, do we actually know that the metadata service on GCP contains privileged information that we need to prevent pods from accessing? Maybe the fix is "only block on AWS", not "only block http".

The [GKE documentation](https://cloud.google.com/kubernetes-engine/docs/how-to/protecting-cluster-metadata) seems to imply that, yes, the cluster metadata endpoint can contain sensitive information.

alexanderConstantinescu:
I agree with @squeed concerning the IPs to block. 

I will update the PR with modifications to the additional file you mentioned @danwinship

alexanderConstantinescu:
Just updated the PR to drop packets to 443 and 80 only. 

Please have a look and let me know

cmd/sdn-cni-plugin/openshift-sdn_linux.go
```diff
@@ -126,9 +126,11 @@ var iptablesCommands = [][]string{
...
 	{"-A", "FORWARD", "-p", "tcp", "-m", "tcp", "--dport", "22624", "-j", "REJECT"},
 
-	// Block cloud metadata IP
-	{"-A", "OUTPUT", "-d", "169.254.169.254", "-j", "REJECT"},
-	{"-A", "FORWARD", "-d", "169.254.169.254", "-j", "REJECT"},
+	// Block cloud metadata IP on ports 80 & 443 for TCP & UDP
+	{"-A", "OUTPUT", "-p", "tcp", "-m", "tcp", "-d", "169.254.169.254", "--dport", "80", "-j", "REJECT"},
+	{"-A", "OUTPUT", "-p", "udp", "-m", "udp", "-d", "169.254.169.254", "--dport", "80", "-j", "REJECT"},
+	{"-A", "FORWARD", "-p", "tcp", "-m", "tcp", "-d", "169.254.169.254", "--dport", "443", "-j", "REJECT"},
+	{"-A", "FORWARD", "-p", "udp", "-m", "udp", "-d", "169.254.169.254", "--dport", "443", "-j", "REJECT"},
```
danwinship: This got messed up in one of the rewrites... now it's blocking port 80 in `OUTPUT` and port 443 in `FORWARD`.

Also, we don't need to block UDP, just TCP. Also, you can simplify using `multiport`. So (I think, not tested):

    {"-A", "OUTPUT", "-d", "169.254.169.254", "-p", "tcp", "-m", "multiport", "--destination-ports", "80,443", "-j", "REJECT"},
    {"-A", "FORWARD", "-d", "169.254.169.254", "-p", "tcp", "-m", "multiport", "--destination-ports", "80,443", "-j", "REJECT"},

but also, @squeed, what's the verdict on 8080?

pkg/network/node/ovscontroller.go
```diff
@@ -161,7 +161,10 @@ func (oc *ovsController) SetupOVS(clusterNetworkCIDR []string, serviceNetworkCID
 	}
 
 	// Link-local traffic
-	otx.AddFlow("table=30, priority=75, ip, nw_dst=169.254.0.0/16, actions=drop")
+	otx.AddFlow("table=30, priority=75, ip, tcp, tcp_dst=80, nw_dst=169.254.0.0/16, actions=drop")
+	otx.AddFlow("table=30, priority=75, ip, tcp, tcp_dst=443, nw_dst=169.254.0.0/16, actions=drop")
+	otx.AddFlow("table=30, priority=75, ip, udp, udp_dst=80, nw_dst=169.254.0.0/16, actions=drop")
+	otx.AddFlow("table=30, priority=75, ip, udp, udp_dst=443, nw_dst=169.254.0.0/16, actions=drop")
```
danwinship: Don't need UDP.
And also, at this point, if we're giving up on the idea of blocking all link-local traffic out of the pod network then this should match `nw_dst=169.254.169.254`, not `nw_dst=169.254.0.0/16`

squeed: Wait, flows are ordered, right? So can we allow packets with a port (src or dst) of 53, and drop all others?

cmd/sdn-cni-plugin/openshift-sdn_linux.go
squeed: Waiting for the west coast to have coffee, then we'll see.

pkg/network/node/ovscontroller.go
danwinship: Well, you can't just generically "allow"; you have to say specifically what to do with the packet next (in this case, `goto_table:100`). But... then you'd have two sets of `goto_table:100` rules with a `drop` rule between them and that kind of violates the code style? Handwave handwave.

Yes, we could do it.

squeed:
To unblock GCP, I've asked @alexanderConstantinescu to add the 53 whitelist to iptables and drop the openflow rule completely for now. I'll file a blocker bug to re-add the openflow rules.

cmd/sdn-cni-plugin/openshift-sdn_linux.go
```diff
@@ -126,9 +126,11 @@ var iptablesCommands = [][]string{
 	{"-A", "FORWARD", "-p", "tcp", "-m", "tcp", "--dport", "22623", "-j", "REJECT"},
 	{"-A", "FORWARD", "-p", "tcp", "-m", "tcp", "--dport", "22624", "-j", "REJECT"},
 
-	// Block cloud metadata IP
-	{"-A", "OUTPUT", "-d", "169.254.169.254", "-j", "REJECT"},
-	{"-A", "FORWARD", "-d", "169.254.169.254", "-j", "REJECT"},
+	// Block cloud metadata IP on ports 80 & 443 for TCP & UDP
```
squeed: "Block cloud provider metadata IP except DNS"

squeed:
@danwinship, would you be OK with removing the hold for now?

danwinship:
/hold cancel
sure but make sure the blocker bug involves figuring out azure too

PR #13, opened 2019-07-24
Bug 1726045: skip OPENSHIFT-MASQ for traffic already marked for masquerade

squeed:
If a packet has already been marked by other kube-proxy rules for masquerade, don't run it through the OPENSHIFT-MASQUERADE chain for further twiddling.

Most notably, this chain is used for Egress IPs.

This change fixes a bug where egress IPs can't access services via their ExternalIP. (bz [1726045](https://bugzilla.redhat.com/show_bug.cgi?id=1726045))


squeed:
@danwinship you already LGTM'd this - but now it's in the correct repo.

PR #8, opened 2019-07-19
Unidling: minimize iptables lock contention 

squeed:
This PR includes

- backport of upstream PR 71735, which moves the userspace proxy over to an asynchronous model
- Changes needed to use that backport
- A small carry patch that lets us tweak the proxy internals slightly
 - A patch that combines both proxy's sync loops, preventing lock contention

Fixes: BZ 1699341

(This is the 4.2 version of https://github.com/openshift/origin/pull/23426)

PR #17, opened 2019-07-26
iptables: partial #13 revert, skip masq in chain

squeed:
PR #13 changed the source rule for OPENSHIFT-MASQUERADE. However, rolling out source-rule changes on a running system isn't clean. Easier to leave the source-rule unchanged, and just add a rule first in the chain that gets us where we want to be.
    
So, -j RETURN for traffic that's already marked for MASQ.

### DIFF
--- a/pkg/network/node/iptables.go
+++ b/pkg/network/node/iptables.go
@@ -179,7 +179,12 @@ func (n *NodeIPTables) getNodeIPTablesChains() []Chain {
 		},
 	)
 
-	var masqRules [][]string
+	masqRules := [][]string{
+		// Skip traffic already marked by kube-proxy for masquerading.
+		// This fixes a bug where traffic destined to a service's ExternalIP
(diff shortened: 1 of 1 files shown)


squeed:
@danwinship PTAL.

danwinship:
> bz1726045 / PR #8

you mean #13 

But the whole reason why created our own separate chains in the first place was so that we didn't have to do complicated cleanup (https://github.com/openshift/origin/pull/13465). We should revert that part of #13 and do this differently. Eg, if OPENSHIFT-MASQUERADE started with a "if marked then -j RETURN" rule, that would have the same effect, while leaving the toplevel rule unchanged from 4.1.

squeed:
> if OPENSHIFT-MASQUERADE started with a "if marked then -j RETURN" rule

You're right, that's a much cleverer solution.

squeed:
OK, updated (and fixed the commit message / title)

PR #19, opened 2019-07-31
Bug 1700431: Pass egress IP packets to conntrack

danwinship:
If a packet is being sent over the VXLAN to a remote egress IP, we have to tell conntrack to remember the packet first, so that when we get the response it will be recognized as such (and thus accepted regardless of NetworkPolicy).

(Note that we can't just pass *all* outbound packets to conntrack; calling `ct(commit)` on a packet will make iptables think it has already processed the packet, so we can't commit anything that will need to be masqueraded, DNAT'ed, etc, later.)


danwinship:
/retest
though it's possible that the change really did totally break things; I only tested a little bit locally

danwinship:
@dcbw @squeed PTAL

squeed:
/lgtm
Does this need to be backported to 4.1?

danwinship:
probably 4.1 and 3.11

PR #20, opened 2019-08-02
Bug 1740741: Initializing runningPods on SDN bootup for 4.x

alexanderConstantinescu:
This is a fix for [BUG 1723924](https://bugzilla.redhat.com/show_bug.cgi?id=1723924)

FIX for 4.X on origin, see linked PR's in origin

4.1:

https://github.com/openshift/origin/pull/23537

3.11:

https://github.com/openshift/origin/pull/23522

The problem:

As mentioned in the link above, when the openshift networking process dies, the "podManager" object and its attribute "runningPods" is wiped for memory, which can lead to stateful inconsistencies 
with openflow ports concerning pods from a multicast-enabled namespaces 

The fix:

We now initialize the "podManager" with the initial state of all running pods when the process is booted, as such we will have a synchronized state.

### DIFF
--- a/pkg/network/node/node.go
+++ b/pkg/network/node/node.go
@@ -396,7 +396,7 @@ func (node *OsdnNode) Start() error {
 		return fmt.Errorf("failed to set up iptables: %v", err)
 	}
 
-	networkChanged, existingPods, err := node.SetupSDN()
+	networkChanged, existingOFPodNetworks, err := node.SetupSDN()
 	if err != nil {
 		return fmt.Errorf("node SDN setup failed: %v", err)
 	}
@@ -419,14 +419,23 @@ func (node *OsdnNode) Start() error {
 		node.watchServices()
 	}
 
+	existingPodSandboxes, err := node.getPodSandboxes()
+	if err != nil {
+		return err
+	}
+
+	if err = node.podManager.InitRunningPods(existingPodSandboxes, existingOFPodNetworks); err != nil {
+		return err
+	}
+
 	klog.V(2).Infof("Starting openshift-sdn pod manager")
 	if err := node.podManager.Start(cniserver.CNIServerRunDir, node.localSubnetCIDR,
 		node.networkInfo.ClusterNetworks, node.networkInfo.ServiceNetwork.String()); err != nil {
 		return err
 	}
 
-	if networkChanged && len(existingPods) > 0 {
-		err := node.reattachPods(existingPods)
+	if networkChanged && len(existingOFPodNetworks) > 0 {
+		err := node.reattachPods(existingPodSandboxes, existingOFPodNetworks)
 		if err != nil {
 			return err
 		}
@@ -447,15 +456,12 @@ func (node *OsdnNode) Start() error {
 // reattachPods takes an array containing the information about pods that had been
 // attached to the OVS bridge before restart, and either reattaches or kills each of the
 // corresponding pods.
-func (node *OsdnNode) reattachPods(existingPods map[string]podNetworkInfo) error {
-	sandboxes, err := node.getPodSandboxes()
-	if err != nil {
-		return err
-	}
+func (node *OsdnNode) reattachPods(existingPodSandboxes map[string]*kruntimeapi.PodSandbox, existingOFPodNetworks map[string]podNetworkInfo) error {
 
 	failed := []*kruntimeapi.PodSandbox{}
-	for sandboxID, podInfo := range existingPods {
-		sandbox, ok := sandboxes[sandboxID]
+	for sandboxID, podInfo := range existingOFPodNetworks {
+		sandbox, ok := existingPodSandboxes[sandboxID]
+
 		if !ok {
 			klog.V(5).Infof("Sandbox for pod with IP %s no longer exists", podInfo.ip)
 			continue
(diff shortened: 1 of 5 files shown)


squeed:
This needs some eyes. @JacobTanenbaum, can you also take a look?

danwinship:
/retitle Bug 1723924: Initializing runningPods on SDN bootup for 4.x

alexanderConstantinescu:
/retitle Bug 1740741: Initializing runningPods on SDN bootup for 4.x

pkg/network/node/ovscontroller.go
```diff
@@ -261,9 +262,16 @@ func (oc *ovsController) GetPodNetworkInfo() (map[string]podNetworkInfo, error)
 			continue
 		}
 
+		ofport, err := oc.ovs.GetOFPort(row["name"])
```
danwinship: you don't need to do this; you should be able to just add `"ofport`" to the list of fields passed to `oc.ovs.Find` above

pkg/network/node/node.go
```diff
@@ -425,8 +425,17 @@ func (node *OsdnNode) Start() error {
...
 	}
 
-	if networkChanged && len(existingPods) > 0 {
-		err := node.reattachPods(existingPods)
+	existingPodSandboxes, err := node.getPodSandboxes()
+	if err != nil {
+		return err
+	}
+
+	if err = node.podManager.InitRunningPods(existingPodSandboxes, existingOFPodNetworks); err != nil {
```
danwinship: might be better to put this a little bit above, before the call to `podManager.Start()`?

pkg/network/node/pod.go
```diff
@@ -160,13 +161,39 @@ func (m *podManager) Start(rundir string, localSubnetCIDR string, clusterNetwork
 	return m.cniServer.Start(m.handleCNIRequest)
 }
 
+func (m *podManager) InitRunningPods(existingPodSandboxes map[string]*kruntimeapi.PodSandbox, existingOFPodNetworks map[string]podNetworkInfo) error {
+	m.runningPodsLock.Lock()
+	defer m.runningPodsLock.Unlock()
+	defer glog.V(5).Infof("Finished initializing podManager with running pods at start-up")
```
danwinship: Don't use `defer`. Just log it at the end.

pkg/network/node/pod.go
```diff
@@ -160,13 +161,39 @@ func (m *podManager) Start(rundir string, localSubnetCIDR string, clusterNetwork
...
+
+	for _, sandbox := range existingPodSandboxes {
+
```
danwinship: (nitpicky, but kill the extra blank line here)

pkg/network/node/pod.go
```diff
@@ -160,13 +161,39 @@ func (m *podManager) Start(rundir string, localSubnetCIDR string, clusterNetwork
...
+		sKey := getPodKey(sandbox.Metadata.Namespace, sandbox.Metadata.Name)
+
+		vnid, err := m.policy.GetVNID(sandbox.Metadata.Namespace)
+		if err != nil {
+			return err
```
danwinship: this should probably be a log-and-continue, not a return

pkg/network/node/pod.go
```diff
@@ -160,13 +161,39 @@ func (m *podManager) Start(rundir string, localSubnetCIDR string, clusterNetwork
...
+		}
+
+		podNetworkInfo, ok := existingOFPodNetworks[sandbox.Id]
+		if !ok {
+			glog.Errorf("Inexistant openflow network for sandbox pod with ID: %s", sandbox.Id)
```
danwinship: "Nonexistent". Also "openflow network" isn't really the right term. I'd say something like `"No network information for pod %s/%s", sandbox.Metadata.Namespace, sandbox.Metadata.Name`. Also, warning not error.

pkg/network/node/pod.go
alexanderConstantinescu: Any reason in particular for this? Performance? I see we are using defer when unlocking the runningPodsLock at every invocation elsewhere in the code. Should I modify those too in that case?

pkg/network/node/pod.go
alexanderConstantinescu: If this happens, we have a sandbox pod without any `podNetworkInfo`, meaning we won't be able initialize `runningPods` correctly, which will probably lead to errors (possibly subtle and hard-to-debug ones). Are you sure that this should be logged as a warning solely?

pkg/network/node/pod.go
```diff
@@ -15,13 +15,14 @@ import (
...
 	"github.com/openshift/sdn/pkg/network/node/cniserver"
-	"k8s.io/klog"
-
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/klog"
+	"k8s.io/klog/glog"
```
squeed: nit: Try to avoid using glog

PR #23, opened 2019-08-15
proxy: add a bit more logging around proxy type and sync status

squeed:
QE pointed out that it's not easy to see if changing the unidling setting has taken effect. So, add a few simple log lines.

### DIFF
--- a/pkg/network/proxyimpl/hybrid/proxy.go
+++ b/pkg/network/proxyimpl/hybrid/proxy.go
@@ -328,12 +328,13 @@ func (p *HybridProxier) Sync() {
 // We do this so that we can guarantee that changes are applied to both
 // proxies, especially when unidling a newly-awoken service.
 func (p *HybridProxier) syncProxyRules() {
-	klog.V(4).Infof("hybrid proxy: syncProxyRules start")
+	klog.V(2).Infof("hybrid proxy: syncProxyRules start")
 
 	p.mainProxy.SyncProxyRules()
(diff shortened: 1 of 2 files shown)


pkg/network/proxyimpl/hybrid/proxy.go
```diff
@@ -328,12 +328,13 @@ func (p *HybridProxier) Sync() {
...
 func (p *HybridProxier) syncProxyRules() {
-	klog.V(4).Infof("hybrid proxy: syncProxyRules start")
+	klog.V(2).Infof("hybrid proxy: syncProxyRules start")
 
 	p.mainProxy.SyncProxyRules()
+	klog.V(2).Infof("hybrid proxy: mainProxy.syncProxyRules complete")
 	p.unidlingProxy.SyncProxyRules()
 
-	klog.V(4).Infof("hybrid proxy: syncProxyRules finished")
+	klog.V(2).Infof("hybrid proxy: syncProxyRules finished")
```
danwinship: Maybe make this consistent with the other new message: `"hybrid proxy: unidlingProxy.SyncProxyRules complete"`. (oh, and capitalize "Sync" in the previous message if you want to be accurate)

squeed: Making `syncProxyRules` exported is a silly carry-patch, hence leaving it lower-case. But the other wording change makes sense.

squeed:
updated, ptal

PR #24, opened 2019-08-16
Stop building CNI loopback and host-local plugins

danwinship:
With https://github.com/openshift/cluster-network-operator/pull/288 merged, openshift-sdn gets its upstream CNI plugins from the container-networking-plugins image, so we can stop building them.

/cc @squeed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,5 @@
 /openshift-sdn
 /sdn-cni-plugin
 /network-controller
-/loopback
-/host-local
 /kube-proxy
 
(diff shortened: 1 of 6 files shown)


squeed:
Something didn't work:

danwinship:
oh, right, I was assuming that was because the CNO half had *just* merged and not been rebuilt yet, but then I forgot to come back later to say
/retest

danwinship:
@squeed: updated. I don't love hardcoding the path in there like that...

squeed:
Oh huh, that's interesting - so we're executing host-local from the sdn-process, not the CNI plugin.

If you don't like hard-coding the path (which is a bit ugly), the right solution would be to set CNI_PATH in the daemonset.

Honestly, we should just vendor the logic in and stop with the binary protocol altogether, someday.

PR #26, opened 2019-08-20
Add support for IPBlocks

Jaanki:
Add support for IPBlocks in network policy.

ipBlocks in NetworkPolicies allow us to specify CIDR ranges
as ingress sources or egress destinations. Generally these
are cluster-external IPs and will allow you to filter incoming
packets based on the source-ip.

This PR implements ingress ipBlocks in NetworkPolicies.
Support for ipBlocks with except rules will be added via separate PR.

Signed-off-by: Sridhar Gaddam <sgaddam@redhat.com>
Signed-off-by: Janki Chhatbar <jchhatba@redhat.com>

### DIFF
--- a/pkg/network/node/networkpolicy.go
+++ b/pkg/network/node/networkpolicy.go
@@ -525,6 +525,16 @@ func (np *networkPolicyPlugin) parseNetworkPolicy(npns *npNamespace, policy *net
 				npp.watchesPods = true
 				peerFlows = append(peerFlows, np.selectPodsFromNamespaces(peer.NamespaceSelector, peer.PodSelector)...)
 			}
+
+			if peer.IPBlock != nil {
+				if peer.IPBlock.Except != nil {
+					// Currently IPBlocks with except rules are skipped.
+					klog.Warningf("IPBlocks with except rules are not supported (NetworkPolicy [%s], Namespace [%s])", policy.Name, policy.Namespace)
+				} else {
+					// Network Policy has ipBlocks, allow traffic from those ips.
+					peerFlows = append(peerFlows, fmt.Sprintf("ip, nw_src=%s, ", peer.IPBlock.CIDR))
+				}
+			}
 		}
 		for _, destFlow := range destFlows {
 			for _, peerFlow := range peerFlows {


pkg/network/node/networkpolicy.go
```diff
@@ -456,6 +456,13 @@ func (np *networkPolicyPlugin) parseNetworkPolicy(npns *npNamespace, policy *net
 					npp.watchesNamespaces = true
 					peerFlows = append(peerFlows, np.selectNamespaces(peer.NamespaceSelector)...)
 				}
+			} else if peer.IPBlock != nil {
+				// As per https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/networking/types.go#L181
+				if peer.NamespaceSelector != nil || peer.PodSelector != nil {
+					return nil, fmt.Errorf("PodSelector or NamespaceSeletor can not be used along with IPBlocks")
```
danwinship: You don't need to check this. The upstream validation code will already have ensured that

pkg/network/node/networkpolicy.go
```diff
@@ -456,6 +456,13 @@ func (np *networkPolicyPlugin) parseNetworkPolicy(npns *npNamespace, policy *net
...
+				}
+				// TODO: Add logic for "Except" block
+				peerFlows = append(peerFlows, fmt.Sprintf("Flow rule with CIDR =%s successfully added", peer.IPBlock.CIDR))
```
danwinship: obviously this needs to be changed to real openflow...

Jaanki: @danwinship thanks for your reviews. I didnt quite understand your comment. Can you please explain :)

danwinship: peerFlows is an array of openflow expressions to match traffic against. That is where all the real work of implementing NetworkPolicy happens. You can't just *say* "Flow rule successfully added", you have to actually write the correct flow rule.

pkg/network/node/networkpolicy.go
```diff
@@ -227,6 +227,11 @@ func (np *networkPolicyPlugin) syncFlows() {
 			npns.dirty = false
 		}
 	}
+
+	//TODO: find correct type of peer
+	if peer
```
Jaanki: { missing which I will rectify in the next patch

sridhargaddam:
@danwinship , @dcbw appreciate if you can take a look at this patch and review it. Thanks.

pkg/network/node/networkpolicy.go
```diff
@@ -238,6 +239,11 @@ func (np *networkPolicyPlugin) syncNamespaceFlows(npns *npNamespace) {
 
 		// Add "allow" rules for all traffic allowed by a NetworkPolicy
 		for _, npp := range npns.policies {
+			// Program the except (aka drop) flows configured in ingress/ipBlock Network Policies
+			for _, flow := range npp.dropFlows {
```
danwinship: `except` doesn't mean "drop". It just means that the excepted IPs aren't allowed by *this* rule. If you have one policy that says:

    - from:
      - ipBlock:
          cidr: 10.0.0.0/8
          except:
          - 10.0.1.0/24

and a second that says:

    - from:
      - ipBlock:
          cidr: 10.0.0.0/8

then the `except` in the first rule essentially has no effect. Or if the second rule had `cidr: 10.0.1.1/32`, then you would be allowed to connect to that, but not the rest of `10.0.1.0/24`.

So...
- You could figure out the intersection of the `cidr` and the `except`s to come up with the set of allowed CIDR blocks, and then add flows accepting each of those. But this gets really terrible really quickly. (eg, in the "`10.0.0.0/8` except for `10.0.1.0/24`" example above, you need to allow `10.0.0.0` - `10.0.0.255` and `10.0.2.0` - `10.255.255.255`, which as CIDR blocks would be: `10.0.0.0/8`, `10.0.2.0/23`, `10.0.4.0/22`, `10.0.8.0/21`, `10.0.16.0/20`, `10.0.32.0/19`, `10.0.64.0/18`, `10.0.128.0/17`, `10.1.0.0/16`, `10.2.0.0/15`, `10.4.0.0/14`, `10.8.0.0/13`, `10.16.0.0/12`, `10.32.0.0/11`, `10.64.0.0/10`, and `10.128.0.0/9`. And that's pretty terrible.
  - Though actually OVS lets you specify arbitrary bitmasks, not just CIDR prefix lengths, so there might be a simpler way to express it
- You could try to figure out some way to sort the "accept" and "deny" flows to get the right result. This seems like it would also be incredibly hard, and possibly impossible in the general case.
- Try to figure something out using multiple tables or conjunctive matches or `resubmit`
- ... ?

I'm not really sure what to do here...

mangelajo: Agreed, the drop does not work. First I thought it was a nice optimization to the problem, but it's not.

Probably we should focus on implementing the first option, then the last one is an optimization we can probably handle later if necessary, btw I'm going to check how ovn-kubernetes handles this... (answer in next comment)

sridhargaddam: Thanks @danwinship for your review comments. Yeah, the optimization does not work :( 
Let me think about it...

sridhargaddam: I had a quick look at OVN-Kubernetes. They seem to do something similar to what this patch is doing (unless I'm missing something).

OVN-kubernetes is programming a drop[1] rule with higher priority[*] for the except flows and allow[2] rules for the ipBlock CIDR.
It seem to apply this for "every" individual Network Policy[3].

[1] https://github.com/ovn-org/ovn-kubernetes/blob/f9787987e2ccbdbfc078e8adb279fb2751e97faf/go-controller/pkg/ovn/policy.go#L247
[2] https://github.com/ovn-org/ovn-kubernetes/blob/f9787987e2ccbdbfc078e8adb279fb2751e97faf/go-controller/pkg/ovn/policy.go#L261
[3] https://github.com/ovn-org/ovn-kubernetes/blob/f9787987e2ccbdbfc078e8adb279fb2751e97faf/go-controller/pkg/ovn/policy.go#L790
[*] https://github.com/ovn-org/ovn-kubernetes/blob/f9787987e2ccbdbfc078e8adb279fb2751e97faf/go-controller/pkg/ovn/policy_common.go#L178

Likewise, I also had a look at WeaveNet and it uses iptables/ipsets for Network Policies.
When we have a Network Policy with ipBlock & Except blocks, it creates the following ipset[4] rules.
It creates an ipset rule only when the Network policy has an except block, otherwise not.
Also, I see that "Reference count" for the ipsets is always 0 (maybe a bug?) and its creating a new ipSet for each Network Policy and not reusing/updating the existing ipset with new except IPs (if they fall in the same CIDR).

First Network Policy:
- from:
  - ipBlock:
      cidr: 10.0.0.0/8
      except:
      - 10.0.1.0/24

Second Network Policy:
- from:
  - ipBlock:
      cidr: 10.0.0.0/8
      except:
      - 10.0.1.1/24


[4] ipset rules:
----------------
Name: weave-@+L(xm/~]~+oyL3:3nquEz4wx
Type: hash:net
Revision: 6
Header: family inet hashsize 1024 maxelem 65536 comment
Size in memory: 481
References: 0
Number of entries: 1
Members:
10.0.1.0/24 comment "excepted from 10.0.0.0/8"

Name: weave-?:@yc!f3lX7+w^hF$%h7I^G3T
Type: hash:net
Revision: 6
Header: family inet hashsize 1024 maxelem 65536 comment
Size in memory: 481
References: 0
Number of entries: 1
Members:
10.0.1.1 comment "excepted from 10.0.0.0/8"

danwinship: Yes, ovn-kube's support looks wrong.
Weave looks right; it appears that it eventually generates an iptables rule looking like

    -s 10.0.0.0/8 -m set ! --match-set weave-xxx

meaning, "if the source is in 10.0.0.0/8 and doesn't match the ipset containing the exceptions", so that's correct; sources in the exception set won't be matched by this rule, and so will have the chance to be matched by the next rule.

danwinship: (filed https://github.com/kubernetes/kubernetes/issues/83191 about adding an e2e test for this case)

sridhargaddam: > -s 10.0.0.0/8 -m set ! --match-set weave-xxx

Yes @danwinship  the approach taken in Weave seems more practical. 
However, looks like even weave has issues.

Since weave is creating an ipSet for each NetworkPolicy that has an except block, it should ideally use multiple "-m set ! --match-set weave-xxx -m set --match-set weave-yyy"
in the iptable command.

But this does not seem to be the case looking at the code [5].
I'm just sharing this observation for info. 

It would have been great if we had the opportunity to use ipset kind of flows with OVS.

[5] https://github.com/weaveworks/weave/blob/05ab1139db615c61b99074c63184076ba72e2416/npc/ipblock.go#L47

sridhargaddam: > (filed [kubernetes/kubernetes#83191](https://github.com/kubernetes/kubernetes/issues/83191) about adding an e2e test for this case)

Cool, thank you :+1:

danwinship: > Since weave is creating an ipSet for each NetworkPolicy that has an except block

No, it's creating an ipset and an iptables rule for each `ipBlock`, not for each NetworkPolicy. A given NetworkPolicy may result in multiple iptables rules.

sridhargaddam: Oops, it was a typo and I wanted to say ipBlock. Thanks @danwinship

sridhargaddam: I was looking at conjunctive actions, sorting allow/drop flows, some other possible solutions and also had a brief discussion with @aswin (thanks) who pointed me to an approach we took in ODL.

I have the following proposal in mind and appreciate if you can take a look at this document[*] and provide your comments, thank you :)
I'll love to hear from you before I start doing any changes in the code. 

Some guidelines, I've followed:
1. If we have an ipBlock without any except rules, traffic that matches the ipBlock will be allowed right away (i.e., takes precedence over ipBlocks with except rules).

2. However, when we have an ipBlock with an except rule[s] and traffic matches the except rule, we will continue to search if there is another NetworkPolicy/ipBlock that allows such traffic.
If we are successful, we allow the traffic, otherwise drop it.

3. I have to take a closer look, but my idea is to program flows that uses multiple tables (with resubmits) ONLY when we have ipBlocks with except rules (for any matching POD in a namespace), otherwise the pipeline would be Table70 -> Table80 -> allow/drop (similar to the one we have currently)

Note: I understand that if there is an update/deletion of a NetworkPolicy (that uses ipblocks/except rules), we will have to update multiple rules :(

[*] https://docs.google.com/document/d/1RQtzt26tqCIrZLVJtdWkw4b0FSKCmQU9FXaa23VwvZI/edit?usp=sharing

danwinship: I commented on the doc, adding a suggested simplification. The general idea seems good.

sridhargaddam: Many thanks @danwinship for reviewing it and for the simplification. Appreciate it.

Can I handle this as two separate patches?
I'll update the current patch to implement ipblocks (without except rules) [*] and propose a new patch to address the ipblocks-with-except-rules.

We have a use-case in Submariner where we would program ipBlocks alone (i.e., without except) and it would really help us if we can get that feature in soon while I work on the second patch that implements ipblocks-with-except (which could take some time for implementation, testing as well as careful review).

[*] i.e.,
#The prio 150 flows from syncNamespaceFlows() for all policies except
#ipblocks-with-except happen here just like before:
table80, prio:150  nw_src: 30.0.3.0/24, Action: output:NXM_NX_REG2[]

danwinship: Yeah, that's fine. Just log a warning when you see a block with an `except` clause.

sridhargaddam: Awesome, I'll update the patch. Thanks @danwinship

sridhargaddam:
@danwinship , @dcbw, @JacobTanenbaum : I've updated the patch to support ipBlocks in NetworkPolicies. As mentioned in my earlier comment, support for ipBlocks with except rules will be added via a different patch. Appreciate your review.

aswin:
Hey everyone, I am not sure why I am getting these emails. Could you please
audit your project and remove me?

- Aswin

danwinship:
@aswin someone referred to you above; probably they meant to tag someone else with a similar name. None of us can remove you at this point but there should be a "mute the thread" link at the bottom of your email notifications that will unsubscribe you

danwinship:
Looks good. Can you squash the commits?

sridhargaddam:
> Looks good. Can you squash the commits?

Thank you @danwinship. Updated the patch.

sridhargaddam:
I see that ci/prow/e2e-aws and ci/prow/e2e-aws-upgrade are failing in other[*] PRs as well. 
Is this some known issue?

[*] https://github.com/openshift/sdn/pull/44#issuecomment-538026136

danwinship:
yes, there's some project-wide CI breakage currently

sridhargaddam:
> yes, there's some project-wide CI breakage currently

Thanks @danwinship 

Looks like the tests are now passing in the latest PR (https://github.com/openshift/sdn/pull/45), let me trigger the re-runs.

PR #27, opened 2019-08-28
Do not cache egressLink details

rcarrillocruz:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1731905

### DIFF
--- a/pkg/network/node/egressip.go
+++ b/pkg/network/node/egressip.go
@@ -4,7 +4,6 @@ package node
 
 import (
 	"fmt"
-	"net"
 	"os/exec"
 	"sync"
 	"syscall"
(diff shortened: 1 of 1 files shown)


pkg/network/node/egressip.go
```diff
@@ -185,18 +180,23 @@ func (eip *egressIPWatcher) releaseEgressIP(egressIP, mark string) error {
 		return nil
 	}
 
-	localEgressIPMaskLen, _ := eip.localEgressNet.Mask.Size()
+	localEgressLink, localEgressNet, err := GetLinkDetails(eip.localIP)
```
squeed: What about making this `GetLinkDetails(egressIP)`, so it works in all cases. Thoughts, @danwinship?

PR #32, opened 2019-09-12
how to do a Kubernetes rebase

danwinship:
It seemed like this belongs in `HACKING.md`, though the fact that we don't have anything else there yet makes it fit less well...

/assign @squeed

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -1,0 +1,178 @@
+## How to do a Kubernetes rebase
+
+First off, wait for someone else to start the origin rebase. It will
+make things easier...
+
+### Creating a new branch in our local kubernetes repo
+
+1. Grab https://github.com/openshift/kubernetes and check out a new
+   branch named
+   `sdn-${OPENSHIFT_RELEASE_NUMBER}-kubernetes-${KUBERNETES_RELEASE_NUMBER}`,
+   based on the latest upstream release/pre-release tag. eg, the first
+   rebase attempt for OCP 4.3 was the branch
+   `sdn-4.3-kubernetes-1.16.0-beta.2`, based off the upstream tag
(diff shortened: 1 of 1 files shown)


HACKING.md
```diff
@@ -0,0 +1,153 @@
...
+2. Update the remaining kubernetes and openshift dependencies in
+   `glide.yaml`. For dependencies under `github.com/openshift`, and
+   for kubernetes dependencies where we currently depend on a specific
+   commit SHA, see what version is now being used by the origin
+   rebase, and use that. (This is why you waited until after someone
+   started the origin rebase...)
+
+3. Try it out... run
+
+       glide update --strip-vendor
```
danwinship: Ah, I was thinking there should be an automated way to do this and it turns out the answer is `make update-deps`
(@rcarrillocruz )

danwinship:
Updated based on @rcarrillocruz 's experience:
- Clarified that the name of the kubernetes base branch should have the complete tag name in it. (eg, including the "`.0`" in this case)
- Clarified that there probably won't be any `<drop>` commits
- Updated with instructions on what to do if you can't push to `openshift/kubernetes`
- Clarified that we shouldn't depend on `master` when a module has no appropriate tag
- Fixed to say `make update-deps` rather than `glide update --strip-vendor` everywhere
- Added notes about possibly needing to explicitly pin new dependencies

squeed:
This looks pretty good to me. @rcarrillocruz, want to give this another pass?

danwinship:
/assign @rcarrillocruz 
It's already got /approved, so if you /lgtm it, it should merge

squeed:
/lgtm
let's get this merged

squeed:
/skip e2e-aws
/skip e2e-aws-upgrade
/skip e2e-gcp
the cloud providers don't need any more money than they already have

PR #33, opened 2019-09-13
Bump netlink to avoid crashing when gretap interfaces are present

rcarrillocruz:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1751458


danwinship:
I'm a little bit dubious about bumping to the current git master HEAD this close to a release.

How about just patching vendor with this one fix. Then when we do the rebase for 4.3, drop the patch and bump the requirement to `07130f38` then, and we'll have the whole 4.3 cycle to shake out any new bugs.

rcarrillocruz:
let's wait for 4.2 release, no point on either bumping dep now nor vendorize patch and later drop it

danwinship:
@rcarrillocruz needs rebase

rcarrillocruz:
just realized a pushed accidentally bogus files to commit, do not merge. Running out of battery and don't have charging cable around :/ , will amend on Monday

rcarrillocruz:
i could squeeze some of my battery, here's new commit amended

danwinship:
@rcarrillocruz build failure; I think you need to update `github.com/docker/libnetwork` too

rcarrillocruz:
The problem is that docker/libnetwork master branch pins vishvananda/netlink to 1.0.0 :

https://github.com/docker/libnetwork/blob/master/vendor.conf

danwinship:
do you actually get an error when you try to update it? I'm not sure of the exact semantics of `vendor.conf` but a lot of times, glide assumes that if a dependency requests a specific version but we request a newer version, then that's ok...

rcarrillocruz:
No, I didn't get any failures when running 'glide update --strip-vendor', just warns.
Not sure how I can circumvent this, unless waiting for new release on netlink, and libnetwork bumps to it.

pecameron:
@rcarrillocruz @danwinship I get warns as well in library-go

PR #34, opened 2019-09-13
Bug 1751954: images: disable cgo

squeed:
We're still seeing random coredumps / segfaults on rhel7 binaries on rhel8. Disable cgo again.

### DIFF
--- a/images/node/Dockerfile.rhel
+++ b/images/node/Dockerfile.rhel
@@ -1,7 +1,7 @@
 FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.12 AS builder
 WORKDIR /go/src/github.com/openshift/sdn
 COPY . .
-RUN make build --warn-undefined-variables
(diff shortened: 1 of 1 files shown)


danwinship:
So what are the downsides of `CGO_ENABLED=0`? Is there any argument for using it only for the CNI plugin, not for the daemons?

danwinship:
also, do we need to do anything in containernetworking-plugins? ovn-kubernetes?

squeed:
> So what are the downsides of CGO_ENABLED=0? Is there any argument for using it only for the CNI plugin, not for the daemons?

AIUI there are none. It should still use the vDSO, which is the only interesting "static"  library we use. It means we lose FIPS support... which we aren't using anyways.

> also, do we need to do anything in containernetworking-plugins? ovn-kubernetes?
Yup, filed bugs for the multus + cni images. Will file one for ovn-k now.

PR #35, opened 2019-09-19
[WIP] DNM: See if we're racing on informer creation

squeed:


squeed:
/hold cancel
because I want tests to run

squeed:
seems fine for now
/retest e2e-aws-upgrade

danwinship:
The last time this came up we spent a week trying to get it to show up in a debug PR and never succeeded. I think we're going to have to actually commit some additional debugging. Maybe doing the "automatically turn off after a while" thing I suggested.

One thing that might be good would be to add a separate set of informers that just listen and log Service/Endpoints events, so we can compare that to what the proxy sees.

squeed:
I think you're right. Let's come up with a strategy for how we can debug and/or minimize this.

I do want to see if we can trigger it easily too, though.
/test e2e-aws-upgrade

danwinship:
This is being fixed upstream
/close

PR #36, opened 2019-09-23
Bug 1752636: networkpolicy: add a namespaceSelector cache

danwinship:
In systems with very many namespaces and very many network policies, simply adding new namespaces becomes extremely slow, because, eg, every time you add a new namespace, it needs to compare 10,000 "allow-from-default-namespace" policies against it to see if the namespaceSelectors match.

This fixes that by adding a cache of namespaceSelectors. To avoid having to do complicated cache management, we just empty out the cache any time any Namespace changes. It seems to help a lot; https://github.com/danwinship/sdn/commit/1a5dfd65 has an test on top of this. With the cache, on my laptop, it takes 21ms for the setup and 434ms to complete. With the cache disabled (eg, comment out the `np.nsMatchCache[cacheKey] = namespaces` line in networkpolicy.go), it takes 61ms for the setup and 7.55s to complete.

I wanted to make that into an actual unit test, where it would fail if the test seemed to be taking too long, but that wasn't working well. We can't just check the run time, because different hosts, different loads, etc. I was hoping I could compare the time that the early part of the loop took against the time the later part of the loop took, and use that to figure out whether we were O(n^2) or O(2^n). And it almost worked, except that garbage collection and stuff means that sometimes one of the specific intervals I was checking would be very non-representative and it wouldn't work. So I didn't bother actually adding the test. But the cache does work, and we can have QE test this.

### DIFF
--- a/pkg/network/node/networkpolicy.go
+++ b/pkg/network/node/networkpolicy.go
@@ -37,6 +37,9 @@ type networkPolicyPlugin struct {
 	kNamespaces map[string]corev1.Namespace
 	pods        map[ktypes.UID]corev1.Pod
 
+	// namespaceSelector match cache; see selectNamespacesInternal
+	nsMatchCache map[string]map[string]uint32
+
 	runner *async.BoundedFrequencyRunner
 }
 
@@ -72,6 +75,8 @@ func NewNetworkPolicyPlugin() osdnPolicy {
 		namespaces:  make(map[uint32]*npNamespace),
 		kNamespaces: make(map[string]corev1.Namespace),
 		pods:        make(map[ktypes.UID]corev1.Pod),
+
+		nsMatchCache: make(map[string]map[string]uint32),
 	}
 }
 
@@ -163,6 +168,7 @@ func (np *networkPolicyPlugin) AddNetNamespace(netns *networkv1.NetNamespace) {
 		return
 	}
 
+	np.flushNSMatchCache()
 	np.namespaces[netns.NetID] = &npNamespace{
 		name:     netns.NetName,
 		vnid:     netns.NetID,
@@ -184,6 +190,7 @@ func (np *networkPolicyPlugin) DeleteNetNamespace(netns *networkv1.NetNamespace)
 	defer np.lock.Unlock()
 
 	if npns, exists := np.namespaces[netns.NetID]; exists {
+		np.flushNSMatchCache()
 		if npns.inUse {
 			npns.inUse = false
 			// We call syncNamespaceFlows() not syncNamespace() because it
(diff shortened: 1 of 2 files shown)


pkg/network/node/networkpolicy.go
```diff
@@ -303,8 +304,34 @@ func (np *networkPolicyPlugin) SyncVNIDRules() {
...
+// cacheKey here is non-deterministic, but that just means that, eg, we might compute the
+// results twice rather than just once, and twice is still better than 10,000 times.
+//
+// (The match cache gets flushed as needed by refreshNetworkPolicies.)
+func (np *networkPolicyPlugin) selectNamespacesInternal(sel labels.Selector) map[string]uint32 {
+	cacheKey := sel.String()
+	namespaces, wasCached := np.nsMatchCache[cacheKey]
+	if !wasCached {
+		namespaces = make(map[string]uint32)
+		for vnid, ns := range np.namespaces {
```
squeed: Do we need to flush the cache on `np.namespaces` changes too?

squeed:
While you're at it, calling `np.syncNamespace()` in a for-loop in `refreshNetworkPolicies` is quite silly. It probably contributes to saturation of the BoundedRunner. What it does is set the namespace to dirty and trigger the runner.

In this case, we should set *all* the namespaces to dirty and trigger the runner once.

squeed:
Skimming through the test cases, I would like to see coverage for labels being added and removed from pods, and namespaces. At a first glance, I don't see those.

pkg/network/node/networkpolicy.go
danwinship: Hm... VNID changes aren't allowed in ovs-networkpolicy, so we don't have to worry about updates. In between the deletion of a NetNamespace and its Namespace, the cache might temporarily contain stale rules referring to non-existent VNIDs, which isn't optimal, but won't hurt anything. The only way we could get actually *wrong* rules would be if we got:

    - ADD Namespace bar
    - DEL NetNamespace foo, vnid 42
    - (cache may contain stale rules)
    - ADD NetNamespace bar, vnid 42
    - (cache may contain incorrect rules)
    - DEL Namespace foo
    - (cache is flushed)

I don't think an ADD by itself could cause problems in any circumstances.

danwinship:
> Skimming through the test cases, I would like to see coverage for labels being added and removed from pods, and namespaces. At a first glance, I don't see those.

Unless you think there's a chance that that case got broken by this patch, let's add that later. People want this backported to 3.11 asap.

squeed:
> While you're at it, calling np.syncNamespace() in a for-loop in refreshNetworkPolicies is quite silly.

Right, none of this matters because `refreshNetworkPolicies` holds the lock. So, we're effectively forcing a batch anyways.

PR #38, opened 2019-09-25
Bump dependencies and assorted code fixes for Kubernetes 1.16

rcarrillocruz:


danwinship:
> Commits 1 Files changed 18,842 

Can you repush with the "update deps" part as its own commit, to make it easier to review the rest?

rcarrillocruz:
> > Commits 1 Files changed 18,842
> 
> Can you repush with the "update deps" part as its own commit, to make it easier to review the rest?

Sure, let me see

glide.yaml
```diff
@@ -2,121 +2,101 @@ package: github.com/openshift/sdn
...
+  version: kubernetes-1.16.0
 - package: k8s.io/code-generator
-  version: kubernetes-1.14.0
+  version: kubernetes-1.16.0
 - package: k8s.io/component-base
-  version: kubernetes-1.14.0
+  version: kubernetes-1.16.0
 - package: k8s.io/csi-api
-  version: kubernetes-1.14.0
+  version: master # FIXME
```
danwinship: So since this still doesn't have any v1.16 tag it should be pinned to the current commit SHA of master, not to "master" itself.
Except it seems like "`make update-deps && make`" still passes even if you remove the `k8s.io/csi-api` entry entirely. So just kill that I guess.

glide.yaml
```diff
@@ -2,121 +2,101 @@ package: github.com/openshift/sdn
...
   version: v1.3.1-coreos.6
+- package: github.com/coreos/go-systemd
+  version: v20
 - package: github.com/google/cadvisor
   version: v0.32.0
 
 # openshift second
 - package: github.com/openshift/api
-  version: master
+  version: prebase-1.16
```
danwinship: Really? They didn't bother to create new tags? lame...

danwinship:
Actually, I think there's probably a lot of cruft in `glide.yaml` that we should remove so that can be a separate PR.
/lgtm
/approve

PR #39, opened 2019-09-26
Bug 1801737: drop firewalld monitoring, add better iptables monitor

danwinship:
Backport of https://github.com/kubernetes/kubernetes/pull/81517 to the 4.3 sdn branch.

This backport will fix the monitoring in kube-proxy and sdn's NodeIPTables. The same PR is also being cherry-picked to origin master where it will fix monitoring in kubelet and add an e2e test. (The e2e test passes with both the old AddReloadFunc code and the new Monitor code, so it doesn't matter whether origin merges before or after sdn.)

This is an update to the original PR that skips trying to update openshift/kubernetes and just puts everything here instead.

Additionally, since it's now fairly late in the 4.3 cycle, I decided to preserve the old metrics behavior, of updating the sync timestamp every `iptablesSyncPeriod` even when nothing changed, as long as there are no errors. (This will also make this simpler to backport, since we were going to need to do that for older releases anyway.)

Part of the fix for https://bugzilla.redhat.com/show_bug.cgi?id=1727441 "kube-proxy periodic iptables reloads are extremely disruptive in large clusters"


danwinship:
@squeed: the backport of the iptables fixes. Note the `<drop>` commit which undoes part of the upstream commit so as to preserve our metrics behavior

juanluisvaladas:
@danwinship The code looks really good, but I think some of the commits should go to github.com/openshift/kubernetes sdn-4.3-kubernetes-1.16.0

First because I think it's a more correct way of patching vendor, but mostly because with https://github.com/openshift/sdn/pull/102 (the migration to go.mod) and I think we really need to figure this out before proceeding...

You're patching vendor/ and glide diff. This is a prolem because there is no glide.diff equivalent in go modules and what I'm doing is adding a patch/ directory where I add the patched modules. It's a horrible hack but since there is only one patched module line at the time of the migration to 4.4 and it's not a big one, that sounded like a reasonable option. However, with that many changes this sounds like a terrible idea...

If we look carefully at the github.com/openshift/kubernetes sdn-4.4-kubernetes-1.17.2 branch the following commits are already in github.com/openshift/kubernetes sdn-4.4-kubernetes-1.17.2:
* UPSTREAM: 81517: Drop iptables firewalld monitoring support
* UPSTREAM: 81517: Add iptables.Monitor, use it from kubelet and kube-p…
* UPSTREAM: 83402: iptables.Monitor: don't be fooled by "could not get …

And the only missing commit in vendor is:
UPSTREAM: <drop>: preserve old health/metrics behavior 
Which must be dropped anyway.

So my proposal is it split this into 2 PRs:
1- Make a PR to github.com/openshift/kubernetes sdn-4.3-kubernetes-1.17.2 branch with:
* UPSTREAM: 81517: Drop iptables firewalld monitoring support
* UPSTREAM: 81517: Add iptables.Monitor, use it from kubelet and kube-p…
* UPSTREAM: 83402: iptables.Monitor: don't be fooled by "could not get …
* UPSTREAM: <drop>: preserve old health/metrics behavior

2- Keep this PR on hold, and once the previous PR mentioned is merged, merge this PR doing only two commits:
* A commit with a glide update to update 
* Commit node: Update for utiliptables API changes

Once this is done, I'll work on the https://github.com/openshift/sdn/pull/102 PR which is the migration to go modules.

danwinship:
The issue is that `openshift/kubernetes` is in a broken state, permissions-and-ci-wise, and we effectively can't modify a branch once it has been pushed.

I guess the way to fix this is that when we first push a new branch as part of a rebase, we should replace the top-level OWNERS file with a copy of sdn's OWNERS file. Then for future changes to that branch, we have permission to tell the bot to merge it, and to ignore failing tests.

For the 4.3 and earlier branches the easiest fix would be to abandon the existing branches and start new ones. (We should probably rebase to a later upstream point release anyway...)

juanluisvaladas:
> For the 4.3 and earlier branches the easiest fix would be to abandon the existing branches and start new ones. (We should probably rebase to a later upstream point release anyway...)

We need to backport this all the way back to 3.11 so having these massive glide.diff all the way back seems like a rather bad idea.

Should I block the migration to go modules https://github.com/openshift/sdn/pull/102 and wait for this to merge first? If we merge #102 first it this means we will be committing to use go modules for all the 4.3 branch instead of glide, which I don't think is our current intention.

juanluisvaladas:
Because some of these fixes are already in kubernetes 1.17, I needed to chose between moving k8s.io/kubernetes/pkg/util/dbus to github.com/openshift/sdn/util dbus, or cherry-picking  https://github.com/openshift/sdn/pull/39/commits/24bbe5065ac2a040a4ffe40c4f23293a2f457652 in the new branch. The latter seemed more adequate to me. 

If you are OK with it, we can merge this in master in master in https://github.com/openshift/sdn/pull/108 and change the target in this PR release-4.3, because it already has everything needed for the backport which shouldn't have conflicts.

danwinship:
> Because some of these fixes are already in kubernetes 1.17

Yeah, I didn't expect that you'd actually merge this whole PR to master after doing the rebase. I left it around more as a basis for the 4.3 commit.

> I needed to chose between moving k8s.io/kubernetes/pkg/util/dbus to github.com/openshift/sdn/util dbus, or cherry-picking [24bbe50](https://github.com/openshift/sdn/commit/24bbe5065ac2a040a4ffe40c4f23293a2f457652) in the new branch.

Moving `pkg/util/dbus` wouldn't help anyway since the only thing it's used for is to generate an argument to pass to `utiliptables.New()`, which no longer takes that argument in 1.17. So yes, you need that cherry-pick.

danwinship:
I don't think we can actually change the branch this is based on. But you can just close it and open a new PR

juanluisvaladas:
> I don't think we can actually change the branch this is based on. But you can just close it and open a new PR

You can, just press the edit on the top right corner (the one next to the title) and then right below the title the target branch becomes a drop down list where you can set release-4.3.

danwinship:
ah, cool. I hadn't noticed that button let you edit more than just the title, and was assuming that "being able to change the branch of a commit" was something you could *normally* do in github, but that wasn't possible the way we set up permissions

juanluisvaladas:
/retitle Bug 1797042: drop firewalld monitoring, add better iptables monitor

bparees:
@danwinship trying to follow this chain....

it says it fixes https://bugzilla.redhat.com/show_bug.cgi?id=1797042

which references 4.4 bug https://bugzilla.redhat.com/show_bug.cgi?id=1787488

but 4.4 was fixed by: https://github.com/openshift/sdn/pull/88

and this PR does not resemble that PR.  Can you help me understand that this is the right fix and that these changes been validated in 4.4 already?

juanluisvaladas:
This is my fault actually, I must have messed up the BZs with the duplicate. I'll fix it.

juanluisvaladas:
/retitle Bug 1801737: drop firewalld monitoring, add better iptables monitor

juanluisvaladas:
The BZ mess is fixed.

Now we reference in this PR:
1801737 - [4.3] kube-proxy periodic iptables reloads are extremely disruptive in large clusters
which references:
1727441 - kube-proxy periodic iptables reloads are extremely disruptive in large clusters
Which is closed duplicate of
1803149 - Rebase SDN 4.4 to kubernetes 1.17
Which  includes te fix in 4.4

Sorry for the mess with BZ.

bparees:
needs a rebase now, but based on priority it's not going to go this week anyway due to the number of other changes already accepted.

bparees:
(I see the BZ priority is now marked urgent.  if you fix up the merge conflict i can tag it tomorrow.  please ping me directly when it's fixed so i don't miss this one)

danwinship:
> (I see the BZ priority is now marked urgent. if you fix up the merge conflict i can tag it tomorrow. please ping me directly when it's fixed so i don't miss this one)

Oops, I fixed it but forgot to ping you. At any rate, it's the 3.11 backport that's urgent, not the 4.3, so it doesn't really matter if this was in 4.3.5 since we don't have the 3.11 backport done yet anyway.

shawn-hurley:
is this still urgent? @danwinship

danwinship:
@shawn-hurley yes. (The 3.11 backport is urgent, and we don't want to merge that before fixing it in the later releases.)

PR #42, opened 2019-09-30
further NetworkPolicy caching fixes

danwinship:
The previous fix (#36) did not solve the customer problem because in addition to having thousands of identical copies of the "allow-from-global-namespace" policies (which the cache handles well), they also have thousands of single-namespace-to-single-namespace policies, which the cache didn't optimize as well, because we keep flushing the entire cache even though 99% of the matches didn't change. So this (specifically the second commit) tries to fix that by just making incremental updates to the cache when namespaces change, rather than destroying it every time. See the commit messages for more details

### DIFF
--- a/pkg/network/node/networkpolicy.go
+++ b/pkg/network/node/networkpolicy.go
@@ -41,7 +41,7 @@ type networkPolicyPlugin struct {
 	// for all flow-generating methods
 	namespaces map[uint32]*npNamespace
 	// nsMatchCache caches matches for namespaceSelectors; see selectNamespaceInternal
-	nsMatchCache map[string]map[string]uint32
+	nsMatchCache map[string]*npCacheEntry
 
 	pods map[ktypes.UID]corev1.Pod
 }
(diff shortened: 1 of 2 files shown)


pkg/network/node/networkpolicy.go
```diff
@@ -314,25 +346,39 @@ func (np *networkPolicyPlugin) SyncVNIDRules() {
...
-				if sel.Matches(labels.Set(kns.Labels)) {
-					namespaces[ns.name] = vnid
-				}
+func (np *networkPolicyPlugin) selectNamespacesInternal(selector labels.Selector) map[string]uint32 {
+	cacheKey := selector.String()
+	match := np.nsMatchCache[cacheKey]
+	if match == nil {
+		match = &npCacheEntry{selector: selector, matches: make(map[string]uint32)}
+		for vnid, npns := range np.namespaces {
+			if npns.gotNamespace && selector.Matches(labels.Set(npns.labels)) {
```
squeed: does this need to be `gotNamespace && gotNetNamespace` ?

danwinship: no; we're iterating over `np.namespaces`, which is indexed by vnid and so necessarily only contains the namespaces for which `gotNetNamespace` is `true`.

squeed: oh duh. good point.

squeed:
Stared at this for a while, everything makes sense.

/lgtm

PR #43, opened 2019-10-01
WIP Implement optional kube-proxy service abstraction from upstream

rcarrillocruz:
PR 71355

### DIFF
--- a/pkg/openshift-sdn/informers.go
+++ b/pkg/openshift-sdn/informers.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 	"time"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kinformers "k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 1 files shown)


danwinship:
Don't reference PRs/issues in commit messages. It ends up totally spamming that PR with comments. :-/

pkg/openshift-sdn/informers.go
```diff
@@ -42,7 +43,10 @@ func (sdn *OpenShiftSDN) buildInformers() error {
 	}
 
 	kubeInformers := kinformers.NewSharedInformerFactory(kubeClient, sdn.ProxyConfig.IPTables.SyncPeriod.Duration)
-	networkInformers := networkinformers.NewSharedInformerFactory(networkClient, defaultInformerResyncPeriod)
+	networkInformers := networkinformers.NewSharedInformerFactoryWithOptions(networkClient, defaultInformerResyncPeriod,
+		networkinformers.WithTweakListOptions(func(options *metav1.ListOptions) {
+			options.LabelSelector = "!service.kubernetes.io/service-proxy-name"
+		}))
```
danwinship: So as you already noticed, you wanted to change `kubeInformers` not `networkInformers`, but you can't do it like this anyway; these informers get used by *all* of the sdn, so if you filter out some services here, then, the sdn Service handling code won't see them either. That's why I was saying to filter them out from `pkg/network/proxy/` instead.

rcarrillocruz: Well, I was thinking of doing two separate PRs, just so we have a 1-1 for https://github.com/kubernetes/kubernetes/pull/71355/files and https://github.com/kubernetes/kubernetes/pull/81612/files .
The filtering for endpoints would come in  other PR.

PR #44, opened 2019-10-03
Garbage collect EgressNetworkPolicies

juanluisvaladas:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1743618

### DIFF
--- a/pkg/network/node/egress_network_policy.go
+++ b/pkg/network/node/egress_network_policy.go
@@ -25,6 +25,7 @@ func (plugin *OsdnNode) SetupEgressNetworkPolicy() error {
 	plugin.egressPoliciesLock.Lock()
 	defer plugin.egressPoliciesLock.Unlock()
 
+	plugin.garbageCollectEgressPolicy()
 	for _, policy := range policies.Items {
 		vnid, err := plugin.policy.GetVNID(policy.Namespace)
 		if err != nil {
@@ -67,7 +68,8 @@ func (plugin *OsdnNode) handleDeleteEgressNetworkPolicy(obj interface{}) {
 func (plugin *OsdnNode) handleEgressNetworkPolicy(policy *networkapi.EgressNetworkPolicy, eventType watch.EventType) {
 	vnid, err := plugin.policy.GetVNID(policy.Namespace)
 	if err != nil {
-		utilruntime.HandleError(fmt.Errorf("Could not find netid for namespace %q: %v", policy.Namespace, err))
+		utilruntime.HandleError(fmt.Errorf("Could not find netid for namespace %q: %v. Attempting garbage collection.", policy.Namespace, err))
+		plugin.garbageCollectEgressPolicy()
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


danwinship:
So this isn't right...

NetworkPolicy needs to do asynchronous garbage collection because some of its flows are being used by multiple policies and it's tricky to figure out from the deletion-handling-code if the flow is now unused or not.

EgressNetworkPolicy doesn't have the same optimizations and so doesn't have the same problems; we know that when an EgressNetworkPolicy is deleted, we can immediately delete every flow associated with that policy, and that's what we're supposed to be doing. If this is failing, then it's a bug in the chain of calls that starts with `handleDeleteEgressNetworkPolicy`.

I suspect if you look at the sdn logs when you do this, you'll find that the "Could not find netid for namespace" error in `handleEgressNetworkPolicy` is getting hit, because for at least some of the namespaces, we're processing the deletion of its NetNamespace before we process the deletion of the EgressNetworkPolicy, so by the time the `plugin.policy.GetVNID()` call happens, we've forgotten the namespace's VNID. I'm not sure off the top of my head what the best fix is because there's some ugliness involving the multitenant plugin and shared VNIDs.

juanluisvaladas:
> NetworkPolicy needs to do asynchronous garbage collection because some of its flows are being used by multiple policies and it's tricky to figure out from the deletion-handling-code if the flow is now unused or not.

> EgressNetworkPolicy doesn't have the same optimizations and so doesn't have the same problems; we know that when an EgressNetworkPolicy is deleted, we can immediately delete every flow associated with that policy, and that's what we're supposed to be doing. If this is failing, then it's a bug in the chain of calls that starts with handleDeleteEgressNetworkPolicy.
> I suspect if you look at the sdn logs when you do this, you'll find that the "Could not find netid for namespace" error in handleEgressNetworkPolicy is getting hit, because for at least some of the namespaces, we're processing the deletion of its NetNamespace before we process the deletion of the EgressNetworkPolicy, so by the time the plugin.policy.GetVNID() call happens, we've forgotten the namespace's VNID.

Yes, all of this is correct and a very accurate description of the original problem.

>  I'm not sure off the top of my head what the best fix is because there's some ugliness involving the multitenant plugin and shared VNIDs.

Are shared VNIDs a case we want to handle? according to our [docs](https://docs.openshift.com/container-platform/3.11/admin_guide/managing_networking.html#admin-guide-limit-pod-access-egress) it shouldn't be possible:
```If you merge two projects together (via oc adm pod-network join-projects), then you cannot use egress policy in any of the joined projects.```

Additionally, there is another issue, when you start the sdn pod with a dirty table 101, there is no clean up mechanism

juanluisvaladas:
Actually we don't allow egressnetworkpolicies at all in shared namespaces:


https://github.com/openshift/sdn/blob/e75836340e2574789fad58abb873cfd027c00f67/pkg/network/node/ovscontroller.go#L490-L497

danwinship:
right, but the code has to be able to *notice* that case, so that puts some constraints on what you can and can't do

juanluisvaladas:
I'm afraid I'm not understanding the problem, sorry.

garbageCollectEgressPolicy only calls plugin.oc.UpdateEgressNetworkPolicyRules on vnid if len(plugin.egressPolicies[vnid]) == 0, so if there is at least one netnamespace with such vnid that has at least egressNetworkPolicy, calling garbageCollectEgressPolicy won't wipe any rule where reg0=vnid.

This is case where joined projects have rules is enforced by OsdnNode.updateEgressNetworkPolicyRules which does this check and removes the rules.

garbageCollectEgressPolicy is called:
1- On line 28 and then in the same function we call OsdnNode.updateEgressNetworkPolicyRules later on line 41

2- On function handleEgressNetworkPolicy if plugin.policy.GetVNID(policy.Namespace) returns an error, so if GetVNID it means the cache has no policy for a namespace on that VNID and therefore should be clear, if there is one or multiple policies it should never be called and it should be handled on that same function on line 94 calling plugin.updateEgressNetworkPolicyRules(vnid)

So my question is: Under which circumstances
 is the problematic one and on which call

danwinship:
No, I just mean, fixing this to deal with the NetNamespace being deleted before the EgressNetworkPolicy means that you have to no longer rely on `policy.GetVNID` working correctly at EgressNetworkPolicy deletion time, which means you have to cache the VNIDs or something, **BUT** however you do that has to take into account the possibility of joined VNIDs, so be careful

PR #48, opened 2019-10-07
Make DNS querying more efficient by querying once per dns name

abhat:
Refactor the datastructures to query DNS names more efficiently as
the same DNS name may appear across multiple policies.

### DIFF
--- a/pkg/network/common/dns.go
+++ b/pkg/network/common/dns.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/miekg/dns"
 
-	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
 )
@@ -78,44 +77,38 @@ func (d *DNS) Add(dns string) error {
 	defer d.lock.Unlock()
 
 	d.dnsMap[dns] = dnsValue{}
-	err, _ := d.updateOne(dns)
+	_, err := d.updateOne(dns)
 	if err != nil {
 		delete(d.dnsMap, dns)
 	}
 	return err
 }
 
-func (d *DNS) Update() (error, bool) {
+func (d *DNS) Delete(dns string) {
+	d.lock.Lock()
+	defer d.lock.Unlock()
+	delete(d.dnsMap, dns)
+}
+
+func (d *DNS) Update(dnsName string) (bool, error) {
 	d.lock.Lock()
 	defer d.lock.Unlock()
 
-	errList := []error{}
-	changed := false
-	for dns := range d.dnsMap {
-		err, updated := d.updateOne(dns)
-		if err != nil {
-			errList = append(errList, err)
-			continue
-		}
-		if updated {
-			changed = true
-		}
-	}
-	return kerrors.NewAggregate(errList), changed
+	return d.updateOne(dnsName)
 }
 
-func (d *DNS) updateOne(dns string) (error, bool) {
+func (d *DNS) updateOne(dns string) (bool, error) {
 	res, ok := d.dnsMap[dns]
 	if !ok {
 		// Should not happen, all operations on dnsMap are synchronized by d.lock
-		return fmt.Errorf("DNS value not found in dnsMap for domain: %q", dns), false
+		return false, fmt.Errorf("DNS value not found in dnsMap for domain: %q", dns)
 	}
 
 	ips, ttl, err := d.getIPsAndMinTTL(dns)
 	if err != nil {
 		res.nextQueryTime = time.Now().Add(defaultTTL)
 		d.dnsMap[dns] = res
-		return err, false
+		return false, err
 	}
 
 	changed := false
@@ -126,7 +119,7 @@ func (d *DNS) updateOne(dns string) (error, bool) {
 	res.ttl = ttl
 	res.nextQueryTime = time.Now().Add(res.ttl)
 	d.dnsMap[dns] = res
-	return nil, changed
+	return changed, nil
 }
 
 func (d *DNS) getIPsAndMinTTL(domain string) ([]net.IP, time.Duration, error) {
@@ -183,20 +176,23 @@ func (d *DNS) getIPsAndMinTTL(domain string) ([]net.IP, time.Duration, error) {
 	return removeDuplicateIPs(ips), ttl, nil
 }
 
-func (d *DNS) GetMinQueryTime() (time.Time, bool) {
+func (d *DNS) GetNextQueryTime() (time.Time, string, bool) {
 	d.lock.Lock()
 	defer d.lock.Unlock()
 
 	timeSet := false
 	var minTime time.Time
-	for _, res := range d.dnsMap {
+	var dns string
+
+	for dnsName, res := range d.dnsMap {
 		if (timeSet == false) || res.nextQueryTime.Before(minTime) {
 			timeSet = true
 			minTime = res.nextQueryTime
+			dns = dnsName
 		}
 	}
 
-	return minTime, timeSet
+	return minTime, dns, timeSet
 }
 
 func ipsEqual(oldips, newips []net.IP) bool {
--- a/pkg/network/common/egress_dns.go
+++ b/pkg/network/common/egress_dns.go
@@ -5,10 +5,11 @@ import (
 	"sync"
 	"time"
 
+	networkv1 "github.com/openshift/api/network/v1"
 	ktypes "k8s.io/apimachinery/pkg/types"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
-
-	networkv1 "github.com/openshift/api/network/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/klog"
 )
 
 type EgressDNSUpdate struct {
@@ -16,11 +17,15 @@ type EgressDNSUpdate struct {
 	Namespace string
 }
 
+type EgressDNSUpdates []EgressDNSUpdate
+
 type EgressDNS struct {
 	// Protects pdMap/namespaces operations
 	lock sync.Mutex
-	// Holds Egress DNS entries for each policy
-	pdMap map[ktypes.UID]*DNS
+	// holds DNS entries globally
+	dns *DNS
+	// this map holds which DNS names are in what policy objects
+	dnsNamesToPolicies map[string]sets.String
 	// Maintain namespaces for each policy to avoid querying etcd in syncEgressDNSPolicyRules()
 	namespaces map[ktypes.UID]string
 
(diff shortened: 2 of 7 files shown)


pkg/network/common/egress_dns.go
```diff
@@ -96,13 +123,13 @@ func (e *EgressDNS) Sync() {
...
 			} else {
-				err, changed := e.Update(policyUID)
+				err, changed := e.Update(dnsName)
 				if err != nil {
 					utilruntime.HandleError(err)
 				}
 
 				if changed {
-					e.Updates <- EgressDNSUpdate{policyUID, policyNamespace}
+					e.Updates <- EgressDNSUpdate{updates}
```
danwinship: so rather than wrapping the list of updates, you could just do

    for _, update := range updates {
            e.Updates <- update
    }

and then you don't need to change `egress_network_policy.go` or `proxy.go` at all

pkg/network/common/egress_dns.go
```diff
@@ -116,39 +143,41 @@ func (e *EgressDNS) Sync() {
 	}
 }
 
-func (e *EgressDNS) GetMinQueryTime() (time.Time, ktypes.UID, string, bool) {
+func (e *EgressDNS) GetMinQueryTime() (time.Time, string, []PolicyUpdate, bool) {
```
danwinship: This function's name was pretty non-obvious before, and is now worse. Maybe `GetNextUpdate`?

pkg/network/common/egress_dns.go
```diff
@@ -116,39 +143,41 @@ func (e *EgressDNS) Sync() {
...
 	e.lock.Lock()
 	defer e.lock.Unlock()
+	policyUpdates := make([]PolicyUpdate, 0)
+	tm, dnsName, timeSet := e.dns.GetMinQueryTime()
+	if !timeSet {
+		klog.Warningf("Time was not set for dns name: %s", dnsName)
```
danwinship: If `timeSet` is `false` then `dnsName` will be `""`. But this is not an error anyway, it just means there are no DNS names in use and so we don't need to poll again until there are.

pkg/network/common/egress_dns.go
```diff
@@ -116,39 +143,41 @@ func (e *EgressDNS) Sync() {
...
 func (e *EgressDNS) GetIPs(policy networkv1.EgressNetworkPolicy, dnsName string) []net.IP {
 	e.lock.Lock()
 	defer e.lock.Unlock()
 
-	dnsInfo, ok := e.pdMap[policy.UID]
-	if !ok {
+	if uids, ok := e.dnsNamesToUIDsMap[dnsName]; ok {
+		for _, uid := range uids {
+			if uid == policy.UID {
+				return e.dns.Get(dnsName).ips
```
danwinship: This is weird. Just change `GetIPs` to no longer take a `policy`.

pkg/network/common/dns.go
```diff
@@ -76,32 +75,32 @@ func (d *DNS) Get(dns string) dnsValue {
...
 	}
-	return err
+	return nil
 }
 
-func (d *DNS) Update() (error, bool) {
+func (d *DNS) Delete(dns string) {
 	d.lock.Lock()
 	defer d.lock.Unlock()
+	delete(d.dnsMap, dns)
```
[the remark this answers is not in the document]
abhat: The refcount is kept at the egressdns instance level. The dnsNamesToUIDMap has a slice of policy UIDs. Once that slice becomes empty, the dns name entry can be deleted from that map as well as from the dnsMap inside the `dnsInfo`. For consistency of API, I will make sure that all the calls to Add are elided correctly by the caller.

pkg/network/common/dns.go
```diff
@@ -76,32 +75,32 @@ func (d *DNS) Get(dns string) dnsValue {
 func (d *DNS) Add(dns string) error {
 	d.lock.Lock()
 	defer d.lock.Unlock()
-
-	d.dnsMap[dns] = dnsValue{}
-	err, _ := d.updateOne(dns)
-	if err != nil {
-		delete(d.dnsMap, dns)
+	if _, ok := d.dnsMap[dns]; !ok {
```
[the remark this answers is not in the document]
abhat: 👍

abhat:
> It would be great to add a `pkg/network/common/egress_dns_test.go` with some unit tests. You could change `DNS` into an interface type, and then pass a mock DNS implementation to `NewEgressDNS` to test it.

Will look into adding some unit-tests for this in a separate PR.

pkg/network/common/dns.go
```diff
@@ -76,32 +75,32 @@ func (d *DNS) Get(dns string) dnsValue {
...
+func (d *DNS) Update(dnsName string) (error, bool) {
+	d.lock.Lock()
+	defer d.lock.Unlock()
+
+	err, updated := d.updateOne(dnsName)
+	if err != nil {
+		return err, false
 	}
-	return kerrors.NewAggregate(errList), changed
+	return nil, updated
```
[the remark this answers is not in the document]
abhat: will update the `updateOne` and the `Update` apis to change the order of the return and make the necessary change here to just return the result of the call to `updateOne`

pkg/network/common/egress_dns.go
```diff
@@ -5,22 +5,27 @@ import (
...
+	"k8s.io/klog"
 )
 
-type EgressDNSUpdate struct {
+type PolicyUpdate struct {
 	UID       ktypes.UID
 	Namespace string
 }
+type EgressDNSUpdate struct {
+	DNSUpdates []PolicyUpdate
```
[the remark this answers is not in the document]
abhat: I am making the API change to batch the policy updates as a single update.

pkg/network/common/egress_dns.go
```diff
@@ -5,22 +5,27 @@ import (
...
 
 type EgressDNS struct {
 	// Protects pdMap/namespaces operations
 	lock sync.Mutex
-	// Holds Egress DNS entries for each policy
-	pdMap map[ktypes.UID]*DNS
+	// holds DNS entries globally
+	dns *DNS
+	//this will be the map that holds which DNS names are in what policy objects
+	dnsNamesToUIDsMap map[string][]ktypes.UID
```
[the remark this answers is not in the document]
abhat: 👍

pkg/network/common/egress_dns.go
```diff
@@ -32,62 +37,84 @@ type EgressDNS struct {
 }
 
 func NewEgressDNS() *EgressDNS {
+	dnsInfo, err := NewDNS("/etc/resolv.conf")
+	if err != nil {
+		utilruntime.HandleError(err)
+		return nil
```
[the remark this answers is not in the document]
abhat: 👍

pkg/network/common/egress_dns.go
```diff
@@ -32,62 +37,84 @@ type EgressDNS struct {
...
+	//if the slice is empty at this point, delete the entry from the dns object too
+	//also remove the policy entry from the namespaces map.
+	for _, rule := range policy.Spec.Egress {
+		if len(rule.To.DNSName) > 0 {
+			if uids, ok := e.dnsNamesToUIDsMap[rule.To.DNSName]; ok {
+				for i, uid := range uids {
+					if uid == policy.UID {
+						uids[i] = uids[len(uids)-1]
+						uids[len(uids)-1] = ""
+						uids = uids[:len(uids)-1]
```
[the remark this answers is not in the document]
abhat: 👍

pkg/network/common/egress_dns.go
abhat: I would rather we batch all the updates as a single channel update to the `Sync` observer. That way we don't incur un-necessary overhead of tic-toc'ing between the two go-routines.

pkg/network/common/egress_dns.go
abhat: Will change the function name.

pkg/network/common/egress_dns.go
abhat: Agreed, will change the log message to be more appropriate.

pkg/network/common/egress_dns.go
abhat: I thought about it. The only reason I am still passing in the `policy` is to make sure we don't return valid IPs for a policy that doesn't have the corresponding `dnsName` in its rule set. So in some ways it is checking that the GetIPs only returns valid ips for policies that indeed have the dns name in the egress rules. I am OK with removing it as well if that additional check is not needed semantically.

pkg/network/common/egress_dns.go
danwinship: If you want to do that then I'd say just return the array of changes directly rather than wrapping it up in a struct

pkg/network/common/egress_dns.go
danwinship: (There doesn't need to be a log message at all.)

pkg/network/common/egress_dns.go
danwinship: The only reason it was taking a policy before though was because it *needed* to in order to look up the information. The thing that the caller wants is the name-to-IP mapping. Being forced to pass a policy was just an implementation detail.

pkg/network/common/egress_dns.go
```diff
@@ -5,89 +5,110 @@ import (
...
 	ktypes "k8s.io/apimachinery/pkg/types"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
-
-	networkv1 "github.com/openshift/api/network/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/klog"
 )
 
-type EgressDNSUpdate struct {
+type PolicyUpdate struct {
```
danwinship: Keep the old name, please; it has to be unique across `pkg/network/common/`, so `EgressDNSUpdate` is better. (Likewise with `PolicyUpdates`, although you don't actually need to have a type alias there, you can just use "`chan []EgressDNSUpdate`", etc

pkg/network/common/egress_dns.go
```diff
@@ -5,89 +5,110 @@ import (
...
 		if len(rule.To.DNSName) > 0 {
-			if err := dnsInfo.Add(rule.To.DNSName); err != nil {
-				utilruntime.HandleError(err)
+			if uids, exists := e.dnsNamesToPolicies[rule.To.DNSName]; !exists {
+				e.dnsNamesToPolicies[rule.To.DNSName] = sets.NewString(string(policy.UID))
+				//only call Add if the dnsName doesn't exist in the dnsNamesToPolicies
+				if err := e.dns.Add(rule.To.DNSName); err != nil {
+					utilruntime.HandleError(err)
+				}
+
```
danwinship: (gratuitous blank line)

pkg/network/common/egress_dns.go
```diff
@@ -5,89 +5,110 @@ import (
...
+			} else {
+				e.dnsNamesToPolicies[rule.To.DNSName] = uids.Insert(string(policy.UID))
```
danwinship: You don't need to re-set `e.dnsNamesToPolicies[rule.To.DNSName]`. `uids` will still point to the same object.

pkg/network/node/node.go
```diff
@@ -155,6 +155,12 @@ func New(c *OsdnNodeConfig) (*OsdnNode, error) {
 		masqBit = uint32(*c.MasqueradeBit)
 	}
 
+	egressDNS, err := common.NewEgressDNS()
+
+	if err != nil {
```
danwinship: (no blank line between setting an `err` and checking it)

pkg/network/proxy/proxy.go
```diff
@@ -76,12 +76,17 @@ type OsdnProxy struct {
 // Called by higher layers to create the proxy plugin instance
 func New(networkClient networkclient.Interface, kClient kubernetes.Interface,
 	networkInformers networkinformers.SharedInformerFactory) (*OsdnProxy, error) {
+	egressDNS, err := common.NewEgressDNS()
+
+	if err != nil {
```
danwinship: likewise

pkg/network/common/egress_dns.go
abhat: Will change the name back to `EgressDNSUpdate`, I prefer the type alias for brevity though.

pkg/network/common/egress_dns.go
abhat: Agreed, this is to make sure that if someone changes the value type back to a slice, the code will still work as is. Of course the `Insert` would have to change to `append()` but then it's easy to forget assigning the result of the `append` as the value type of the map entry.

PR #51, opened 2019-10-09
Bug 1753216: Clean up egress IPs on startup

danwinship:
If a node lost an egress IP while the sdn process was being restarted, the sdn wouldn't remove the stale egress IP from the node's interface, so you'd end up with two nodes fighting over the packets.

This fixes that by (1) labeling the egress IPs we add so that we can recognize them later, and (2) running a cleanup after we've finished syncing with the master on startup.

(Also, `node.egressIPWatcher` had a mutex that it wasn't using... but it turns out that this isn't a problem because its methods are only called from `common.EgressIPTracker`, which *does* use a mutex when calling the watcher methods. I probably meant to delete the mutex when I split out the `common` parts of the egress IP code, but forgot.)


dcbw:
clever
/lgtm

PR #53, opened 2019-10-15
WIP Remove selinux packages from SDN images

rcarrillocruz:


images/node/Dockerfile
```diff
@@ -14,6 +14,7 @@ RUN INSTALL_PKGS=" \
       iputils binutils xz sysvinit-tools dbus nftables \
       " && \
     yum install -y --setopt=tsflags=nodocs --setopt=skip_missing_names_on_install=False $INSTALL_PKGS && \
+    rpm -e --nodeps openvswitch-selinux-extra-policy && \
```
danwinship: The problem isn't just that package though, it's the packages that get dragged in as its dependencies (especially selinux-policy-targeted). (The overall problem that that jira card is targeting is that the SDN image is too big.)

I was thinking the fix for that jira card might be partly upstream. eg, get them to make `openvswitch-selinux-extra-policy` a "weak dependency" and then we could install OVS with "`dnf --setopt=install_weak_deps=False openvswitch`" to avoid pulling it in. But I don't know if that's possible on RHEL7.

pecameron: @rcarrillocruz @danwinship How much is this change going to reduce the sdn image size? Are we going to look at OVN image size as well? Every image has a large almost common wedge in it. This increases Linux memory resource requirements. We should look at busybox and combine a lot of images into a common image with different entry points. The decision to make a lot of separate images has gotten us to this point.

danwinship: Container images are already designed in a way that allows reuse of shared files; this is why everything builds on top of the OCP base images. (`FROM registry.svc.ci.openshift.org/ocp/4.2:base`). "Making the SDN image smaller" is just about reducing what we add on top of the base image.

Yes, the upstream card covers reducing the OVN image size too. I had started playing around with that a while ago (https://github.com/openshift/ovn-kubernetes/pull/11) but didn't get very far.

pecameron: @danwinship part of this should be the before and after sizes.

PR #54, opened 2019-10-15
Remove watching Endpoints of Headless Services

rcarrillocruz:


pkg/network/proxy/proxy.go
```diff
@@ -312,49 +314,58 @@ func (proxy *OsdnProxy) OnEndpointsAdd(ep *corev1.Endpoints) {
 	proxy.Lock()
 	defer proxy.Unlock()
 
-	pep := &proxyEndpoints{ep, proxy.endpointsBlocked(ep)}
-	proxy.allEndpoints[ep.UID] = pep
-	if !pep.blocked {
-		proxy.baseProxy.OnEndpointsAdd(ep)
+	labels := ep.GetLabels()
+	if _, ok := labels[IsHeadLessService]; ok {
+		pep := &proxyEndpoints{ep, proxy.endpointsBlocked(ep)}
```
danwinship: Don't indent the whole function body; return early in the no-op case instead.

And the check is backwards anyway isn't it? (We want to process the endpoints if it *doesn't* have the annotation, and skip processing if it does.)

pkg/network/proxy/proxy.go
```diff
@@ -312,49 +314,58 @@ func (proxy *OsdnProxy) OnEndpointsAdd(ep *corev1.Endpoints) {
...
-
-	switch {
-	case wasBlocked && !pep.blocked:
-		proxy.baseProxy.OnEndpointsAdd(ep)
-	case !wasBlocked && !pep.blocked:
-		proxy.baseProxy.OnEndpointsUpdate(old, ep)
-	case !wasBlocked && pep.blocked:
-		proxy.baseProxy.OnEndpointsDelete(ep)
+	labels := ep.GetLabels()
+	if _, ok := labels[IsHeadLessService]; ok {
```
danwinship: maybe have a helper function

pkg/network/proxy/proxy.go
```diff
@@ -26,6 +26,8 @@ import (
 	"github.com/openshift/sdn/pkg/network/common"
 )
 
+const IsHeadLessService string = "service.kubernetes.io/headless"
```
danwinship: This is already exported upstream: `corev1.IsHeadlessService`

squeed:
The upstream implementation of this applies the label selector directly to the watch, which means the services aren't even sent down the wire. This is definitely a big improvement, and will help ups because of the large services created for monitoring (e.g. node-exporter, openshift-sdn, etc).

The problem is that our watches are shared between kube-proxy and the network policy code. Do you know if network policy also uses endpoints? If nothing besides kube-proxy watches endpoints, then it would be safe for us to implement this in the same manner as upstream.

squeed:
I did a quick scan of the codebase, and we don't use the Endpoint informer anywhere else. So I'd say it's safe to add the label selector directly.

danwinship:
Hm... I hadn't considered that doing the filter at the informer level would also improve performance.

So maybe it would be better to do it that way. OK, sorry, @rcarrillocruz, I screwed up. You did it right the first time...

PR #55, opened 2019-10-21
Add support for IPBlocks with except rules

sridhargaddam:
ipBlocks in NetworkPolicies allow us to specify CIDR ranges
as ingress sources or egress destinations. Generally these
are cluster-external IPs and will allow you to filter incoming
packets based on the source-ip.

When an ipBlock has an allowedCidr with except CIDRs, it implies
that traffic from the allowed CIDR excluding the IPs/CIDRs mentioned
in the except block should be allowed as part of the current ipBlock.
At the same time, except does not mean "drop", it just means that the
excepted IPs aren't allowed by this rule.

This PR implements ingress ipBlocks with except rules in NetworkPolicies.

Signed-off-by: Sridhar Gaddam <sgaddam@redhat.com>

### DIFF
--- a/pkg/network/node/networkpolicy.go
+++ b/pkg/network/node/networkpolicy.go
@@ -68,6 +68,12 @@ type npPolicy struct {
 
 	flows       []string
 	selectedIPs []string
+	ipBlocks    map[int]*ipBlockExceptFlows
+}
+
+type ipBlockExceptFlows struct {
+	allowed string
+	except  []string
 }
 
 // npCacheEntry caches information about matches for a LabelSelector
@@ -268,6 +274,8 @@ func (np *networkPolicyPlugin) syncNamespaceFlows(npns *npNamespace) {
 	klog.V(5).Infof("syncNamespace %d", npns.vnid)
 	otx := np.node.oc.NewTransaction()
 	otx.DeleteFlows("table=80, reg1=%d", npns.vnid)
+	otx.DeleteFlows("table=81, reg1=%d", npns.vnid)
+	otx.DeleteFlows("table=82, reg1=%d", npns.vnid)
 	if npns.inUse {
 		allPodsSelected := false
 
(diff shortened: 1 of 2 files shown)


sridhargaddam:
@danwinship, @dcbw : Appreciate your review when you find some time. Thanks.

sridhargaddam:
reviews please :)

pkg/network/node/ovscontroller.go
```diff
@@ -193,7 +193,7 @@ func (oc *ovsController) SetupOVS(clusterNetworkCIDR []string, serviceNetworkCID
 	// Table 80: IP policy enforcement; mostly managed by the osdnPolicy
 	otx.AddFlow("table=80, priority=300, ip, nw_src=%s/32, actions=output:NXM_NX_REG2[]", localSubnetGateway)
 	// eg, "table=80, priority=100, reg0=${tenant_id}, reg1=${tenant_id}, actions=output:NXM_NX_REG2[]"
-	otx.AddFlow("table=80, priority=0, actions=drop")
+	otx.AddFlow("table=82, priority=0, actions=drop")
```
juanluisvaladas, line 196: There is a comment on ovscontroller.go for every table. I  expect a comment for tables 81 and 82 too. Even if flows for for table 81 aren't created at all in this file.

pkg/network/node/networkpolicy.go
```diff
@@ -624,6 +698,10 @@ func (np *networkPolicyPlugin) handleDeleteNetworkPolicy(obj interface{}) {
 
 	if npns, exists := np.namespaces[vnid]; exists {
 		np.cleanupNetworkPolicy(policy)
+		for i := range npns.policies[policy.UID].ipBlocks {
```
juanluisvaladas, line 701: I would expect this to be part of cleanupNetworkPolicy.
cleanupNetworkPolicy is a private function and it's only used here, so changing the signature of the function shouldn't be a problem.

pkg/network/node/networkpolicy.go
```diff
@@ -281,6 +289,38 @@ func (np *networkPolicyPlugin) syncNamespaceFlows(npns *npNamespace) {
...
+		otx.AddFlow("table=80, priority=1, reg1=%d, actions=load:%d->NXM_NX_REG3[],goto_table:81", npns.vnid, ipBlockIndex)
+
+		for _, npp := range npns.policies {
+			for _, ipBlock := range npp.ipBlocks {
+				priority := 100
+				// Default low-priority rule for the current ipBlock set. This rule is hit if traffic does not match
+				// the current ipBlock. So, we reiterate to see if it matches some other ipBlock set.
+				otx.AddFlow("table=81, priority=%d, reg1=%d, reg3=%d, actions=load:%d->NXM_NX_REG3[],resubmit:81", priority, npns.vnid, ipBlockIndex, ipBlockIndex+1)
+				// Each flow in a given ipBlock has a unique priority
+				priority++
```
juanluisvaladas, line 303: We start with lower ipBlockIndex, and we evaluate ipBlocks from lower to higher ipBlockIndex, therefore I think the lowest ipBlockIndex should get the highest priority.

Also reading this is counter intuitive because the flows are defined in reverse order than they are processed.

Can we start with a very high priority and lower as we go through the loop?

danwinship:
/close
a different implementation of this feature already merged in 4.10

PR #56, opened 2019-10-21
Bug 1768436: UPSTREAM: 74840: kube-proxy: Drop packets in INVALID state

danwinship:
I initially tried to file this as https://github.com/openshift/kubernetes/pull/89 but that's failing CI for infrastructural reasons.

With openshift/origin, they mostly don't modify openshift/kubernetes itself after the initial creation of a new branch, and I think the repo isn't really set up to deal with us wanting to do that. (eg, the branches are protected but the OWNERS files aren't changed from upstream. And CI seems to be broken.). So maybe we should shift to having a clean(er) openshift/kubernetes branch and keeping (at least some) divergences here instead of there?

/cc @dcbw @squeed

### DIFF
--- a/glide.diff
+++ b/glide.diff
@@ -3,3 +3,19 @@ diff --no-dereference -N -r current/vendor/github.com/vishvananda/netlink/link_l
 < 			gre.FlowBased = true
 ---
 > 			gre.FlowBased = int8(datum.Value[0]) != 0
+diff --no-dereference -N -r current/vendor/k8s.io/kubernetes/pkg/proxy/iptables/proxier.go updated/vendor/k8s.io/kubernetes/pkg/proxy/iptables/proxier.go
+37c37
+< 	v1 "k8s.io/api/core/v1"
+---
+> 	"k8s.io/api/core/v1"
+1316,1325d1315
+< 
+< 	// Drop the packets in INVALID state, which would potentially cause
+< 	// unexpected connection reset.
+< 	// https://github.com/kubernetes/kubernetes/issues/74839
+< 	writeLine(proxier.filterRules,
+< 		"-A", string(kubeForwardChain),
+< 		"-m", "conntrack",
+< 		"--ctstate", "INVALID",
+< 		"-j", "DROP",
+< 	)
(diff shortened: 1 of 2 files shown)


squeed:
Huh. I wonder what the right way to go about this is. On one hand, I see that it's structurally difficult to change our kubernetes fork. OTOH, for patches that should persist across versions (carry patches, or patches that don't get merged quickly) we could easily lose track of them and suffer regressions.

danwinship:
For origin, all of the patches are in origin itself, not in openshift/kubernetes. You can just do `git log vendor/k8s.io/kubernetes` to see all of them...

eparis:
this is going to require a proper BZ and team review/lgtm/approval to move forward.

danwinship:
/assign @squeed 
I think we've kind of agreed that the plan is to patch kubernetes inside openshift/sdn rather than openshift/kubernetes now? (We'll have to update the rebase docs... and we can fix it to all be consistently done this way for the 4.4 rebase.)

squeed:
> I think we've kind of agreed that the plan is to patch kubernetes inside openshift/sdn rather than openshift/kubernetes now?

@danwinship From my conversations with @deads2k, it would be "better" if we put our patches in o/k if possible (especially if they're CARRY). I recall the problem with that being the lack of CI and OWNERS files issues.

danwinship:
> From my conversations with @deads2k, it would be "better" if we put our patches in o/k if possible (especially if they're CARRY).

So why doesn't origin do it that way?

> I recall the problem with that being the lack of CI and OWNERS files issues.

*lack* of CI would be an improvement. It has completely-broken CI.

squeed:
> So why doesn't origin do it that way?

Hah. Perhaps the conversation was aspirational.

> _lack_ of CI would be an improvement. It has completely-broken CI.

Well, at least there are no false positives.

I'm convinced. If we should be merging our fixes in to o/k a different way, @deads2k, speak now or forever hold your peace.

/lgtm

mfojtik:
ping @deads2k to reply to https://github.com/openshift/sdn/pull/56#issuecomment-552920626

holding cherry pick approval until he say the last words ;-)

PR #57, opened 2019-10-24
slightly improve logging

danwinship:
1. I don't think the NetID logging is ever especially useful
2. The absence of pod create/delete logging is frequently annoying

/hold until the tests run and I can look at the logs


pkg/network/node/pod.go
```diff
@@ -273,6 +271,7 @@ func (m *podManager) processRequest(request *cniserver.PodRequest) *cniserver.Po
 			}
 		}
 		if err != nil {
+			klog.Warningf("CNI_ADD %s/%s failed: %v", request.PodNamespace, request.PodName, err)
```
dcbw: @danwinship you can use `pk` here since that is a string that is already <namespace>/<name>

PR #58, opened 2019-10-24
Fix pod startup reattach/kill loop

danwinship:
On startup, if we destroy and recreate the OVS bridge, then we run through all the pods that were previously attached to the bridge, try to reattach them, and kill them if we fail. But this is slightly wrong; it means that if a pod is already detached from the bridge or the bridge is already destroyed when we restart, then we will neither reattach nor kill the detached pod(s). #52 is one example of when this could happen. It's possible that the (3.11) CI cluster ran into some example of this last night.

/hold until I carefully examine the test results


danwinship:
/hold cancel
/assign @alexanderConstantinescu 
since you worked with this code recently

alexanderConstantinescu:
/lgtm

I guess you know better than I do, but I assume this needs a backport all the way back to 3.11? https://github.com/openshift/origin/blob/release-3.11/pkg/network/node/node.go#L405

Conan-Kudo:
@danwinship @alexanderConstantinescu There's a PR for this to fix in 3.11 in openshift/origin#24318.

PR #59, opened 2019-10-28
Create a merged node/master image

danwinship:
I'd been meaning to do this since we migrated repos...

1. Simpler
2. Single image uses less space on the node
3. Makes it possible to override the sdn-controller image when using `./hack/run-locally.sh`, which isn't currently possible.
4. Fixes the historical oddity of the sdn-node image being named "node" rather than "sdn-node"

I also added a `Dockerfile.fedora` for doing test builds with, and added a `build-image-sdn-test` target to build it. I originally made it use the same `build-image` build macro that the rest of the Makefile is using, but that uses `imagebuilder`, which uses `docker`, which is just pointlessly difficult (and I can't even figure out how to get it to work... somehow imagebuilder is invoking docker in a way that nullifies my attempts to log in to `registry.svc.ci`).

(I'll remove the old images once this change has been plumbed through...)

/assign @squeed


squeed:
Seems reasonable. Thanks for looking in to this. BTW, can we just rename `Dockerfile.rhel` to `Dockerfile`? I don't think we'll ever need a centos-based build.

danwinship:
> BTW, can we just rename `Dockerfile.rhel` to `Dockerfile`?

That would require changes in `release` as well because it expects the current name. The current name is also consistent with other repos. But we might as well drop the CentOS `Dockerfile`s since you can't actually build them anyway...

images/sdn/Dockerfile.rhel
```diff
@@ -1,10 +1,11 @@
 FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.12 AS builder
 WORKDIR /go/src/github.com/openshift/sdn
 COPY . .
-RUN make build --warn-undefined-variables
+RUN CGO_ENABLED=0 make build --warn-undefined-variables
```
squeed: I think this is going to break the FIPS people, since openshift-sdn won't be using the hacked ssl.

squeed: Aha. We still have this problem in current openshift-sdn. I'll make sure the fips people know.

PR #64, opened 2019-11-03
client: Use protobuf for SDN clients

smarterclayton:


danwinship:
So does this work now? We had intentionally disabled protobuf support before because it didn't work with CRD types.

If it does work you probably need an update to `pkg/openshift-network-controller/` too; this patch will only fix sdn-node, not sdn-controller.

smarterclayton:
You should only have disabled protobuf for crd resources.  Right now 1/2 of CPU time in SDN is json decoding of pods and nodes, which is roughly a tenth of a core per machine.

The client should properly negotiate json watches for CRDs in 1.16 but I need to verify there are no writes being done.

smarterclayton:
The backport for 4.2 and 4.1 will use proto for core resources and json for SDN resources.

smarterclayton:
SDN controller isn’t showing up in profiles yet, I’m mostly trying to get per node cpu back under control.

smarterclayton:
If OVN has done the same thing we need to fix that as well (1.16 plus, should look like this, 1.16 minus should have proto for core resources)

pkg/openshift-sdn/proxy.go
```diff
@@ -215,13 +216,25 @@ func (sdn *OpenShiftSDN) runProxy(waitChan chan<- bool) {
...
+		sdn.ProxyConfig.MetricsBindAddress = "*:9665"
+	}
 	if len(sdn.ProxyConfig.MetricsBindAddress) > 0 {
 		mux := http.NewServeMux()
 		mux.HandleFunc("/proxyMode", func(w http.ResponseWriter, r *http.Request) {
 			fmt.Fprintf(w, "%s", sdn.ProxyConfig.Mode)
 		})
 		mux.Handle("/metrics", prometheus.Handler())
+
+		mux.HandleFunc("/debug/pprof/", pprof.Index)
```
squeed: This port is open to the whole cluster - is there an information leak here?

danwinship: note the "DO NOT MERGE" on that commit :slightly_smiling_face:

smarterclayton: :) yeah I’m trying to look at what else might be broken.  Protobuf is the cheap easy win, but I’m seeing a lot of time in GC

smarterclayton:
It looks like the biggest win is to not use the hybrid proxier if we don't use it - about 15-20% of CPU time is spent syncing the user space proxy rules.

It looks like most of the remaining time is in exec to iptables.

Removed the debugging commit, I think this is sufficient on its own for a small win.  PTAL

pkg/openshift-sdn/proxy.go
squeed: It's not (wasn't) on the PR :-)

squeed:
/approve
/lgtm

If you want to file a JIRA card, we can start looking in to ways to make the unidling proxy do nothing faster.

smarterclayton:
/retest

I guess with unidling we have to sort through what the OVN strategy will be.

danwinship:
> I guess with unidling we have to sort through what the OVN strategy will be.

OVN will do unidling natively, not with the hybrid proxy

PR #66, opened 2019-11-03
Add IPv6 support and multiple CIDR support to SubnetAllocator

danwinship:
This extends SubnetAllocator to support IPv6 ClusterNetworkCIDRs, and then also moves the code from OsdnMaster for handling multiple subnets and makes it part of SubnetAllocator instead.

(Of course, with stock OCP you can't configure an IPv6 ClusterNetwork so this should have no effect on end users at this point, and thus doesn't count as a feature, and would be exempt on IPv6 grounds even if it did.)

ovn-kubernetes currently vendors an older copy of SubnetAllocator (via `"github.com/openshift/origin/pkg/util/netutils"`) and then independently reimplements the multiple CIDR stuff on top of that. My intention was that it could be updated to either copy+paste or vendor this version instead, to provide IPv6 support there as well (and to remove the duplicated multiple CIDR code).

### DIFF
--- a/pkg/network/master/util/subnet_allocator.go
+++ b/pkg/network/master/util/subnet_allocator.go
@@ -1,4 +1,4 @@
-package master
+package util
 
 import (
 	"fmt"
--- a/pkg/network/master/master.go
+++ b/pkg/network/master/master.go
@@ -38,9 +38,7 @@ type OsdnMaster struct {
 	netNamespaceInformer networkinformers.NetNamespaceInformer
 
 	// Used for allocating subnets in order
-	subnetAllocatorList []*SubnetAllocator
-	// Used for clusterNetwork --> subnetAllocator lookup
-	subnetAllocatorMap map[common.ParsedClusterNetworkEntry]*SubnetAllocator
+	subnetAllocator *SubnetAllocator
 
 	// Holds Node IP used in creating host subnet for a node
 	hostSubnetNodeIPs map[ktypes.UID]string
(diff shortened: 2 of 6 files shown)


pkg/network/master/util/subnet_allocator.go
```diff
@@ -0,0 +1,204 @@
...
+	//
+	// To try to avoid this confusion, if the subnet extends into the next higher
+	// octet, we rotate the bits of the subnet number so that we use the subnets with
+	// all 0s in the shared octet first. So again given network="10.1.0.0/16",
+	// hostBits=6, we first allocate 10.1.0.0/26, 10.1.1.0/26, etc, through
+	// 10.1.255.0/26 (just like we would with /24s in the hostBits=8 case), and only
+	// if we use up all of those subnets do we start allocating 10.1.0.64/26,
+	// 10.1.1.64/26, etc.
+	//
+	// (For IPv6 we just don't bother worrying about this.)
```
markmc: Good call

pkg/network/master/util/subnet_allocator.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+// allocateNetwork returns a new subnet, or nil if the range is full
+func (snr *subnetAllocatorRange) allocateNetwork() *net.IPNet {
+	netMaskSize, addrLen := snr.network.Mask.Size()
+	numSubnets := uint32(1) << snr.subnetBits
+	if snr.subnetBits > 24 {
+		// If snr.subnetBits > 32, then numSubnets will have overflowed. (And if
+		// snr.subnetBits + snr.hostBits%8 >= 32 then the math below will
+		// eventually overflow.) "16M subnets ought to be enough for anybody."
```
markmc: Hmm, I can't convince myself not to worry about "if snr.subnetBits + snr.hostBits%8 >= 32 then the math below will eventually overflow". Could you be more explicit?

danwinship:
updated to skip the 0 subnet on IPv6, because it's weird that pod IPs on the first node look like "fd01::P", but pod IPs on all other nodes look like "fd01:0:0:S::P"

PR #70, opened 2019-11-13
Additional event logging for add/delete/update informer events

abhat:


pkg/openshift-sdn/informers.go
```diff
@@ -121,3 +125,31 @@ func applyClientConnectionOverrides(kubeConfig *rest.Config) {
...
+type eventHandlingLogger struct{}
+
+func describe(obj interface{}) string {
+	if svc, ok := obj.(*v1.Service); ok {
+		if svc.Spec.Type == v1.ServiceTypeExternalName {
+			return fmt.Sprintf("externalname service %s/%s", svc.Namespace, svc.Name)
+		} else if svc.Spec.ClusterIP == v1.ClusterIPNone {
+			return fmt.Sprintf("headless service %s/%s", svc.Namespace, svc.Name)
+		} else {
+			return fmt.Sprintf("service %s/%s", svc.Namespace, svc.Name)
```
alexanderConstantinescu: What do you think about logging the entire object? We might realize later on that a couple of attributes are not enough to debug the issue completely?

abhat: Sure, that makes sense. let me update the PR.

squeed: yeah, agreed. I'd serialize the service object to json and log it.

abhat: does it need serialization, I am logging it as `%+v`?

PR #71, opened 2019-11-14
[WIP] stab in the dark logging to get to the bottom of endpoint issues

squeed:
An attempt to get to the bottom of https://bugzilla.redhat.com/show_bug.cgi?id=1765280

### DIFF
--- a/pkg/openshift-sdn/informers.go
+++ b/pkg/openshift-sdn/informers.go
@@ -1,10 +1,13 @@
 package openshift_sdn
 
 import (
+	"fmt"
 	"net"
 	"net/http"
 	"time"
 
+	networkclient "github.com/openshift/client-go/network/clientset/versioned"
+	networkinformers "github.com/openshift/client-go/network/informers/externalversions"
 	v1 "k8s.io/api/core/v1"
 	v1meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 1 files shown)


squeed:
Need https://github.com/openshift/release/pull/5918 to merge so we can e2e-gcp

alexanderConstantinescu:
Hey, @squeed could you also log the value of the variable `hasEndpoints` that originates from 

```
hasEndpoints := len(proxier.endpointsMap[svcName]) > 0
```

I can't rap my head around what the hell is going with this thing, during and between all, syncs.

squeed:
Darn, this change with EVEN MORE LOGGING seems to be running correctly!

squeed:
of course, with all the extra logging (and no code changes), the tests pass. Hmmm.

/test e2e-gcp

squeed:
Yet another test run where everything comes up correctly.

squeed:
green again!

/test e2e-gcp

pecameron:
You have an interesting race here. The logging is slowing down or synchronizing the problem enough to lose the race. Maybe turn off all extra logging, just errors and maybe it will mostly fail. Aside from taking extra time, logging can serialize operations enough to get past races.

squeed:
@pecameron yeah, I did consider that. Most of the delay in logging is in serializing the structs, so the most recent set of pushes drops all of that and adds just a few small logs. But yeah, ultimately, logging is a serialization point. 

The thing is, the race I'm considering is in the unidling proxy, which is already behind a lock (`usingUserspaceLock`) - so we're already serializing endpoint + service changes.

danwinship:
> The thing is, the race I'm considering is in the unidling proxy, which is already behind a lock (`usingUserspaceLock`) - so we're already serializing endpoint + service changes.

Maybe the wrong thing happens when events `A` and `B` are delivered separately but the right thing happens when they get coalesced into event `B'` by the informer?

squeed:
We have two handlers attached to the informer - the proxy, and a separate debugger. For whatever reason, the proxy handler is not seeing the only transition we're interested in (No LB IP - > LB IP). But the debug handelr (which runs *after* the proxy hander) does.

It happens to be the last transition, so all subsequent updates (which **are** seen) are short-circuited since (A = A'). 

One thing we could do is compare A' against the known state, not against A. But we should never be missing these things.

squeed:
Just checked the CI cluster, this run will pass too. So annoying.

pecameron:
@squeed Clearly something is different with the logging on. Working with that may back you into the problem.

PR #73, opened 2019-11-15
Bug 1772905: [4.2] Clean up egress IPs on startup

danwinship:
backport of #51

### DIFF
--- a/pkg/network/node/egressip.go
+++ b/pkg/network/node/egressip.go
@@ -5,7 +5,6 @@ package node
 import (
 	"fmt"
 	"os/exec"
-	"sync"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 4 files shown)


pkg/network/node/egressip.go
```diff
@@ -61,6 +60,45 @@ func (eip *egressIPWatcher) Start(networkInformers networkinformers.SharedInform
...
+		return
+	}
+
+	for _, addr := range addrs {
+		ip := addr.IP.String()
+		if addr.Label == label && eip.iptablesMark[ip] == "" {
+			klog.Infof("Cleaning up stale egress IP %s", addr.IP.String())
+			err = netlink.AddrDel(link, &addr)
+			if err != nil {
+				utilruntime.HandleError(fmt.Errorf("Could not clean up stale egress IP: %v", err))
```
pecameron: Should also include the IP that is causing the problem.

danwinship: odd... I could have sworn I already responded to this.

Anyway, it already logged the IP a few lines earlier, so it will be clear. (Also, this is a backport so we don't want it to be different from the version in master anyway.)

PR #77, opened 2019-12-05
node: log pod operation errors

dcbw:
Otherwise they only show up in Multus/CRIO logs. We should have them
in the SDN pod logs too, if only to allow correlating failure
timelines more easily.

@squeed @danwinship @smarterclayton


danwinship:
/lgtm cancel
i think #57 is a superset of this

dcbw:
Closign in favor of https://github.com/openshift/sdn/pull/57

PR #79, opened 2019-12-06
proxy: add handler with same ResyncPeriod as shared informer.

squeed:
Works around a nasty bug where having multiple handlers with non-homogeneous resync times causes updates during connection disruption to be missed by one or more handlers.


danwinship:
ok
/lgtm

PR #84, opened 2019-12-12
Bug 1782847: sdn: undo debug logging

squeed:
It was added to track down the informer issue; we caught it. It causes logs to be way too large.

### DIFF
--- a/pkg/openshift-sdn/informers.go
+++ b/pkg/openshift-sdn/informers.go
@@ -1,7 +1,6 @@
 package openshift_sdn
 
 import (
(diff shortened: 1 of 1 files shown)


squeed:
I want to backport this to 4.3.0 before it releases. Thus, this is pretty important. @juanluisvaladas can you review?

juanluisvaladas:
/lgtm
Commit with adding the additional logging: https://github.com/openshift/sdn/commit43452fe37f51a41ea1d56dd66dda9ae1d10bbda7

PR #91, opened 2020-01-15
Bug 1762580: Add OPENSHIFT-PREROUTING chain

juanluisvaladas:
Fixes BZ#1762580. When using egressIP the packets going to
externalIP services are matched by KUBE-SERVICES rules, this means
the packets destination IP is modified to the endpoint IP address.

By adding this chain and a few rules on it, we guarantee the
packets marked for egressIP will skip the KUBE-SERVICES chains,
hence, the traffic to externalIP is sent to the externalIP rather
than an endpoint.

### DIFF
--- a/pkg/network/node/iptables.go
+++ b/pkg/network/node/iptables.go
@@ -57,6 +57,14 @@ func (n *NodeIPTables) Setup() error {
 	return nil
 }
 
+// FinishSetup performs bootstrap action that must be made after kube-proxy
+// initialization
+func (n *NodeIPTables) FinishSetup() error {
+	n.mu.Lock()
+	defer n.mu.Unlock()
+	return n.syncIPTableChains(n.getNodeIPTablesPreChains())
+}
+
 // syncLoop periodically calls syncIPTableRules().
 // This is expected to run as a go routine or as the main loop. It does not return.
 func (n *NodeIPTables) syncLoop() {
@@ -95,19 +103,7 @@ func (n *NodeIPTables) addChainRules(chain Chain) (bool, error) {
 	return allExisted, nil
 }
 
-// syncIPTableRules syncs the cluster network cidr iptables rules.
-// Called from SyncLoop() or firewalld reload()
-func (n *NodeIPTables) syncIPTableRules() error {
-	n.mu.Lock()
-	defer n.mu.Unlock()
-
-	start := time.Now()
-	defer func() {
-		klog.V(4).Infof("syncIPTableRules took %v", time.Since(start))
-	}()
-	klog.V(3).Infof("Syncing openshift iptables rules")
-
-	chains := n.getNodeIPTablesChains()
+func (n *NodeIPTables) syncIPTableChains(chains []Chain) error {
 	for i := len(chains) - 1; i >= 0; i-- {
 		chain := chains[i]
 		// Create chain if it does not already exist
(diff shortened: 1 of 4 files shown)


juanluisvaladas:
@danwinship can you please review?

pkg/network/node/iptables.go
```diff
@@ -57,6 +57,14 @@ func (n *NodeIPTables) Setup() error {
 	return nil
 }
 
+// FinishSetup performs bootstrap action that must be made after kube-proxy
+// initialization
```
danwinship: This won't work. Kube-proxy may recreate its rules later, in various circumstances. There's no way to reliably guarantee that your rule gets inserted before kube-proxy's rule.

You may be able to prepend a rule to the beginning of kube-proxy's chain though, assuming it's not one of the chains that gets recreated from scratch on each proxy sync?

juanluisvaladas: Does that include the rules in the PREROUTING chain? I see children of KUBE-SERVICES: 
https://github.com/openshift/sdn/blob/master/vendor/k8s.io/kubernetes/pkg/proxy/iptables/proxier.go#L368-L403

However, I can't find anywhere where KUBE-SERVICES  is ever deleted outside the function CleanupLeftovers.

danwinship: The exact details of the code may change in the future. We've learned that you should not try to depend on a particular rule always ending up before or after kube's rules, or it will eventually break.

pkg/network/node/iptables.go
```diff
@@ -154,8 +169,28 @@ func (n *NodeIPTables) syncIPTableRules() error {
...
+func (n *NodeIPTables) getNodeIPTablesPreChains() []Chain {
+	var chainArray []Chain
+	chainArray = append(chainArray,
+		Chain{
+			table:    "nat",
+			name:     "OPENSHIFT-PREROUTING",
+			srcChain: "PREROUTING",
+			srcRule:  []string{"-m", "comment", "--comment", "rules for prerouting OpenShift traffic"},
+			rules: [][]string{
+				{"-m", "mark", "--mark", "0x1/0x1", "-m", "comment", "--comment", "KUBE-MARK-MASQ pass", "-j", "RETURN"},
```
danwinship: You can't hardcode this; it's configurable. `n.masqueradeBitHex`

danwinship:
note that if we're changing `pkg/network/node/iptables.go` anyway, then maybe it makes sense to change the existing rules? Or add another per-egress-IP rule somewhere? (I don't remember the exact details of where things are going wrong.)

juanluisvaladas:
> Or add another per-egress-IP rule somewhere? (I don't remember the exact details of where things are going wrong.)
The issue is well documented here in https://bugzilla.redhat.com/show_bug.cgi?id=1762580#c29 . But here's a quick summary:
Assuming pod is node A and the egressIP belongs to nodeB:
1- Packet leaves node A to node B encapsulated
2- Packet arrives to the br0 in node B
3- Packet matches these flows (skipping the ones without relevant actions):
```
 cookie=0x0, duration=759.329s, table=100, n_packets=0, n_bytes=0, priority=100,ip,reg0=0x489109 actions=set_field:92:e9:65:f6:5a:92->eth_dst,set_field:0x1489108->pkt_mark,goto_table:101
 cookie=0x0, duration=967.344s, table=101, n_packets=13, n_bytes=962, priority=0 actions=output:tun0
```
4- Packets leave the switch and get to nat PREROUTING
5- In PREROUTING they eventually get to KUBE-SERVICES
6- IN KUBE-SERVICES they eventally hit a rule of KUBE-SVC-XXXXXXX because kube-proxy intercepts packets sent to externalIPs. Desitnation IP is replaced with service IP

If we can't rely on the order of the rules evaluation, then I don't think it can be fixed in iptables. The only table that comes first is raw, and to the best of my knowledge we can't skip nat's PREROUTING from it.

The only solution that I can think of (and I don't like it) is adding the egressIP to the br0 NIC (not sure if feasible) and send the packet directly from ovs so that we skip iptables. I'm kind of comfortable with taking the packet out, but receiving the response back sounds problematic. Plus I'm not quite sure how would that affect the rest of the traffic. The whole idea seems just bad.

danwinship:
commented in bugzilla
/hold

juanluisvaladas:
This is a dead end...

PR #92, opened 2020-01-17
belatedly update openshift deps to release-4.3 branches

danwinship:
When we first rebased to kube 1.16, there were no release-4.3 branches in client-go, library-go, and api, so we used the `prebase-1.16` branches that had been created for the origin rebase. Unbeknownst to us, these were considered temporary and recently went away. Fix by updating to the correct branches. Which pulls in a bunch of updates...

(Yes, we will be re-rebasing master soon, but this change will need to be backported to our release-4.3.)


danwinship:
closing in favor of #93

danwinship:
oops, i said I was closing this but then didn't

PR #95, opened 2020-01-20
Bug 1794029: [4.2] Use k8s.io/client-go fork with informer fix, run update-deps

danwinship:
4.2 version of #93. Also includes belated glide.yaml fixup for 4.2 deps...


ashcrow:
There are a lot of differences between the original and this backport. However, I believe this to be due to the amount of change and enhancements that have occurred between 4.2 and master.

danwinship:
> There are a lot of differences between the original and this backport. However, I believe this to be due to the amount of change and enhancements that have occurred between 4.2 and master.

It's not 4.2 vs master, it's just that we previously had basically 4.2.0 (or maybe even 4.2 alpha or something), and now we have something 4.2.z-ish with various bugfixes.

PR #96, opened 2020-01-20
Bug 1790440: Fix reinitialization of deny-all NetworkPolicy state on restart

juanluisvaladas:
If a project has only a deny-all rule and pods running on it,
npNameSpace.inUse was set to false, this commit sets it to true.

If inUse is set to false, the node will not update the
npNamespace if new rules are created until new pods are created.

More details on BZ#1790440

### DIFF
--- a/pkg/network/node/ovscontroller.go
+++ b/pkg/network/node/ovscontroller.go
@@ -672,9 +672,11 @@ func (oc *ovsController) UpdateVXLANMulticastFlows(remoteIPs []string) error {
 
 // FindPolicyVNIDs returns the set of VNIDs for which there are currently "policy" rules
 // in OVS. (This is used to reinitialize the osdnPolicy after a restart.)
+// We also include inUseVNIDs because if a namespace has only a deny all rule
+// policyVNIDs won't include that namespace.
 func (oc *ovsController) FindPolicyVNIDs() sets.Int {
-	_, policyVNIDs := oc.findInUseAndPolicyVNIDs()
-	return policyVNIDs
+	inUseVNIDs, policyVNIDs := oc.findInUseAndPolicyVNIDs()
+	return inUseVNIDs.Union(policyVNIDs)
 }
 
 // FindUnusedVNIDs returns a list of VNIDs for which there are table 80 "policy" rules,
--- a/pkg/network/node/ovscontroller_test.go
+++ b/pkg/network/node/ovscontroller_test.go
@@ -15,6 +15,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	"github.com/containernetworking/plugins/pkg/utils/hwaddr"
 )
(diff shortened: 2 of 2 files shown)


juanluisvaladas:
/retitle Bugzilla 1790440:  Fix reinitialization of deny-all NetworkPolicy state on restart

juanluisvaladas:
/retitle Bug 1790440: Fix reinitialization of deny-all NetworkPolicy state on restart

pkg/network/node/ovscontroller.go
```diff
@@ -673,8 +673,10 @@ func (oc *ovsController) UpdateVXLANMulticastFlows(remoteIPs []string) error {
 // FindPolicyVNIDs returns the set of VNIDs for which there are currently "policy" rules
 // in OVS. (This is used to reinitialize the osdnPolicy after a restart.)
 func (oc *ovsController) FindPolicyVNIDs() sets.Int {
-	_, policyVNIDs := oc.findInUseAndPolicyVNIDs()
-	return policyVNIDs
+	inUseVNIDs, policyVNIDs := oc.findInUseAndPolicyVNIDs()
+	// we need to incldue inUseVNIDs as well because if a namespace has only a deny all
+	// rule, policyVNIDs won't include it.
```
danwinship: can you remove this comment and merge it with the description of the function above (which is not incorrect with your change)

juanluisvaladas: Should be fixed now.

pkg/network/node/ovscontroller.go
```diff
@@ -672,9 +672,11 @@ func (oc *ovsController) UpdateVXLANMulticastFlows(remoteIPs []string) error {
 
 // FindPolicyVNIDs returns the set of VNIDs for which there are currently "policy" rules
 // in OVS. (This is used to reinitialize the osdnPolicy after a restart.)
+// We also include inUseVNIDs because if a namespace has only a deny all rule
+// policyVNIDs won't include that namespace.
 func (oc *ovsController) FindPolicyVNIDs() sets.Int {
-	_, policyVNIDs := oc.findInUseAndPolicyVNIDs()
-	return policyVNIDs
+	inUseVNIDs, policyVNIDs := oc.findInUseAndPolicyVNIDs()
+	return inUseVNIDs.Union(policyVNIDs.Difference(inUseVNIDs))
```
danwinship: Oops, should have noticed this before: the `.Difference(inUseVNIDs)` there has no effect on the result. You want just `inUseVNIDs.Union(policyVNIDs)`.

pkg/network/node/ovscontroller_test.go
```diff
@@ -968,6 +965,79 @@ func TestFindUnusedVNIDs(t *testing.T) {
...
+		for _, flow := range tc.flows {
+			otx.AddFlow(flow)
+		}
+		if err := otx.Commit(); err != nil {
+			t.Fatalf("(%d) unexpected error from AddFlow: %v", i, err)
+		}
+
+		policyVNIDs := oc.FindPolicyVNIDs()
+
+		if policyVNIDs.Difference(tc.policyVNIDs).Len() > 0 || tc.policyVNIDs.Difference(policyVNIDs).Len() > 0 {
```
danwinship: `if !policyVNIDs.Equal(tc.policyVNIDs)`

pkg/network/node/ovscontroller.go
juanluisvaladas: Fixed, thanks

pkg/network/node/ovscontroller_test.go
juanluisvaladas: D'oh! fixed

PR #97, opened 2020-01-21
Add rcarrillocruz as approver

rcarrillocruz:
This is to improve throughput of merges during EMEA TZ


danwinship:
1) alphabetize!
2) please add @juanluisvaladas to `reviewers` while you're here
3) likewise for the CNO PR, and can you also submit one to openshift/ovn-kubernetes?

danwinship:
oh, and don't bother /retest'ing, we can override the tests

PR #102, opened 2020-02-04
Bug 1801357: Migrate from glide to go modules

juanluisvaladas:
Migrate from glide to go modules
  
It was mandatory to migrate to go modules due to syntax incompatibility
on the sigs.k8s.io-structured-merge-diff module definition. Besides
this incompatibility, it was a matter of time until we found more
problems and glide is abandoned at this stage.

There are a few non obvious things going on:

* go.mod: Unnecessary large, however, to avoid changing dependencies
          some stuff it was necessary to add some stuff that normally
          wouldn't make sense. After the 1.17 rebase it'll be cleaner.
          There is a require and replace discrepancy on
          github.com/miekg/dns. This is because go mod tidy modifies
          the require because one dependency asks for a newer version.
          The k8s.io/ packages are replaced beacuse this issue:
          https://github.com/golang/go/issues/26366

* pkg/dependencymagnet: go mod vendor won't actually pull a module
          unless it's required by some go code. We need to explictly
          add dependencies on a fake moduleso that we can actually
          build.

* Removed files: go mod vendor removes unnecessary files, lots of
          files have been removed, but those shouldn't be required at
          all.

* Updated github.com/openshift/library-go: The version we were using
          was incompatible with go mod. So I updated to the latest.

* Makefile: export GO111MODULE=on library go runs some stuff with
          -mod=vendor. Prior to go 1.13 this environment variable
          is mandatory for it to work.
          Additionally we need to explictly define
          GO_BUILD_PACKAGES_EXPANDED for the same reason.
          we need to add unxeport GOPATH and export GOPROXY are
          necessary due to some odd behavior with go get modernc.org/*. 
          The queries to this domain ultimately get converted to
          https://gitlab.com/cznic/* and go tries to run
          git clone https://gitlab.com/cznic/*
          instead of
          git clone https://gitlab.com/cznic/*.git.
          For some reason adding these two variables fixes it.

* patch directory: We needed to patch a single file in the netlink
          library. This is the most similar solution to glide.diff I
          could find. It's not elegant and doesn't scale well, but 
          since we are only patching one file in one library it's
          not so terrible.

* HACKING.md: This document is now outdated, it wil be updated after
          the 1.17 rebase.


juanluisvaladas:
/hold

This is missing the glide.diff fix, I want to see if it passess the tests first and I'll deal with that later.

juanluisvaladas:
/test verify-deps

Doesn't look like a legit failure

juanluisvaladas:
I'm particularly concerned about the patch, directory, I'm quite unhappy about this solution...

juanluisvaladas:
/hold cancel
This has to work now, there is no way make verify-deps fails again...

juanluisvaladas:
/hold
There is a conflict with https://github.com/openshift/sdn/pull/39 which must be addressed before merging this

juanluisvaladas:
I guess we can remove that, I removed all non go files and tests. Needs to pass tests again, particularly the make verify-deps, but it should be fine.

danwinship:
>     * patch directory: We needed to patch a single file in the netlink
>       library.

You can just update the vishvananda/netlink dep to `v1.1.0`. At the time we first took the fix we didn't want to update to a random git master SHA because we might have ended up getting new bugs along with the bugfix, but now there's been an official release containing the fix we need, so we can just take that.

Other than that, I think this seems OK? Ugh, vendoring. Sigh.

juanluisvaladas:
Regarding upgrading the netlink library, kube-proxy has github.com/docker/libnetwork/ipvs as a dependency and this has this lines:
https://github.com/docker/libnetwork/blob/master/ipvs/netlink.go#L220
Which is incompatible with v1.1.0 because now this function signature has changed and returns 3 arguments.
https://github.com/vishvananda/netlink/blob/v1.1.0/nl/nl_linux.go#L625

I could try to upgrade github.com/docker/libnetwork upstream to v1.1.0 but even if they accepted it, that would take some time.

danwinship:
blah. ok

btw, as seen in the github status, this needs `bugzilla/valid-bug`, meaning you need to create a bug targeted at 4.4 to go with this

juanluisvaladas:
/retitle Bug 1801357: Migrate from glide to go modules
/hold cancel

juanluisvaladas:
@danwinship can you approve it?

PR #105, opened 2020-02-06
Bug 1793952: [4.2] Fix reinitialization of deny-all NetworkPolicy state on restart

danwinship:
backport of #96 and #103 
/cc @juanluisvaladas


ashcrow:
Both backports have been verified in their upstream bugs.

PR #108, opened 2020-02-11
Bug 1803149: Rebase SDN to kubernetes 1.17.2

juanluisvaladas:
* Merged  the contents of https://github.com/openshift/sdn/pull/39/ into master

* Rebase SDN to Kubernetes 1.17.2

* Replaced go 1.12 for 1.13 in images.

* Removed unnecessary dependencies: During the migration to go modules
  a bunch of dependencies were defined explicitly so that we had an
  exact equivalent.

* OsdnProxy: Added OnNodeXxx functions so that it's compatible with
  service topology.

* HybridProxy: Added empty OnNodeXxx functions so that it implements
  the proxier interface. This proxy is not compatible with service
  topology.


juanluisvaladas:
Needs a review: The few things I didn't mention earlier:

We need https://github.com/openshift/release/pull/7121 merged before make verify-deps can pass. This happens because `go mod vendor` produces different results in go 1.12 and go 1.13

This PR includes a commit which doesn't compile: 3b71b6bba38c1eeb0b92155ee97d20e9e9b9b199
Because according to the conversation today having every commit compilable is preferred but not mandatory I preferred preserving git blame over having everything compiling

I sneaked a exception for the directory .vscode in .gitignore

juanluisvaladas:
Verify-deps passed with go 1.13 which means after this is merged stuff will be mergeable again

juanluisvaladas:
/retitle Bug 1803149: Rebase SDN to kubernetes 1.17.2

PR #110, opened 2020-02-21
Bug 1806648: Fix pod startup reattach/kill loop

abhat:
We should kill every (non-hostNetwork) sandbox that we can't reattach
to OVS, but previously we were only killing "every (non-hostNetwork)
sandbox that we knew was previously attached to OVS that we couldn't
reattach to OVS".

### DIFF
--- a/pkg/network/node/runtime.go
+++ b/pkg/network/node/runtime.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"time"
 
+	"k8s.io/klog"
+
 	kwait "k8s.io/apimachinery/pkg/util/wait"
 	kubeletapi "k8s.io/kubernetes/pkg/kubelet/apis/cri"
 	kruntimeapi "k8s.io/kubernetes/pkg/kubelet/apis/cri/runtime/v1alpha2"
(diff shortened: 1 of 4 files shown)


pkg/network/node/runtime.go
```diff
@@ -85,6 +87,17 @@ func (node *OsdnNode) getPodSandboxes() (map[string]*kruntimeapi.PodSandbox, err
 
 	podSandboxMap := make(map[string]*kruntimeapi.PodSandbox)
 	for _, sandbox := range podSandboxList {
+		status, err := runtimeService.PodSandboxStatus(sandbox.Id)
```
dcbw: PodSandboxStatus() might end up calling the CNI plugin (eg, openshift-sdn again) and if the SDN isn't running yet could return an error. Do we really want to ignore the sandbox in that case?

danwinship: > PodSandboxStatus() might end up calling the CNI plugin again

No, `PodSandboxStatus()` does not mean "do a CNI `CHECK`". It just gets the runtime's current data for the pod.

danwinship:
@abhat when backporting a patch please link to the original PR as well

danwinship:
also, 1792164 is the bug for 4.3; you'll need a new clone of it for 4.2

danwinship:
This backports #58 but it needs to include #52 as well.

abhat:
Cherry-picked commit from #52

abhat:
@danwinship can you approve please?

danwinship:
/lgtm
sorry, I actually reviewed this earlier today but I guess I forgot to actually "/lgtm" it...

PR #112, opened 2020-02-26
[release 4.2] Bug 1801742: Multiple iptables improvements

juanluisvaladas:
This backport has a few conflicts on each commit that that must be resolved manually.
The commit https://github.com/openshift/sdn/pull/112/commits/7f6e36fa477f5654cbc2102dee68ca6226e1a555 is not part of the original backport, I added it because there were a few conflicts and adding it seemed the easiest way to solve them


danwinship:
> This backport has a few conflicts on each commit that that must be resolved manually.

Hm... look into what commits it's conflicting with; it may indicate something else we need to add to the backport

juanluisvaladas:
@danwinship these changes are quite big so most certainly all kube-proxy will need to be revalidated. I will do a single backport for:
  - https://github.com/kubernetes/kubernetes/pull/80368
  - https://github.com/kubernetes/kubernetes/pull/82602
  - https://github.com/openshift/sdn/pull/39
Instead of 3 separate ones so that QA only have to validate all this stuff once, if that's OK with you.

juanluisvaladas:
/retitle [release 4.2] Bug 1801742: Multiple iptables improvements

juanluisvaladas:
@danwinship can you please review?

danwinship:
/hold
found some problems when it hit 4.3, we'll need 2 more patches here

danwinship:
Drop 78547; the patchset is big enough without pulling in unrelated fixes.
Also, you need to add back the e2e test from 81517.

This will also need the fixes from https://github.com/openshift/sdn/pull/126 and https://github.com/openshift/sdn/pull/127 once that's resolved.

juanluisvaladas:
/retitle [WIP][release 4.2] Bug 1801742: Multiple iptables improvements
Back to WIP becuase waiting for #127 to get merged.

juanluisvaladas:
@danwinship also added here the e2e-test changes from  [danwinship@c50267c](https://github.com/danwinship/origin/commit/c50267cf)

danwinship:
This is missing two of the commits from #39 ("`UPSTREAM: <drop>: preserve old health/metrics behavior`" and "`UPSTREAM: 83402: iptables.Monitor: don't be fooled by "could not get lock" errors`")

danwinship:
> This is missing two of the commits from #39 ("`UPSTREAM: <drop>: preserve old health/metrics behavior`" and "`UPSTREAM: 83402: iptables.Monitor: don't be fooled by "could not get lock" errors`")

@juanluisvaladas ping!

juanluisvaladas:
@JacobTanenbaum @danwinship sorry for the delay. All issues addressed.
/test e2e-aws

danwinship:
e2e-aws appears to be currently broken on 4.2
https://bugzilla.redhat.com/show_bug.cgi?id=1825372

knobunc:
Pulled the cherry-pick label since this is consistently failing tests.  @juanluisvaladas PTAL and we may be able to make next week's z-stream.

dcbw:
@juanluisvaladas something is odd here; the SDN pods are terminating:

```
Jun 16 11:12:01.140: INFO: Running AfterSuite actions on all nodes
Jun 16 11:12:01.140: INFO: Running AfterSuite actions on node 1
fail [github.com/openshift/origin/test/extended/operators/cluster.go:118]: Expected
    <[]string | len:1, cap:1>: [
        "Pod openshift-sdn/sdn-2qvng is not healthy: container sdn has restarted more than 5 times",
    ]
to be empty
```

```
Jun 16 11:12:01.133: INFO: At 2020-06-16 10:24:55 +0000 UTC - event for sdn-wfnb4: {kubelet ip-10-0-154-51.us-east-2.compute.internal} Created: Created container sdn
Jun 16 11:12:01.133: INFO: At 2020-06-16 10:24:55 +0000 UTC - event for sdn-wfnb4: {kubelet ip-10-0-154-51.us-east-2.compute.internal} Started: Started container sdn
Jun 16 11:12:01.133: INFO: At 2020-06-16 10:32:33 +0000 UTC - event for sdn-2qvng: {kubelet ip-10-0-144-137.us-east-2.compute.internal} Unhealthy: Liveness probe failed: HTTP probe failed with statuscode: 503
Jun 16 11:12:01.133: INFO: At 2020-06-16 10:32:36 +0000 UTC - event for sdn-mg8f6: {kubelet ip-10-0-136-216.us-east-2.compute.internal} Unhealthy: Liveness probe failed: HTTP probe failed with statuscode: 503
Jun 16 11:12:01.133: INFO: At 2020-06-16 10:32:37 +0000 UTC - event for sdn-cj6mr: {kubelet ip-10-0-138-246.us-east-2.compute.internal} Unhealthy: Liveness probe failed: HTTP probe failed with statuscode: 503
Jun 16 11:12:01.133: INFO: At 2020-06-16 10:32:39 +0000 UTC - event for sdn-5xnnw: {kubelet ip-10-0-133-22.us-east-2.compute.internal} Unhealthy: Liveness probe failed: HTTP probe failed with statuscode: 503
Jun 16 11:12:01.134: INFO: At 2020-06-16 10:32:40 +0000 UTC - event for sdn-rxndq: {kubelet ip-10-0-138-183.us-east-2.compute.internal} Unhealthy: Liveness probe failed: HTTP probe failed with statuscode: 503
Jun 16 11:12:01.134: INFO: At 2020-06-16 10:32:40 +0000 UTC - event for sdn-wfnb4: {kubelet ip-10-0-154-51.us-east-2.compute.internal} Unhealthy: Liveness probe failed: HTTP probe failed with statuscode: 503
Jun 16 11:12:01.134: INFO: At 2020-06-16 10:41:54 +0000 UTC - event for sdn-2qvng: {kubelet ip-10-0-144-137.us-east-2.compute.internal} BackOff: Back-off restarting failed container
```

danwinship:
4.2 has a liveness probe for the sdn pod based on kube-proxy /healthz. So it looks like probably that's failing and kubelet is killing it. Not sure why though; it looks like the proxy-health-related bits of the PR are correct...

danwinship:
So, yes, sdn is being killed because the liveness probe is failing because, apparently, `proxier.healthzServer.UpdateTimestamp()` is not being called every `syncPeriod`, for some reason that I can't figure out with the default debug level.

juanluisvaladas:
Checking...

bparees:
@danwinship @juanluisvaladas 
this will need to be unheld today if the intent is to get it into this week's 4.2.z

juanluisvaladas:
I'll work on it, thanks.

vendor/k8s.io/kubernetes/pkg/proxy/iptables/proxier.go
```diff
@@ -1366,6 +1366,11 @@ func (proxier *Proxier) syncProxyRules(force bool) {
 		return
 	}
 	proxier.changesPending = false
+	// If there are no pending changes, update the healthz server timestamp
+	if proxier.healthzServer != nil {
+		proxier.healthzServer.UpdateTimestamp()
+	}
+	metrics.SyncProxyRulesLastTimestamp.SetToCurrentTime()
```
danwinship, line 1373: This doesn't seem like it should have any effect; we're already doing exactly this 20 lines further down, and there are no `return`s between here and there...

juanluisvaladas:
/test e2e-aws
just triggering the test to get the image..

danwinship:
> `[UPSTREAM 83387: If an iptables proxier sync fails, retry after iptablesSyncPeriod]`

This should not be necessary given `UPSTREAM: <drop>: preserve old health/metrics behavior`. In git master, we pass `time.Hour` as the `maxInterval` to `NewBoundedFrequencyRunner`, so if an iptables update fails, and we didn't call `RetryAfter(syncPeriod)`, then it might be an hour before we tried again. But in 4.2, with the "preserve old behavior" commit, we are passing `syncPeriod` as the `maxInterval`, so after a failed iptables update, we'll try again after `syncPeriod` already.

juanluisvaladas:
I see what happened, for some reason the bounded frequency runner is getting the max sync period to one hour, instead of 30 seconds. I need to review this but I think I finally got the breakthrough

danwinship:
> I see what happened, for some reason the bounded frequency runner is getting the max sync period to one hour, instead of 30 seconds. I need to review this but I think I finally got the breakthrough

oh... it's probably the hybrid proxier. It replaces the iptables proxier's boundedfrequencyrunner with its own

We had to change the hybrid proxier to have a maxInternal of time.Hour too (or else it still forced iptables to resync every 30s, destroying the benefit of the patches). But maybe the fix there (at least in 4.2/4.3) is that it should run with a maxInternal of 30s, but it should call `maybeSyncProxyRules` rather than `forceSyncProxyRules`?

juanluisvaladas:
I'm pretty sure https://github.com/openshift/sdn/pull/112/commits/e35bf98516b27420738cb593210e343096ae11f3 fixed it, but nevertheless I'm gonna make yet another execution with more logging.

danwinship:
> But maybe the fix there (at least in 4.2/4.3) is that it should run with a maxInternal of 30s, but it should call `maybeSyncProxyRules` rather than `forceSyncProxyRules`?

You need to do this too (in `vendor/k8s.io/kubernetes/pkg/proxy/iptables/proxier_openshift.go`). Otherwise, the hybrid proxier will be forcing the iptables proxier to do a full `iptables-restore` every 30s, meaning the entire rest of the patchset is useless...

(You should confirm that the end result is both that (a) kube-proxy stays healthy, AND (b) kube-proxy doesn't run iptables-restore except when things actually change.)

juanluisvaladas:
Oh, I missed the previous update, will check that

danwinship:
closing in favor of #178

PR #113, opened 2020-03-04
[WIP]: I want some debugging tools

alexanderConstantinescu:
Debugging tools...

### DIFF
--- a/images/sdn/Dockerfile.fedora
+++ b/images/sdn/Dockerfile.fedora
@@ -12,7 +12,8 @@ COPY --from=builder /go/src/github.com/openshift/sdn/sdn-cni-plugin /opt/cni/bin
 COPY --from=builder /go/src/github.com/openshift/sdn/host-local /usr/bin/cni/osdn-host-local
 
 RUN INSTALL_PKGS=" \
-      openvswitch container-selinux socat ethtool nmap-ncat \
+      openvswitch tcpdump python3-netifaces openvswitch-test \
+      container-selinux socat ethtool nmap-ncat \
       libmnl libnetfilter_conntrack conntrack-tools \
       libnfnetlink iproute bridge-utils procps-ng \
       iputils binutils xz \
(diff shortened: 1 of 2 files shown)


danwinship:
The sdn image gets deployed to every node; we don't want to make it bigger with stuff that isn't needed for ordinary use. You can deploy your own debugging pods to nodes. (eg, just a privileged hostNetwork `image: fedora:31` pod and then `dnf install` whatever you need).

Maybe we should create a standard network debugging pod definition in CNO or something

alexanderConstantinescu:
> The sdn image gets deployed to every node; we don't want to make it bigger with stuff that isn't needed for ordinary use. You can deploy your own debugging pods to nodes. (eg, just a privileged hostNetwork `image: fedora:31` pod and then `dnf install` whatever you need).
> 
> Maybe we should create a standard network debugging pod definition in CNO or something

I never planned on merging this. I just needed to build an image with these tools inside as it's impossible to install packages in existing containers outside our build system due to:

```
$ yum install PACKAGE_OF_YOUR_CHOICE -y
...
a bunch of logs
...
failure: repodata/repomd.xml from rhel-fast-datapath-beta: [Errno 256] No more mirrors to try.
http://base-4-3.ocp.svc/rhel-fast-datapath-beta/repodata/repomd.xml: [Errno 14] curl#6 - "Could not resolve host: base-4-3.ocp.svc; Unknown error"
```

alexanderConstantinescu:
> Maybe we should create a standard network debugging pod definition in CNO or something

I like that idea however. Would we have any possibilities to deploy that pod in an easy way? I.e configuring or creating a CRD which could do something like (just thinking out loud....):

`oc network-debug node/X`

And would launch a daemonset on all nodes with those tools installed? Anything like that existing?

alexanderConstantinescu:
> we could just have a DaemonSet yaml file in `cluster-network-operator/hack/` or something...

Well, that's....much easier. I'll try to cook something like that up

PR #114, opened 2020-03-09
Bug 1811739: Fix kube_proxy metrics

juanluisvaladas:
On the rebase to kubernetes 1.17 the metrics were moved to
"k8s.io/component-base/metrics/legacyregistry" as part of:
https://github.com/kubernetes/enhancements/blob/master/keps/sig-instrumentation/20190404-kubernetes-control-plane-metrics-stability.md
But we were using the old endpoint which was
github.com/prometheus/client_golang/prometheus .

This effort is work in progress so I expect to have problems in future
rebases, most certainly we won't face issues here on the 1.18 and 1.19
rebases.

### DIFF
--- a/pkg/network/node/metrics.go
+++ b/pkg/network/node/metrics.go
@@ -12,11 +12,11 @@ import (
 	"sync"
 	"time"
 
-	"github.com/prometheus/client_golang/prometheus"
-
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 
 	"github.com/openshift/sdn/pkg/network/node/ovs"
+	"k8s.io/component-base/metrics"
+	"k8s.io/component-base/metrics/legacyregistry"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


danwinship:
> This effort is work in progress so I expect to have problems in future
> rebases

maybe add a note to HACKING.md?

/lgtm

juanluisvaladas:
> maybe add a note to HACKING.md?
Good point, actually I need to rewrite the whole rebase thing. I forgot to update it after the change to go modules, so I'll make sure to add this in HACKING.md.

juanluisvaladas:
@danwinship added the note to HACKING.md
https://github.com/openshift/sdn/pull/117

juanluisvaladas:
/hold 
The failures are legit, looks like by fixing the kube-proxy metrics I broke some other metrics.

juanluisvaladas:
Once this is merged I'll modify https://github.com/openshift/origin/blob/master/test/extended/prometheus/prometheus.go#L292-L305 to make sure we also include the kubeproxy_ metrics

PR #116, opened 2020-03-09
Bug 1812063: UPSTREAM 83911: Fix DeltaFIFO Replace method

abhat:
Backport https://github.com/openshift/sdn/pull/60

### DIFF
--- a/glide.diff
+++ b/glide.diff
@@ -3,6 +3,52 @@ diff --no-dereference -N -r current/vendor/github.com/vishvananda/netlink/link_l
 < 			gre.FlowBased = true
 ---
 > 			gre.FlowBased = int8(datum.Value[0]) != 0
+diff --no-dereference -N -r current/vendor/k8s.io/client-go/tools/cache/delta_fifo.go updated/vendor/k8s.io/client-go/tools/cache/delta_fifo.go
+350a351,357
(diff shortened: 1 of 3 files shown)


danwinship:
#95 is supposed to fix shared informers in release-4.2.

It seems that kubernetes/kubernetes#83911 was not backported to kube 1.16 and 1.14. Are you sure that this commit is actually necessary or is it just something that we had randomly tried when we were trying to fix the informer problems?

knobunc:
@abhat - Sorry, this seems to have been fixed in 4.2.19 per https://bugzilla.redhat.com/show_bug.cgi?id=1747532#c14 can you confirm?

abhat:
@danwinship you are right, this is the backport for the upstream k/k PR: https://github.com/kubernetes/kubernetes/pull/83911. I think this combined with the informer changes that @squeed made in https://github.com/openshift/sdn/pull/79/files should make sure we are covered. The fixes in https://github.com/openshift/sdn/pull/79/files are already backported via: https://github.com/openshift/sdn/pull/90.

abhat:
As per @danwinship this backport is not releveant since k/k didn't seem to have the backport into 1.14 either. The informer fix in PR#95 should be sufficient to address the problem.

PR #120, opened 2020-03-16
Add --node-name and --node-ip flags to openshift-sdn-node

danwinship:
openshift-sdn-node tries to figure out the node IP by resolving the node name as a hostname and/or picking a default IP off the primary network interface. While this made sense historically, there's no point in doing this now since it runs as a pod and so we can just get kubelet to tell us the "official" node IP and should do that instead.

Relatedly, while we are already getting the node name from kubelet, we were passing it in a non-obvious way via environment variable, but we should allow passing it explicitly instead.

(cf https://bugzilla.redhat.com/show_bug.cgi?id=1802557, where we seem to be autodetecting the wrong egress IP)

### DIFF
--- a/pkg/network/node/node.go
+++ b/pkg/network/node/node.go
@@ -63,8 +63,8 @@ type osdnPolicy interface {
 }
 
 type OsdnNodeConfig struct {
-	Hostname string
-	SelfIP   string
+	NodeName string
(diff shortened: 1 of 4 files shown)


dcbw:
/retest

AWS throttling

PR #122, opened 2020-03-18
Bug 1813846: handle default-deny rule properely

alexanderConstantinescu:
Why this is needed:

Today the SDN networkpolicy code does not create the OVS drop flows for any pod which is created _after and matches_ a "default-deny" networkpolicy. This stems from the fact that `parseNetworkPolicy` only attaches flows to the `npPolicy` should the defined networkpolicy define a `.Spec.Ingress`, which is not always the case, as in the example for the mentioned bug. We need an additional check on the parsed 
`selectedIPs` so that this corner case can be handled and the proper OVS flows created accordingly here: https://github.com/openshift/sdn/blob/master/pkg/network/node/networkpolicy.go#L299

This will need a back-port to 4.1 eventually

/assign @danwinship

### DIFF
--- a/pkg/network/node/networkpolicy.go
+++ b/pkg/network/node/networkpolicy.go
@@ -577,7 +577,7 @@ func (np *networkPolicyPlugin) updateNetworkPolicy(npns *npNamespace, policy *ne
 	oldNPP, existed := npns.policies[policy.UID]
 	npns.policies[policy.UID] = npp
 
-	changed := !existed || !reflect.DeepEqual(oldNPP.flows, npp.flows)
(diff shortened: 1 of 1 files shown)


danwinship:
> Today the SDN networkpolicy code does not create the OVS drop flows for any pod which is created _after and matches_ a "default-deny" networkpolicy.

Huh. Yeah, more precisely, a policy *with a non-empty `spec.podSelector`* and an empty `spec.Ingress`. The more common case of empty `spec.podSelector` (to make *all* pods in the namespace default-deny) works; it's only broken if you try to make only *some* of the pods be default-deny.

/lgtm

PR #123, opened 2020-03-24
 Use a cache of pods in network policy

JacobTanenbaum:
Add a podMatchCache to replace constant matching based on label selectors.
Increases the perfomance of network policies.

### DIFF
--- a/pkg/network/node/networkpolicy.go
+++ b/pkg/network/node/networkpolicy.go
@@ -42,6 +42,8 @@ type networkPolicyPlugin struct {
 	namespaces map[uint32]*npNamespace
 	// nsMatchCache caches matches for namespaceSelectors; see selectNamespaceInternal
 	nsMatchCache map[string]*npCacheEntry
+	// podMatchCache caches matches for podSelectors; see selectPodsInternal
+	podMatchCache map[string]map[string]*npPodCacheEntry
 
 	pods map[ktypes.UID]corev1.Pod
 }
@@ -76,6 +78,11 @@ type npCacheEntry struct {
 	matches  map[string]uint32
 }
 
+type npPodCacheEntry struct {
+	selector labels.Selector
+	matches  map[string]string
+}
+
 type refreshForType string
 
 const (
@@ -89,7 +96,8 @@ func NewNetworkPolicyPlugin() osdnPolicy {
 		namespacesByName: make(map[string]*npNamespace),
 		pods:             make(map[ktypes.UID]corev1.Pod),
 
-		nsMatchCache: make(map[string]*npCacheEntry),
+		nsMatchCache:  make(map[string]*npCacheEntry),
+		podMatchCache: make(map[string]map[string]*npPodCacheEntry),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/network/node/networkpolicy.go
```diff
@@ -399,12 +448,11 @@ func (np *networkPolicyPlugin) selectPodsFromNamespaces(nsLabelSel, podLabelSel
...
 
 	namespaces := np.selectNamespacesInternal(nsSel)
-	for _, pod := range np.pods {
-		vnid, exists := namespaces[pod.Namespace]
-		if exists && podSel.Matches(labels.Set(pod.Labels)) {
-			peerFlows = append(peerFlows, fmt.Sprintf("reg0=%d, ip, nw_src=%s, ", vnid, pod.Status.PodIP))
+	for namespace := range namespaces {
+		pods := np.selectPodsInternal(namespace, podSel)
+		for _, podIPs := range pods {
+			peerFlows = append(peerFlows, fmt.Sprintf("reg0=%d, ip, ne_src=%s, ", namespaces[namespace], podIPs))
```
juanluisvaladas: typo: ne_src instead of nw_src

JacobTanenbaum: fixed

JacobTanenbaum:
@juanluisvaladas I am not sure why it was doing that either, I fixed the other issue and now it is not happining anymore...

juanluisvaladas:
I misunderstood how the test error printed, the only issue was the typo.
The whole thing looks good to me, did you do any performance comparison to see how it performs compared to the old implementation? Other than this question it looks good to me.

danwinship:
There's a test of the namespace selector cache in `pkg/network/node/networkpolicy_test.go`. It's disabled (by putting a "_" in front of its name) because I couldn't find a reliable way to test the timing under CI. But anyway, you could add a similar test for pod selectors, and run it locally against the old and new code, to confirm that the cache is working, and then also leave it in `networkpolicy_test.go` but disabled, so other people can also use it in the future.

danwinship:
> you could add a similar test

Well, the namespaceSelector cache test is designed to basically emulate the reproducer from the original bug report about that. The podSelector cache test might end up having to be fairly different to reproduce the slow pod selecting bug, but it would be similar in idea anyway...

pkg/network/node/networkpolicy.go
```diff
@@ -42,6 +42,8 @@ type networkPolicyPlugin struct {
 	namespaces map[uint32]*npNamespace
 	// nsMatchCache caches matches for namespaceSelectors; see selectNamespaceInternal
 	nsMatchCache map[string]*npCacheEntry
+	// podMatchCache caches matches for podSelectors; see selectPodsInternal
+	podMatchCache map[string]map[string]*npPodCacheEntry
```
danwinship: Maybe rename `npCacheEntry` to `npNamespaceCacheEntry` or `npNSCacheEntry` now. (As a separate commit before this one)

pkg/network/node/networkpolicy.go
```diff
@@ -361,6 +369,22 @@ func (np *networkPolicyPlugin) selectNamespacesInternal(selector labels.Selector
 	return match.matches
 }
 
+func (np *networkPolicyPlugin) selectPodsInternal(namespaceName string, selector labels.Selector) map[string]string {
+	cacheKey := selector.String()
+	match := np.podMatchCache[namespaceName][cacheKey]
+	if match == nil {
+		np.podMatchCache[namespaceName] = make(map[string]*npPodCacheEntry)
```
danwinship: You need to have two checks here, one for if `ns.podMatchCache[namespaceName]` exists, and one for if `ns.podMatchCache[namespaceName][cacheKey]` exists. The way the code is now, if you call `selectPodsInternal("namespace1", selectorA)` and then `selectPodsInternal("namespace1", selectorB)`, then the second call will reinitialize `np.podMatchCache["namespace1"]` with an empty map again and delete the cached results for `selectorA`.

pkg/network/node/networkpolicy.go
```diff
@@ -371,6 +395,31 @@ func (np *networkPolicyPlugin) updateMatchCache(npns *npNamespace) {
...
 
+func (np *networkPolicyPlugin) updatePodMatchCache(pod, oldPod *corev1.Pod) {
+	var potentialMatches map[string]*npPodCacheEntry
+	if pod != nil {
+		potentialMatches = np.podMatchCache[pod.Namespace]
+	}
+
+	//pod deletion case
+	if pod == nil && oldPod != nil {
+		potentialMatches = np.podMatchCache[oldPod.Namespace]
```
danwinship: You could simplify this a little by just passing the namespace name in from the caller so that then you don't have to worry about which of `pod` or `oldPod` is `nil` here

pkg/network/node/networkpolicy.go
```diff
@@ -371,6 +395,31 @@ func (np *networkPolicyPlugin) updateMatchCache(npns *npNamespace) {
...
+	}
+
+	if oldPod != nil {
+		for _, cacheEntry := range potentialMatches {
+			delete(cacheEntry.matches, pod.Name)
```
danwinship: `oldPod.name`

pkg/network/node/networkpolicy.go
```diff
@@ -399,12 +448,11 @@ func (np *networkPolicyPlugin) selectPodsFromNamespaces(nsLabelSel, podLabelSel
...
 
 	namespaces := np.selectNamespacesInternal(nsSel)
-	for _, pod := range np.pods {
-		vnid, exists := namespaces[pod.Namespace]
-		if exists && podSel.Matches(labels.Set(pod.Labels)) {
-			peerFlows = append(peerFlows, fmt.Sprintf("reg0=%d, ip, nw_src=%s, ", vnid, pod.Status.PodIP))
+	for namespace := range namespaces {
+		pods := np.selectPodsInternal(namespace, podSel)
+		for _, podIPs := range pods {
+			peerFlows = append(peerFlows, fmt.Sprintf("reg0=%d, ip, nw_src=%s, ", namespaces[namespace], podIPs))
```
danwinship: oh, I was like "how does that work with multiple pod IPs?" but it's just the variable name that's confusing; it should be `podIP` not `podIPs`

pkg/network/node/networkpolicy.go
JacobTanenbaum: done

pkg/network/node/networkpolicy.go
JacobTanenbaum: done

pkg/network/node/networkpolicy.go
JacobTanenbaum: done

pkg/network/node/networkpolicy.go
JacobTanenbaum: I see what your saying and I believe that I corrected it

PR #127, opened 2020-04-04
Bug 1822351: Fix hybrid proxier for iptables.Monitor

danwinship:
The iptables proxier has been fixed to not constantly resync its rules when not needed... except that the hybrid proxier was forcing it to do it anyway. Fix that by moving the NodeIPTables monitor up out to the top level of openshift-sdn-node and sharing it between the node and proxy code.

/hold
for testing


squeed:
Does this break any of our alerts? otherwise lgtm

PR #132, opened 2020-04-16
Bug 1824203: Make egressVXLANMonitor updates channel buffered

juanluisvaladas:
The egressIPTracker has methods that lock eit.mutex and that call
evm functions that lock evm.mutex.

The problem with this is that evm.mutex has to write to the evm.updates
channel which isn't buffered and becomes blocked until
eit.setNodeOffline, which also locks eit.mutex, is running.

This causes a deadlock. By adding a buffered channel with 300 elements
we should avoid the locking problem because "640k ought to be enough
for anybody"

### DIFF
--- a/pkg/network/node/egressip.go
+++ b/pkg/network/node/egressip.go
@@ -52,7 +52,7 @@ func newEgressIPWatcher(oc *ovsController, localIP string, masqueradeBit *int32)
 func (eip *egressIPWatcher) Start(networkInformers networkinformers.SharedInformerFactory, iptables *NodeIPTables) error {
 	eip.iptables = iptables
 
-	updates := make(chan *egressVXLANNode)
+	updates := make(chan *egressVXLANNode, 300)
 	eip.vxlanMonitor = newEgressVXLANMonitor(eip.oc.ovs, eip.tracker, updates)
 	go eip.watchVXLAN(updates)
 


juanluisvaladas:
BTW I'm fully aware this is a horrible hack.

juanluisvaladas:
@danwinship perhaps you should have a look before actually merging this.

danwinship:
yeah, I did mean to look at this....

Any channel size other than 0 or 1 is a code smell. (They talk about "massive" egress IP migrations in the bug... Are you *sure* 300 is enough?) The real fix would be to reorganize the locking to put the channel write outside the lock.

Or alternatively, rather than pushing the actual updates over the channel, do [like BoundedFrequencyRunner does](https://github.com/kubernetes/kubernetes/blob/release-1.18/pkg/util/async/bounded_frequency_runner.go#L215), and have a length 1 `chan struct{}` where the channel is used purely to signal the other thread that it needs to act, but not to pass the actual data, and so if there's already an item in the channel, you don't need to signal it again, and when the reader thread finally reads from the channel, it would call some `evm.GetUpdatedNodes()` method to get the list of pending updates.

juanluisvaladas:
> Any channel size other than 0 or 1 is a code smell. 
Agree.

> (They talk about "massive" egress IP migrations in the bug... Are you sure 300 is enough?)
Their largest node has 92 egress IPs so this buys us their 3 largest full nodes plus some margin, and that'd be without executing eit.setNodeOffline a single time in the meantime.
I'm not entirely certain this will be enough

> The real fix would be to reorganize the locking to put the channel write outside the lock.
Agree, should I put this on hold and do that instead?

danwinship:
I guess I'd be OK approving this as long as you at least filed an issue about fixing things better and left a FIXME in the code pointing to that (so anyone looking at the current code when thinking about ovn-kubernetes egress IPs will understand the problem)

PR #134, opened 2020-04-27
Bug 1825355: node/vnids: Correctly handle case where NetNamespace watch is far behind

squeed:
When adding a pod, if the NetNamespace isn't found, we'll issue a GET directly to the apiserver and treat it as an ADD. Except we didn't actually handle it correctly, and caused NetworkPolicy to ignore this NetNS forever.

Fixes: rhbz 1825355

### DIFF
--- a/pkg/network/node/vnids.go
+++ b/pkg/network/node/vnids.go
@@ -119,7 +119,7 @@ func (vmap *nodeVNIDMap) WaitAndGetVNID(name string) (uint32, error) {
 			return 0, fmt.Errorf("failed to find netid for namespace: %s, %v", name, err)
 		}
 		klog.Warningf("Netid for namespace: %s exists but not found in vnid map", name)
-		vmap.setVNID(netns.Name, netns.NetID, netnsIsMulticastEnabled(netns))
+		vmap.handleAddOrUpdateNetNamespace(netns, nil, watch.Added)
 		return netns.NetID, nil
 	}
 }
--- a/pkg/network/node/networkpolicy.go
+++ b/pkg/network/node/networkpolicy.go
@@ -102,6 +102,9 @@ func (np *networkPolicyPlugin) SupportsVNIDs() bool {
 }
 
 func (np *networkPolicyPlugin) Start(node *OsdnNode) error {
+	np.lock.Lock()
+	defer np.lock.Unlock()
+
 	np.node = node
 	np.vnids = newNodeVNIDMap(np, node.networkClient)
 	if err := np.vnids.Start(node.networkInformers); err != nil {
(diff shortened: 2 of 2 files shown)


squeed:
@JacobTanenbaum @danwinship you've both touched this recently

pkg/network/node/vnids.go
```diff
@@ -119,7 +119,7 @@ func (vmap *nodeVNIDMap) WaitAndGetVNID(name string) (uint32, error) {
 			return 0, fmt.Errorf("failed to find netid for namespace: %s, %v", name, err)
 		}
 		klog.Warningf("Netid for namespace: %s exists but not found in vnid map", name)
-		vmap.setVNID(netns.Name, netns.NetID, netnsIsMulticastEnabled(netns))
+		vmap.handleAddOrUpdateNetNamespace(netns, nil, watch.Added)
```
danwinship, line 122: I don't think it's legitimate to call `handleAddOrUpdateNetNamespace` from here. In fact, it's *definitely* not, as seen by the fact that you had to change a bunch of other places to make it work. But we can't just change places that call `WaitAndGetVNID` to call `getVNID` instead and expect everything will keep working.

Maybe the fix is to just remove the `setVNID` call here. Though I think if we were going to do that I'd want to make the backoff shorter...

squeed: I had considered that. We cache NetNamespaces in two places (networkpolicy.go and vnids.go), and doing that would make those caches diverge. It's not clear what the implication of such a divergence is, given that the code is so tightly coupled.

I actually think it's an error for `networkPolicyPlugin.initNamespaces()` to call `WaitAndGetVNID()`, because we're still in startup and haven't even added our handlers yet.

squeed: To clarfiy, the flow in `npp.Start()` is:
1. `vmap.Start()`, which calls
2. `vmap.populateVNIDs()`, which does a synchronous List and calls `vmap.setVNID()`
3. `npp.initNamespaces()`, which does a synchronous List
4. The rest of the informers are configured

Not the prettiest. So I'm not surprised we have deadlocks. But that's why I think it's wrong to call an informer handler before we "expect" to see informers running.

danwinship: ok, but you need to separate out the startup time vs non-startup time behavior more. Currently, at startup the behavior is:

- if a VNID is missing from the cache, pointlessly wait 5 seconds, then fetch it manually

With this patch, it becomes

- if a VNID is missing from the cache, abort openshift-sdn startup

squeed: > * if a VNID is missing from the cache, abort openshift-sdn startup

We swallow errors (and always have), so that's not a risk.

squeed:
@danwinship I switched the locking around a bit, to make the difference between startup and running clearer.

squeed:
@danwinship any final thoughts on this?

danwinship:
/lgtm
but the change to `WaitAndGetVNID` is tricky and this patch should not simply be backported as fast as possible.

PR #135, opened 2020-05-01
Bug 1781575: Discard destination unreachable from rate mask

juanluisvaladas:
We don't know why yet but in OCP 4.x we're hitting the ICMP rate limit and this is causing the "Services should be rejected when no endpoints exist" test to fail often.

Until we figure out why add this tunable to avoid hitting such limit

https://www.kernel.org/doc/Documentation/networking/ip-sysctl.txt

From kernel docs:
```
icmp_ratemask - INTEGER
	Mask made of ICMP types for which rates are being limited.
	Significant bits: IHGFEDCBA9876543210
	Default mask:     0000001100000011000 (6168)

	Bit definitions (see include/linux/icmp.h):
		0 Echo Reply
		3 Destination Unreachable *
```
6168 - 2³ = 6160

### DIFF
--- a/pkg/network/node/sdn_controller.go
+++ b/pkg/network/node/sdn_controller.go
@@ -111,6 +111,13 @@ func (plugin *OsdnNode) SetupSDN() (bool, map[string]podNetworkInfo, error) {
 		return false, nil, fmt.Errorf("net/ipv4/ip_forward=0, it must be set to 1")
 	}
 
+	// In OCP 4 apparently we have a lot of rejects.
+	// Default rate mask is 6188, 6160 is default without destination unreachable
+	// FIXME: Why is this new to OCP 4?
+	if err = sysctl.SetSysctl("net/ipv4/icmp_ratemask", 6160); err != nil {
+		return false, nil, fmt.Errorf("Failed to setup icmp_ratemask: %s", err)
+	}
+
 	localSubnetCIDR := plugin.localSubnetCIDR
 	_, ipnet, err := net.ParseCIDR(localSubnetCIDR)
 	if err != nil {


pkg/network/node/sdn_controller.go
```diff
@@ -111,6 +111,13 @@ func (plugin *OsdnNode) SetupSDN() (bool, map[string]podNetworkInfo, error) {
 		return false, nil, fmt.Errorf("net/ipv4/ip_forward=0, it must be set to 1")
 	}
 
+	// In OCP 4 apparently we have a lot of rejects.
+	// Default rate mask is 6188, 6160 is default without destination unreachable
```
danwinship, line 115: typo. default is 6168. but it would be better to read the current value and unset the relevant bit?

pkg/network/node/sdn_controller.go
```diff
@@ -111,6 +111,13 @@ func (plugin *OsdnNode) SetupSDN() (bool, map[string]podNetworkInfo, error) {
...
+	// FIXME: Why is this new to OCP 4?
```
danwinship, line 116: It's not. We skipped the "should be rejected" test in 3.x, then it seemed to be working in 4.x so we re-enabled it, except then it turned out to still be flaky for reasons unknown.

danwinship:
Not clear that this fix belongs in openshift-sdn...

pkg/network/node/sdn_controller.go
juanluisvaladas: Same symptoms but not exactly the same cause, the 3.x bug was happening because we didn't have an iptables rules that would reject *forwarded* packets, however, if you trid to reach a service without endpoints from the host network it was rejected.
More https://bugzilla.redhat.com/show_bug.cgi?id=1782857

pkg/network/node/sdn_controller.go
juanluisvaladas: Sure, can be done.

juanluisvaladas:
@danwinship 
> Not clear that this fix belongs in openshift-sdn...
Any suggestion?
I think I can do a PR in CNO to create a tuned resource (as in oc get tuned, I actually need to test if this is feasible) or perhaps would it be on MCO a better place?.
 
CNO isn't running anywhere so it can't read the previous value and setting the 4th bit to 0 on each node, MCO can probably achieve this with some hacking but it doesn't seem to have any existing code for distinguishing ovn-kubernetes, openshift-sdn or a different sdn .

danwinship:
> but it doesn't seem to have any existing code for distinguishing ovn-kubernetes, openshift-sdn or a different sdn .

No, that's my point. This fix isn't specific to openshift-sdn. If "connections to services with no endpoints should be rejected" is a core k8s feature, then we need to un-rate-limit ICMP messages no matter what the plugin is.

cgwalters:
> If "connections to services with no endpoints should be rejected" is a core k8s feature, then we need to un-rate-limit ICMP messages no matter what the plugin is.

Probably.  I guess we could do that in the MCO by default?  Though I wonder if we more want to have different behavior when packets come over the SDN versus e.g. external interfaces.  Not sure how hard that'd be.

This ICMP ratelimiting is...old.  It dates to the [original Linux kernel git commit](https://github.com/torvalds/linux/blob/1da177e4c3f41524e886b7f1b8a0c1fc7321cac2/Documentation/networking/ip-sysctl.txt#L409).  More recently updated [here](https://github.com/torvalds/linux/commit/6dbf4bcac98bbc76ef425b3a2b4169f31199f6c7) by someone who was presumably trying to change it for some reason.

cgwalters:
Where's the successor to this PR?

juanluisvaladas:
I should have one PR to MCO later today

PR #117, opened 2020-03-10
Bug 1812052: Update HACKING.md rebase to use go modules

juanluisvaladas:
I forgot to update HACKING.md after the migration to go modules.
This includes it.

PR #136, opened 2020-05-05
Bug 1832153: Upstream v1.18.2 rebase

squeed:
Nothing too exciting, but a lot of mechanical changes:

- All client-go operations need a `context`
- Some former `string` types now have explicit names, so lots of boilerplate
- Some small kube-proxy cmd changes


squeed:
This currently points to a personal fork of kubernetes - once CI looks reasonable will push to origin.

juanluisvaladas:
Oh, I hope you were aware of this this: https://github.com/openshift/sdn/pull/117 or at least that you didn't miss it

squeed:
> Oh, I hope you were aware of this this: #117 or at least that you didn't miss it

Yup, I followed it - it worked quite well!

pkg/network/common/common.go
```diff
@@ -224,3 +225,19 @@ func GetHostIPNetworks(skipInterfaces []string) ([]*net.IPNet, []net.IP, error)
 	}
 	return hostIPNets, hostIPs, kerrors.NewAggregate(errList)
 }
+
+func HSEgressIPsToStrings(ips []networkv1.HostSubnetEgressIP) []string {
```
danwinship, line 229: Sigh. This is a mess and we should revert the `openshift/api` change that required it.
(At a minimum, split the egressip changes into a separate commit so they're easily revertable when we fix this later.)

juanluisvaladas: In case someone else gets here later and also wonders the rationale for this:
https://github.com/openshift/api/commit/a63a88e3fd742e3522b20f47b40e38336570dc25

juanluisvaladas:
Looks good to me so far. Also checked your local 1.18 branch and only has the two old carry patches so it looks fine there too.

I'm launching the e2e-gcp for you because I'm pretty sure the failing test is unrelated.
/test e2e-gcp

pkg/network/common/common.go
squeed: grump: go has exactly two "magic" types. Why it can't implement type conversion across literally the only two polymorphic types in the whole world is infuriating.

squeed: @abhat can fill us in, but I believe the issue is that it's not possible to add validation to an array

danwinship: Yes, but we don't *need* validation there if it's just going to make our lives more difficult. We have to validate the network CRD types at runtime anyway because the CRD validation only handles part of it. There is talk about adding a ValidatingWebHook at some point so we can do this properly.

squeed:
okay, just waiting for someone with more power to push to `openshift/kubernetes`

squeed:
ok, @danwinship I split the commits.

Since CI passed, I've pushed the tag to openshift/kubernetes, and updated go.mod accordingly. This is ready to go.

juanluisvaladas:
Oof, looks like the verify-deps strikes again... @squeed the make update-deps-overrides produces different vendor depending on your go version, env, and god knows what.

I strongly recommend that you get pull verify-deps pod image, run it locally, run the make update-deps-overrides inside that container and copy the contents of it into your vendor directory with docker cp or whatever container runtime engine you're using.

squeed:
> Oof, looks like the verify-deps strikes again... 

Derp, forgot to `go mod tidy`.

squeed:
Dan observed that kubernetes branch should have been a fork off of upstream, not origin.

Fixed that, revendored, repushed.

danwinship:
lgtm but should we wait for https://github.com/openshift/kubernetes/pull/126 first to avoid a second bump?

squeed:
> lgtm but should we wait for [openshift/kubernetes#126](https://github.com/openshift/kubernetes/pull/126) first to avoid a second bump?

Hah, I figured you'd be mad if I conflated that pr.

PR #139, opened 2020-05-19
Bug 1824203: Fix egressVXLANMonitor and egressIPTracker deadlock

juanluisvaladas:
The egressIPTracker has methods that lock eit.mutex and that call
evm functions that lock evm.mutex.

The problem with this is that evm.mutex has to write to the evm.updates
channel which isn't buffered and becomes blocked until
eit.setNodeOffline, which also locks eit.mutex, is running.

This causes a deadlock. Initially I tried doing a horrible hack by
making the updates channel huge, however it wasn't enough.

Instead this fix adds a shared nodes list between eit and evm and the
evm uses the updates channel just to notify the eit that there are
updates in the shared list.

### DIFF
--- a/pkg/network/node/egressip.go
+++ b/pkg/network/node/egressip.go
@@ -52,10 +52,9 @@ func newEgressIPWatcher(oc *ovsController, localIP string, masqueradeBit *int32)
 func (eip *egressIPWatcher) Start(networkInformers networkinformers.SharedInformerFactory, iptables *NodeIPTables) error {
 	eip.iptables = iptables
 
-	updates := make(chan *egressVXLANNode, 300)
+	updates := make(chan struct{}, 1)
 	eip.vxlanMonitor = newEgressVXLANMonitor(eip.oc.ovs, eip.tracker, updates)
 	go eip.watchVXLAN(updates)
-
 	eip.tracker.Start(networkInformers.Network().V1().HostSubnets(), networkInformers.Network().V1().NetNamespaces())
 	return nil
 }
@@ -248,8 +247,10 @@ func (eip *egressIPWatcher) releaseEgressIP(egressIP, mark string) error {
 	return nil
 }
 
-func (eip *egressIPWatcher) watchVXLAN(updates chan *egressVXLANNode) {
-	for node := range updates {
-		eip.tracker.SetNodeOffline(node.nodeIP, node.offline)
+func (eip *egressIPWatcher) watchVXLAN(updates chan struct{}) {
+	for range updates {
+		for _, node := range eip.vxlanMonitor.GetUpdates() {
+			eip.tracker.SetNodeOffline(node.nodeIP, node.offline)
+		}
 	}
 }
--- a/pkg/network/node/vxlan_monitor.go
+++ b/pkg/network/node/vxlan_monitor.go
@@ -31,9 +31,12 @@ import (
 // polled just a few milliseconds after some packets went out, in which case we obviously
 // need to give the remote node more time to respond before declaring it offline.)
 //
-// When the monitor decides a node has gone offline, it alerts its owner via the updates
-// channel, and then starts periodically pinging the node's SDN IP address, until the
-// incoming packet count increases, at which point it marks the node online again.
+// When the monitor decides a node has gone offline, the monitor adds the node to the
+// updateNodes map and it alerts its owner via the updates channel. Then it starts
+// periodically pinging the node's SDN IP address, until the incoming packet count
+// increases, at which point it marks the node online again by using the same mechanism.
+// The updatesNodes map is cleaned every time evm.GetUpdates() is called so that we aren't
+// needlessly calling egressIPTracker.SetNodeOffline().
 //
 // The fact that we (normally) use pod-to-egress traffic to do the monitoring rather than
 // actively pinging the nodes means that if an egress node falls over while no one is
@@ -45,7 +48,8 @@ type egressVXLANMonitor struct {
 
 	ovsif        ovs.Interface
 	tracker      *common.EgressIPTracker
-	updates      chan<- *egressVXLANNode
+	updates      chan<- struct{}
+	updateNodes  map[string]*egressVXLANNode
 	pollInterval time.Duration
 
 	monitorNodes map[string]*egressVXLANNode
@@ -69,11 +73,12 @@ const (
 	maxRetries          = 2
 )
 
-func newEgressVXLANMonitor(ovsif ovs.Interface, tracker *common.EgressIPTracker, updates chan<- *egressVXLANNode) *egressVXLANMonitor {
+func newEgressVXLANMonitor(ovsif ovs.Interface, tracker *common.EgressIPTracker, updates chan<- struct{}) *egressVXLANMonitor {
 	return &egressVXLANMonitor{
 		ovsif:        ovsif,
 		tracker:      tracker,
 		updates:      updates,
+		updateNodes:  make(map[string]*egressVXLANNode),
 		pollInterval: defaultPollInterval,
 		monitorNodes: make(map[string]*egressVXLANNode),
 	}
@@ -212,7 +217,7 @@ func (evm *egressVXLANMonitor) check(retryOnly bool) bool {
 			if in > node.in {
 				klog.Infof("Node %s is back online", node.nodeIP)
 				node.offline = false
-				evm.updates <- node
+				evm.updateNode(node)
 			} else if evm.tracker != nil {
 				// We can ignore the return value because if the node responds
 				// (with either success or "connection refused") we'll see it
@@ -238,7 +243,7 @@ func (evm *egressVXLANMonitor) check(retryOnly bool) bool {
 					klog.Warningf("Node %s is offline", node.nodeIP)
 					node.retries = 0
 					node.offline = true
-					evm.updates <- node
+					evm.updateNode(node)
 				} else {
 					klog.V(2).Infof("Node %s may be offline... retrying", node.nodeIP)
 					retry = true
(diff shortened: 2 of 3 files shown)


juanluisvaladas:
The fix should be done, but I need to actually test it manually. Just made the PR in order to pass the automatic tests first.

juanluisvaladas:
@palonsoro I know it's not up to you to lgtm or not, but since you filed the original bug, would you mind having a look too?

juanluisvaladas:
/retitle [WIP] Bug 1824203: Fix egressVXLANMonitor and egressIPTracker deadlock

pkg/network/node/vxlan_monitor.go
```diff
@@ -31,9 +31,10 @@ import (
 // polled just a few milliseconds after some packets went out, in which case we obviously
 // need to give the remote node more time to respond before declaring it offline.)
 //
-// When the monitor decides a node has gone offline, it alerts its owner via the updates
-// channel, and then starts periodically pinging the node's SDN IP address, until the
-// incoming packet count increases, at which point it marks the node online again.
+// When the monitor decides a node has gone offline, adds it to the updateNodes map and
```
pecameron, line 34: offline, the monitor adds the node to
wording

juanluisvaladas:
Phil, ack, will fix that on the next push when I fix the unit test.
/test verify 
The unit test failure is a legit one.

palonsoro:
The only things I see:
- Lock in `watchVXLAN()` is wrong. If it gets the lock and then waits for stuff to come from `updates` channel, `updateNodes` won't be able to acquire the lock to add nodes to `updateNodes` and send notification through the channel.
- I would suggest renaming the `nodeIP` parameter on `updateNode()` method to something like `node` or `evn` for better readability.

The rest looks very good to me. Thanks!

juanluisvaladas:
Verify is probably still broken and I need to add a couple more tests cases but the original test suite is OK now

juanluisvaladas:
/hold
Manual testing has shown some problems, need to properly debug this.

juanluisvaladas:
/hold cancel
/test verify

Actually it wasn't triggering the change because the node was failing to detect the node off-line because there wasn't outgoing traffic on the netnamespace.

Also I added a new test that actually recreates the deadlock conditions.

juanluisvaladas:
@danwinship Can you PTAL?

pkg/network/node/egressip.go
```diff
@@ -52,9 +53,11 @@ func newEgressIPWatcher(oc *ovsController, localIP string, masqueradeBit *int32)
 func (eip *egressIPWatcher) Start(networkInformers networkinformers.SharedInformerFactory, iptables *NodeIPTables) error {
 	eip.iptables = iptables
 
-	updates := make(chan *egressVXLANNode, 300)
-	eip.vxlanMonitor = newEgressVXLANMonitor(eip.oc.ovs, eip.tracker, updates)
-	go eip.watchVXLAN(updates)
+	updates := make(chan struct{}, 1)
+	updateNodes := make(map[string]*egressVXLANNode)
+	var updateNodesMux sync.Mutex
+	eip.vxlanMonitor = newEgressVXLANMonitor(eip.oc.ovs, eip.tracker, updates, &updateNodes, &updateNodesMux)
```
danwinship, line 59: This is kind of a wacky API. You should just have evm maintain the map by itself, and when eip needs to know, it can call `evm.GetUpdates()` which will lock the existing evm mutex, grab the current map and replace it with an empty one, and return the old map to eip.

pkg/network/node/vxlan_monitor.go
```diff
@@ -31,9 +31,10 @@ import (
 // polled just a few milliseconds after some packets went out, in which case we obviously
 // need to give the remote node more time to respond before declaring it offline.)
 //
-// When the monitor decides a node has gone offline, it alerts its owner via the updates
-// channel, and then starts periodically pinging the node's SDN IP address, until the
-// incoming packet count increases, at which point it marks the node online again.
+// When the monitor decides a node has gone offline, the monitor adds the node to the
+// updateNodes map and it alerts its owner via the updates channel. Then starts periodically
```
danwinship, line 35: "Then IT starts..."

pkg/network/node/vxlan_monitor.go
```diff
@@ -256,6 +261,18 @@ func (evm *egressVXLANMonitor) check(retryOnly bool) bool {
...
 
+func (evm *egressVXLANMonitor) updateNode(node *egressVXLANNode) {
+	evm.updateNodesMux.Lock()
+	defer evm.updateNodesMux.Unlock()
+	evm.updateNodes[node.nodeIP] = node
+
+	// Don't update the updates channel if it already contains an entry.
+	// Otherwise we'll block this goroutine.
+	if len(evm.updates) == 0 {
+		evm.updates <- struct{}{}
```
danwinship, line 272: The better idiom is

    select {
            case evm.updates <- struct{}{}:
            default:
    }

which is race-proof. (Which, admittedly, doesn't actually matter here because you have a mutex too, but...)

juanluisvaladas:
@danwinship PTAL addressed everything, looks much better now thanks!

juanluisvaladas:
Actually I need to update the comment which is incorrect now, but I don't plan any code change

juanluisvaladas:
OK comment fixed

juanluisvaladas:
I just fixed a couple typos in the comment and removed a couple blank lines, nothing relevant.

pkg/network/node/vxlan_monitor.go
```diff
@@ -31,9 +31,10 @@ import (
 // polled just a few milliseconds after some packets went out, in which case we obviously
 // need to give the remote node more time to respond before declaring it offline.)
 //
-// When the monitor decides a node has gone offline, it alerts its owner via the updates
-// channel, and then starts periodically pinging the node's SDN IP address, until the
-// incoming packet count increases, at which point it marks the node online again.
+// When the monitor decides a node has gone offline, the monitor adds the node to the
+// updateNodes map and it alerts its owner via the updates channel. Then it starts
+// periodically pinging the node's SDN IP address, until the incoming packet count
+// increases, at which// point it marks the node online again by using the same mechanism.
```
danwinship, line 37: stray `//` inside the comment

pkg/network/node/vxlan_monitor.go
```diff
@@ -267,3 +281,15 @@ func (evm *egressVXLANMonitor) poll() (bool, error) {
...
+
+func (evm *egressVXLANMonitor) GetUpdates() []egressVXLANNode {
+	evm.Lock()
+	defer evm.Unlock()
+	values := []egressVXLANNode{}
+	for _, value := range evm.updateNodes {
+		values = append(values, *value)
+	}
+	// cleanup so that when GetUpdates is called again we don't
+	evm.updateNodes = make(map[string]*egressVXLANNode)
```
danwinship, line 293: comment is missing the exciting conclusion?

pkg/network/node/vxlan_monitor_test.go
```diff
@@ -67,7 +67,7 @@ func TestEgressVXLANMonitor(t *testing.T) {
 
 	// Everything should be fine at startup
 	retry := evm.check(false)
-	if update := peekUpdate(updates); update != nil {
+	if update := peekUpdate(updates, evm); update != nil {
 		t.Fatalf("Initial check showed updated node %#v", update)
```
danwinship, line 71: while you're updating this anyway, you changed "node" to "nodes" in about half of the error messages here, but not the other half...

pkg/network/node/vxlan_monitor_test.go
```diff
@@ -211,15 +218,56 @@ func TestEgressVXLANMonitor(t *testing.T) {
...
+
+	if update := peekUpdate(updates, evm); update != nil {
 		t.Fatalf("Check erroneously showed additional updated node %#v", update)
 	}
 	if retry {
 		t.Fatalf("Check erroneously requested retry")
 	}
+
+	// evm.check doesn't block when evm.updates is full.
+	// https://bugzilla.redhat.com/show_bug.cgi?id=1824203
```
danwinship, line 237: "Ensure evm.check doesn't block ...". The way it is now it looks like you're just quoting the bug description, which would imply the desired behavior was the opposite of that

pkg/network/node/vxlan_monitor_test.go
```diff
@@ -211,15 +218,56 @@ func TestEgressVXLANMonitor(t *testing.T) {
...
+
+	evm.updates <- struct{}{} // make updates buffer artificially full
+
+	// Send some traffic out but don't receive any.
+	// Do it sequentially in order to ensure it's OK for it two happen
```
danwinship, line 242: for it TO happen

pkg/network/node/vxlan_monitor_test.go
```diff
@@ -211,15 +218,56 @@ func TestEgressVXLANMonitor(t *testing.T) {
...
+			t.Fatal("Unexpected flush of the updates channel")
+		}
+		for i := 0; i < 3; i++ {
+			packetsOut(ovsif, outCounts, node, 0x41)
+			retry = evm.check(retry)
+		}
+	}
+
+	update := peekUpdate(updates, evm)
+	// because we're doing the ndoe updates sequentially we can expect
```
danwinship, line 257: "node"

pkg/network/node/vxlan_monitor_test.go
```diff
@@ -211,15 +218,56 @@ func TestEgressVXLANMonitor(t *testing.T) {
...
+	// the the nodes to be marked offline in an specific order.
```
danwinship, line 258: "the the"
"A specific"

juanluisvaladas:
/retest
unrelated failure

PR #142, opened 2020-05-23
build-image-sdn-test: Make binaries debuggable

juanluisvaladas:
Previous make build-image-sdn-test had a few issues:
1- Build-machinery-go strips binaries by defualt and doesn't produce any
debuginfo.
2- Because we were building the image ocp/builder:golang-1.13, and
running it on fedora, gcore had problems producing core dumps because of
incompatible libthread and libpthread.

Now we build the images directly in fedora, the image is significantly
larger but we now can obtain core dumps and even install delve in the
pod and attach to the live process and we even have the sources.

### DIFF
--- a/images/sdn/Dockerfile.fedora
+++ b/images/sdn/Dockerfile.fedora
@@ -1,32 +1,38 @@
 # This can be used to build images for testing
-
-FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.13 AS builder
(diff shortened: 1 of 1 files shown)


juanluisvaladas:
@pecameron @squeed PTAL

squeed:
definitely a hack, but I like it.
/lgtm
/approve

PR #143, opened 2020-05-26
Bug 1839107: node/vnids: Correctly handle case where NetNamespace watch is far behind

squeed:
This is a backport of #134

When adding a pod, if the NetNamespace isn't found, we'll issue a GET
directly to the apiserver and treat it as an ADD. Except we didn't
actually handle it correctly, and caused NetworkPolicy to ignore this
NetNS forever.

Fixes: rhbz 1839107


ashcrow:
/retitle Bug 1839107: node/vnids: Correctly handle case where NetNamespace watch is far behind

PR #149, opened 2020-06-15
Enable conntrack for ovs-multitenant

juanluisvaladas:
ovs-multitenant needs useConntrack to cover the scenario where trying to
reach an externalIP service from a pod in a project which uses egressIP.


juanluisvaladas:
/close
accidentally pushed this to the 4.5 branch instead of the 4.6

PR #138, opened 2020-05-06
Bug 1826339: vendor: bump our k8s vendor

squeed:
This is to pick up the backport of upstream 90175.

It also seems to fix some weirdness with go.mod that I don't really understand, but go mod tidy is happy.

PR #150, opened 2020-06-15
Bug 1762580: Enable conntrack for ovs-multitenant

juanluisvaladas:
ovs-multitenant needs useConntrack to cover the scenario where trying to
reach an externalIP service from a pod in a project which uses egressIP.

### DIFF
--- a/pkg/network/node/iptables.go
+++ b/pkg/network/node/iptables.go
@@ -18,7 +18,6 @@ import (
 type NodeIPTables struct {
 	ipt                iptables.Interface
 	clusterNetworkCIDR []string
-	masqueradeServices bool
 	vxlanPort          uint32
 	masqueradeBitHex   string // the masquerade bit as hex value
 
@@ -27,11 +26,10 @@ type NodeIPTables struct {
 	egressIPs map[string]string
 }
 
-func newNodeIPTables(ipt iptables.Interface, clusterNetworkCIDR []string, masqueradeServices bool, vxlanPort uint32, masqueradeBit uint32) *NodeIPTables {
+func newNodeIPTables(ipt iptables.Interface, clusterNetworkCIDR []string, vxlanPort uint32, masqueradeBit uint32) *NodeIPTables {
 	return &NodeIPTables{
 		ipt:                ipt,
 		clusterNetworkCIDR: clusterNetworkCIDR,
-		masqueradeServices: masqueradeServices,
 		vxlanPort:          vxlanPort,
 		masqueradeBitHex:   fmt.Sprintf("%#x", 1<<masqueradeBit),
 		egressIPs:          make(map[string]string),
@@ -161,17 +159,15 @@ func (n *NodeIPTables) getNodeIPTablesChains() []Chain {
 	var masq2Rules [][]string
 	var filterRules [][]string
 	for _, cidr := range n.clusterNetworkCIDR {
-		if n.masqueradeServices {
-			masqRules = append(masqRules, []string{"-s", cidr, "-m", "comment", "--comment", "masquerade pod-to-service and pod-to-external traffic", "-j", "MASQUERADE"})
-		} else {
-			masqRules = append(masqRules, []string{"-s", cidr, "-m", "comment", "--comment", "masquerade pod-to-external traffic", "-j", "OPENSHIFT-MASQUERADE-2"})
-			masq2Rules = append(masq2Rules, []string{"-d", cidr, "-m", "comment", "--comment", "masquerade pod-to-external traffic", "-j", "RETURN"})
-		}
+		masqRules = append(masqRules, []string{"-s", cidr, "-m", "comment", "--comment", "masquerade pod-to-external traffic", "-j", "OPENSHIFT-MASQUERADE-2"})
+		masq2Rules = append(masq2Rules, []string{"-d", cidr, "-m", "comment", "--comment", "masquerade pod-to-external traffic", "-j", "RETURN"})
 
 		filterRules = append(filterRules, []string{"-s", cidr, "-m", "comment", "--comment", "attempted resend after connection close", "-m", "conntrack", "--ctstate", "INVALID", "-j", "DROP"})
 		filterRules = append(filterRules, []string{"-d", cidr, "-m", "comment", "--comment", "forward traffic from SDN", "-j", "ACCEPT"})
 		filterRules = append(filterRules, []string{"-s", cidr, "-m", "comment", "--comment", "forward traffic to SDN", "-j", "ACCEPT"})
 	}
+
+	masq2Rules = append(masq2Rules, []string{"-j", "MASQUERADE"})
 
 	chainArray = append(chainArray,
 		Chain{
@@ -182,6 +178,11 @@ func (n *NodeIPTables) getNodeIPTablesChains() []Chain {
 			rules:    masqRules,
 		},
 		Chain{
+			table: "nat",
+			name:  "OPENSHIFT-MASQUERADE-2",
+			rules: masq2Rules,
+		},
+		Chain{
 			table:    "filter",
 			name:     "OPENSHIFT-FIREWALL-FORWARD",
 			srcChain: "FORWARD",
@@ -189,16 +190,6 @@ func (n *NodeIPTables) getNodeIPTablesChains() []Chain {
 			rules:    filterRules,
 		},
 	)
-	if !n.masqueradeServices {
-		masq2Rules = append(masq2Rules, []string{"-j", "MASQUERADE"})
-		chainArray = append(chainArray,
-			Chain{
-				table: "nat",
-				name:  "OPENSHIFT-MASQUERADE-2",
-				rules: masq2Rules,
-			},
-		)
-	}
 
 	// HACK: block access to MCS until we can secure it properly. Note that we share
 	// the same chain between OUTPUT and FORWARD.
--- a/pkg/network/node/node.go
+++ b/pkg/network/node/node.go
@@ -124,7 +124,6 @@ func New(c *OsdnNodeConfig) (*OsdnNode, error) {
 
 	var policy osdnPolicy
 	var pluginId int
-	var minOvsVersion string
 	switch strings.ToLower(networkInfo.PluginName) {
 	case networkutils.SingleTenantPluginName:
 		policy = NewSingleTenantPlugin()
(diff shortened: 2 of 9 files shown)


danwinship:
So I see that this is for a 3.11 bug. Are you planning to backport this to 3.11?

This is a slightly bigger change than it looks like. Among other things, Service handling is different between ovs-multitenant and ovs-networkpolicy, and I'm not sure if the conntrack-based version deals with multitenant semantics?

/test e2e-aws-multitenant

Also, if multitenant is going to use conntrack mode then ovs-subnet should too, and then there's no reason to keep non-conntrack mode. https://github.com/danwinship/sdn/commits/drop-multitenant has commits that do that, although it's in the context of dropping multitenant/subnet mode entirely and would need to be updated. (That branch was part of a half-baked idea that we could simplify the sdn code by moving multitenant mode to a separate fork.)

You'll need to bump `ruleVersion` in `ovscontroller.go` to make upgrades work.

juanluisvaladas:
> So I see that this is for a 3.11 bug. Are you planning to backport this to 3.11?
Yes, I already tested it on 3.11 too.

> This is a slightly bigger change than it looks like. Among other things, Service handling is different between ovs-multitenant and ovs-networkpolicy, and I'm not sure if the conntrack-based version deals with multitenant semantics?

So, your concern is that with this change a pod from service A can reach a pod from service B? If so I have *not* verified that I haven't broken that manually, I assume there is a test for this somewhere but anyway I will verify it manually if that's the case.

> Also, if multitenant is going to use conntrack mode then ovs-subnet should too, and then there's no reason to keep non-conntrack mode.

Makes sense to me, now that you mention it, does OCP 4 support ovs-subnet at all? Maybe we should just drop it entirely

> You'll need to bump ruleVersion in ovscontroller.go to make upgrades work.
Will do, thanks!

danwinship:
> > This is a slightly bigger change than it looks like. Among other things, Service handling is different between ovs-multitenant and ovs-networkpolicy, and I'm not sure if the conntrack-based version deals with multitenant semantics?
> 
> So, your concern is

More broadly my concern is "there could be non-obvious breakage" and we don't want to rush it into 3.11.

As for the Service issue specifically, e2e-aws-multitenant tests both that services that should work work and that services that shouldn't work don't work, so that looks good.

> does OCP 4 support ovs-subnet at all? Maybe we should just drop it entirely

We probably should have dropped it in 4.1 but we can't really now. It doesn't really add any complexity anyway; it's just NetworkPolicy mode without obeying NetworkPolicy.

danwinship:
(Also, FTR, assuming we do this, we can un-skip `Services should preserve source pod IP for traffic thru service cluster IP` in e2e-aws-multitenant.)

juanluisvaladas:
/hold

@danwinship should I also remove teh minOvsVersion? at the end of the day in ocp 4 we're guaranteed to ship a higher ovs than that. OCP 3.11 also upgrades the OVS daemonset, and it has been using 2.9 for a long time.

juanluisvaladas:
don't review this yet.

danwinship:
yeah, the minversion check has been pointless for a really long time

juanluisvaladas:
@danwinship can you PTAL?

danwinship:
needs rebase? For some reason part of #138 is showing up as a commit here...

pkg/network/node/iptables.go
```diff
@@ -189,16 +183,14 @@ func (n *NodeIPTables) getNodeIPTablesChains() []Chain {
...
-		masq2Rules = append(masq2Rules, []string{"-j", "MASQUERADE"})
-		chainArray = append(chainArray,
-			Chain{
-				table: "nat",
-				name:  "OPENSHIFT-MASQUERADE-2",
-				rules: masq2Rules,
-			},
-		)
-	}
+	masq2Rules = append(masq2Rules, []string{"-j", "MASQUERADE"})
```
danwinship, line 186: move this above to just after the loop...

pkg/network/node/iptables.go
```diff
@@ -189,16 +183,14 @@ func (n *NodeIPTables) getNodeIPTablesChains() []Chain {
...
+	chainArray = append(chainArray,
+		Chain{
+			table: "nat",
+			name:  "OPENSHIFT-MASQUERADE-2",
+			rules: masq2Rules,
+		},
+	)
```
danwinship, line 193: ...and then this can be part of the previous `chainArray = append(chainArray,` (in the order `OPENSHIFT-MASQUERADE`, `OPENSHIFT-MASQUERADE-2`, `OPENSHIFT-FIREWALL-FORWARD`)

pkg/network/node/node.go
```diff
@@ -136,13 +134,11 @@ func New(c *OsdnNodeConfig) (*OsdnNode, error) {
 	case networkutils.NetworkPolicyPluginName:
 		policy = NewNetworkPolicyPlugin()
 		pluginId = 2
-		minOvsVersion = "2.6.0"
```
danwinship, line 139: you are now never setting `minOvsVersion` and that code should go away too. Just pull https://github.com/danwinship/sdn/commit/8e3a59b8 in this branch before this commit.

pkg/network/node/node.go
```diff
@@ -370,9 +365,6 @@ func (node *OsdnNode) Start() error {
 			return err
 		}
 	}
-	if !node.useConnTrack {
-		node.watchServices()
```
danwinship, line 374: This makes `watchServices` dead code, which then makes a bunch of other things dead code. See https://github.com/danwinship/sdn/commit/fc0e4b06

pkg/network/node/ovscontroller.go
```diff
@@ -97,9 +96,7 @@ func (oc *ovsController) SetupOVS(clusterNetworkCIDR []string, serviceNetworkCID
 	otx := oc.ovs.NewTransaction()
 
 	// Table 0: initial dispatch based on in_port
-	if oc.useConnTrack {
-		otx.AddFlow("table=0, priority=300, ip, ct_state=-trk, actions=ct(table=0)")
-	}
+	otx.AddFlow("table=0, priority=300, ip, ct_state=-trk, actions=ct(table=0)")
```
danwinship, line 99: oh, while we're changing this anyway, there's a bug here: this was supposed to be the highest-priority rule in table 0, but there's a `priority=400` rule below. So can you bump this to `priority=1000`?

pkg/network/node/ovscontroller.go
```diff
@@ -177,13 +168,7 @@ func (oc *ovsController) SetupOVS(clusterNetworkCIDR []string, serviceNetworkCID
...
 
 	// Table 60: IP to service from pod
-	if oc.useConnTrack {
-		otx.AddFlow("table=60, priority=200, actions=output:2")
-	} else {
-		otx.AddFlow("table=60, priority=200, reg0=0, actions=output:2")
-		// vnid/port mappings; filled in by AddServiceRules()
-		// eg, "table=60, priority=100, reg0=${tenant_id}, ${service_proto}, nw_dst=${service_ip}, tp_dst=${service_port}, actions=load:${tenant_id}->NXM_NX_REG1[], load:2->NXM_NX_REG2[], goto_table:80"
-	}
+	otx.AddFlow("table=60, priority=200, actions=output:2")
```
danwinship, line 171: table 60 is dead now... again, check out the commit in my branch

juanluisvaladas:
Don't review this yet, I haven't done any of the changes you asked for yet. Just the rebase to see if it doesn't get garbage anymore

pkg/network/node/iptables.go
juanluisvaladas: Since I'm here, do I join the next append ChainArray too?
```
	// HACK: block access to MCS until we can secure it properly. Note that we share
	// the same chain between OUTPUT and FORWARD.
	chainArray = append(chainArray,
		Chain{
			table:    "filter",
			name:     "OPENSHIFT-BLOCK-OUTPUT",
			srcChain: "OUTPUT",
			srcRule:  []string{"-m", "comment", "--comment", "firewall overrides"},
			rules: [][]string{
				{"-p", "tcp", "-m", "tcp", "--dport", "22623", "-j", "REJECT"},
				{"-p", "tcp", "-m", "tcp", "--dport", "22624", "-j", "REJECT"},
			},
		},
		Chain{
			table:    "filter",
			name:     "OPENSHIFT-BLOCK-OUTPUT",
			srcChain: "FORWARD",
			srcRule:  []string{"-m", "comment", "--comment", "firewall overrides"},
			rules:    nil,
		},
	)
```

danwinship: The masquerading rules and MCS-blocking rules are unrelated, so it makes sense to have them be separate. As opposed to the masquerade stuff, where it was previously in two pieces just because we needed to support doing it two different ways.

juanluisvaladas:
It's probably better to wait for it to run the tests before actually reviewing, as this removes a big chunk of code I probably broke something.

The first two commits remain unchanged, so your previous review is valid. 
There are two additional commits,  the first one, `Drop non-conntrack-based Service support`, is removes everything related to table 60.

There is another new commit which only removes unused vars, I found this using a linter by accident and is entirely unrelated, but it's so small that I decided to sneak it in so that I wouldn't need to create a new bugzilla just to remove those two variables

juanluisvaladas:
The CNI fix related to the metrics already merged
/retest

juanluisvaladas:
rebased so that cluster-bot doesn't complain.

juanluisvaladas:
/test e2e-aws
@danwinship I'm pretty sure this is fine now.

pkg/network/node/node.go
```diff
@@ -137,13 +135,11 @@ func New(c *OsdnNodeConfig) (*OsdnNode, error) {
...
 		pluginId = 2
-		minOvsVersion = "2.6.0"
-		useConnTrack = true
 	default:
 		return nil, fmt.Errorf("Unknown plugin name %q", networkInfo.PluginName)
 	}
 
-	if useConnTrack && c.ProxyMode == kubeproxyconfig.ProxyModeUserspace {
+	if c.ProxyMode == kubeproxyconfig.ProxyModeUserspace {
 		return nil, fmt.Errorf("%q plugin is not compatible with proxy-mode %q", networkInfo.PluginName, c.ProxyMode)
```
danwinship, line 139: Argh. I had forgotten about this part. We still, in theory, allow configuring the userspace proxy if you are using multitenant mode.

_Maybe_ no one is using that in 4.x but I'm not sure we have any way to know.

Sigh. Sorry, I think the simplest thing is to just make it so that ovs-multitenant uses conntrack mode when proxymode is iptables, and non-conntrack mode when proxymode is userspace. So then you can't actually get rid of any of the non-conntrack code and this reverts back to mostly how it was in your first version.

danwinship (changes requested):

pkg/network/node/node.go
juanluisvaladas: I didn't know that either, I'm will close this PR and open a new one on Monday so that we have this code as a reference in the future if we decide

juanluisvaladas:
/close
opened https://github.com/openshift/sdn/pull/189 instead

PR #152, opened 2020-06-16
kube-proxy use node-ip to detect the IP family

aojea:
If the bind address is an unspecified address, 0.0.0.0 or ::,
kube-proxy uses the nodeIP to detect the IP Family.

It also fixes a minor bug where the IPv6 zero address was not
falling back to the node address.

Signed-off-by: Antonio Ojea <aojeagar@redhat.com>


aojea:
/assign @danwinship @dcbw 

upstream related PR https://github.com/kubernetes/kubernetes/pull/91725

danwinship:
well, openshift-sdn only supports single-stack IPv4, so this doesn't really matter, but if there's going to be code there thinking about the IP family at all, it should at least be right...

/lgtm

PR #153, opened 2020-06-25
Bug 1851182: Retry commiting OVS transactions in case of failure

alexanderConstantinescu:
There can be situations where we cannot program OVS flows because OVS has issues communicating with its unix socket, among other. We should retry these transactions, to make sure we don't loose out on programming OVS correctly because of "flakes". 

EDIT: I am not sure on how often / much we should retry these. Should we use a `DefaultRetry` or do it like this? 

/assign @dcbw @danwinship

### DIFF
--- a/pkg/network/node/ovs/ovs.go
+++ b/pkg/network/node/ovs/ovs.go
@@ -5,10 +5,11 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
-
-	"k8s.io/klog"
+	"time"
 
 	utilversion "k8s.io/apimachinery/pkg/util/version"
+	utilwait "k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog"
 	"k8s.io/utils/exec"
 )
 
@@ -100,6 +101,7 @@ type Transaction interface {
 const (
 	OVS_OFCTL = "ovs-ofctl"
 	OVS_VSCTL = "ovs-vsctl"
+	RETRIES   = 10
 )
 
 // ovsExec implements ovs.Interface via calls to ovs-ofctl and ovs-vsctl
@@ -424,9 +426,19 @@ func (tx *ovsExecTx) DeleteFlows(flow string, args ...interface{}) {
 }
 
 func (tx *ovsExecTx) Commit() error {
-	err := tx.ovsif.bundle(tx.flows)
-
-	// Reset flow context
-	tx.flows = []string{}
-	return err
+	defer func() {
+		tx.flows = []string{}
+	}()
+	backoff := utilwait.Backoff{
+		Duration: 500 * time.Millisecond,
+		Factor:   1.25,
+		Steps:    RETRIES,
+	}
+	return utilwait.ExponentialBackoff(backoff, func() (bool, error) {
+		err := tx.ovsif.bundle(tx.flows)
+		if err == nil {
+			return true, nil
+		}
+		return false, nil
+	})
 }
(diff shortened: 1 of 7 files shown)


danwinship:
Interesting. My initial thought was that you should fix `networkpolicy.go:syncFlows()` to not clear `npns.dirty` if `syncNamespaceFlows` returns an error. (After making `syncNamespaceFlows` return an error on error of course.) I guess both fixes are independently good.

pkg/network/node/ovs/ovs.go
```diff
@@ -424,9 +425,19 @@ func (tx *ovsExecTx) DeleteFlows(flow string, args ...interface{}) {
...
-	err := tx.ovsif.bundle(tx.flows)
-
-	// Reset flow context
-	tx.flows = []string{}
-	return err
+	defer func() {
+		tx.flows = []string{}
+	}()
+	retries := 30
+	for {
```
danwinship, line 432: The idea is good, but I'd use `utilwait.ExponentialBackoff` instead of doing this by hand.

alexanderConstantinescu:
EDIT: added a metric for OVS retries based on comment: https://bugzilla.redhat.com/show_bug.cgi?id=1851182#c1

Let's discuss it in this PR if that is really needed or if I should drop that commit or not. 

/cc @squeed

pkg/network/node/ovs/ovs.go
```diff
@@ -424,9 +440,20 @@ func (tx *ovsExecTx) DeleteFlows(flow string, args ...interface{}) {
...
+		Duration: 500 * time.Millisecond,
+		Factor:   1.25,
+		Steps:    RETRIES,
+	}
+	return utilwait.ExponentialBackoff(backoff, func() (bool, error) {
+		err := tx.ovsif.bundle(tx.flows)
+		if err == nil {
+			return true, nil
+		}
+		metrics.OVSRetries.Inc()
```
squeed, line 456: We want to catch the number of failed transactions, right? So this metric won't be right... it should only be incremented on the second (and further) retries.

alexanderConstantinescu: OK, not sure I follow on these concepts. But I take it you're saying: "have a counter incremented for every failed OVS commit, and have a go-routine sync the metric `OVSRetries` every second with that counter value", a bit like we're doing for `UpdateOVSMetrics`...?

squeed: Oh, I'm a dummy; this is after the return.

However, the metric is confusing; it's the number of failed OVS operations, not the number of retries. Can you update the metric to reflect that? And, while you're at it, add one for OVS successes?

alexanderConstantinescu: I changed the name to `OVSFailedOperations`, and added `OVSSuccessfulOperations`

squeed:
FYI, the convention is to add a CounterVec rather than a Counter. Then you'd add the label `result` and set that to either "success" or "failure" depending on the status.

squeed:
But the metrics are basically fine.

alexanderConstantinescu:
> FYI, the convention is to add a CounterVec rather than a Counter. Then you'd add the label `result` and set that to either "success" or "failure" depending on the status.

Hmm, I don't fully understand as the convention we seem to be using `CounterVec` and `Counter` almost equally frequently. Anyways, I just did what we currently do for `PodOperationsErrors`, but with the labels you mentioned.

smarterclayton:
Just a general note - if you are a controller, you need to ensure *something* is retrying forever up to a maximum backoff (i.e. we need to keep trying to fix broken things forever - nothing should ever be tried once).  Every component should be checking and fixing everything every X minutes, if your retry loop backs off up to that limit, that's fine.

PR #155, opened 2020-07-01
Bug 1853197: [release-4.4] Backport OVS-retries

alexanderConstantinescu:
4.4 backport of #153 and #159 

/assign @dcbw @JacobTanenbaum

PR #154, opened 2020-07-01
Bug 1853193: [release-4.5] Backport OVS-retries

alexanderConstantinescu:
4.5 backport of https://github.com/openshift/sdn/pull/153 and https://github.com/openshift/sdn/pull/159

/assign @juanluisvaladas @rcarrillocruz

PR #156, opened 2020-07-01
Bug 1853198: [release-4.3] Backport OVS-retries

alexanderConstantinescu:
4.3 backport of #153 and #159

/assign @danwinship @aojea

### DIFF
--- a/pkg/network/node/metrics/metrics.go
+++ b/pkg/network/node/metrics/metrics.go
@@ -1,6 +1,6 @@
 // +build linux
 
-package node
+package metrics
 
 import (
 	"fmt"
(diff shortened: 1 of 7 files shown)


alexanderConstantinescu:
@danwinship PTAL

pkg/network/node/metrics/metrics.go
```diff
@@ -1,6 +1,6 @@
 // +build linux
 
-package node
+package metrics
```
danwinship, line 3: To the extent possible, you should be backporting only the new change you made. So don't backport moving `metrics.go` into a different directory

pkg/network/node/metrics/metrics.go
```diff
@@ -44,6 +49,15 @@ var (
 			Help:      "Number of Open vSwitch flows",
 		},
 	)
+	OVSOperationsResult = metrics.NewCounterVec(
+		&metrics.CounterOpts{
```
danwinship, line 53: This should be using `prometheus.`, not `metrics.`, like the other metrics in this file. (Apparently at some point the dependency got moved around or whatever? But you want to be using the same version of it as everything else in 4.3.)

pkg/network/node/metrics/metrics.go
alexanderConstantinescu: All back-port included this exact patch:

https://github.com/openshift/sdn/pull/155/files
https://github.com/openshift/sdn/pull/154/files

So I am not sure what you're referring to?

alexanderConstantinescu:
@danwinship : I've addressed your comment. PTAL

PR #157, opened 2020-07-06
Bug 1852119: an iptables proxier sync fails, retry after iptablesSyncPeriod

juanluisvaladas:
Backport of https://github.com/kubernetes/kubernetes/pull/83387

### DIFF
--- a/glide.diff
+++ b/glide.diff
@@ -113,9 +113,17 @@ diff --no-dereference -N -r current/vendor/k8s.io/kubernetes/pkg/kubelet/kubelet
 ---
 > func (kl *Kubelet) syncNetworkUtil() {}
 diff --no-dereference -N -r current/vendor/k8s.io/kubernetes/pkg/proxy/iptables/proxier.go updated/vendor/k8s.io/kubernetes/pkg/proxy/iptables/proxier.go
+35c35
+< 	v1 "k8s.io/api/core/v1"
+---
+> 	"k8s.io/api/core/v1"
 191d190
 < 	changesPending       bool
-327,333c326
+193d191
+< 	syncPeriod           time.Duration
+305d302
+< 		syncPeriod:               syncPeriod,
+329,335c326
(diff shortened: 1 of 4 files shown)


juanluisvaladas:
I've seen in the prometheus metrics of the e2e test metric is being updated timely.

juanluisvaladas:
After adding PR 83387 we can see a large improvement since we change from having usually a few dozen restarts of the sdn pods to just two in the e2e-aws and 3 in the e2e-aws-upgrade.

But it still had 3 restarts that must be investigated.

danwinship:
as mentioned in https://github.com/openshift/sdn/pull/112#issuecomment-654461534, this _should_ be a no-op...

danwinship:
/hold
@juanluisvaladas can you explain what this is doing? As mentioned in #112, the flow without this PR should be:

1. `syncProxyRules` is called
2. `proxier.changesPending` is `true`, so it skips the early exit (L704) and prepares an iptables update
3. `iptables-restore` fails, so `syncProxyRules` returns at L1454 without having unset `changesPending`
4. (SDN may become unhealthy, since iptables-restore has failed.)
5. `BoundedFrequencyRunner` calls `maybeSyncProxyRules` again after `maxInterval`, which is `syncPeriod`
6. `proxier.changesPending` is still `true`, so we do another iptables update, and hopefully succeed this time
7. Assuming we succeed, `syncProxyRules` clears `changesPending` (L1456) and updates health/metrics timestamps (L1474).
8. (SDN is now healthy again.)

AFAICT, with your patch, the sequence of events doesn't change at all; at step 3, `syncProxyRules` calls `RetryAfter(syncPeriod)`, but that just changes step 4 to "`BoundedFrequencyRunner` calls `maybeSyncProxyRules` again after `retryInterval`, which is `syncPeriod`". ie, either way, the next call to `maybeSyncProxyRules` is after `syncPeriod`. (In git master, `RetryAfter` is needed because the bfr's `maxInterval` is `time.Hour`, not `syncPeriod`.)

PR #158, opened 2020-07-08
Add support for --may-exist when adding the bridge in sdn

abhat:
There is a race condition between ovs and sdn pods. When sdn pod starts, it tries to check if we are `alreadySetup` by checking for 

1. tun0 link being present
2. vxlan port being present
3. resourceVersion table having a len of 1 flow.

If any of these conditions fail, we delete the bridge, and recreate it including syncing all existing pods by collating data from the db and the API server. Imagine a sequence where:

0) SDN has cached ports and interface data in `existingOFPodNetworkInfo`
1) SDN deletes the bridge in `SetupOVS()`
2) ovs daemonset script adds the bridge as a part of save/restore flows.
3) SDN tries to add the bridge, and when trying to create it fails.

At step 3. the pod restarts because the setup couldn't be completed and with that we lose all the cached ports and interface information. Now when the pod restarts, we have nothing in ovsdb and that causes the said problem we are seeing with bug: 1852618

The fix here is to now allow AddBridge to fail by passing `--may-exist` in the args list to `ovs-vsctl`

### DIFF
--- a/pkg/network/node/ovs/ovs.go
+++ b/pkg/network/node/ovs/ovs.go
@@ -19,14 +19,13 @@ import (
 type Interface interface {
 	// AddBridge creates the bridge associated with the interface, optionally setting
 	// properties on it (as with "ovs-vsctl set Bridge ..."). If the bridge already
-	// exists this errors.
+	// exists, this will NOT result in error since it sets --may-exist internally.
 	AddBridge(properties ...string) error
 
-	// DeleteBridge deletes the bridge associated with the interface. The boolean
-	// that can be passed determines if a bridge not existing is an error. Passing
-	// true will delete bridge --if-exists, passing false will error if the bridge
-	// does not exist.
-	DeleteBridge(ifExists bool) error
+	// DeleteBridge deletes the bridge associated with the interface. This always
+	// calls del-br by passing the --if-exists flag so the resulting call
+	// will not error out if the bridge doesn't exist
+	DeleteBridge() error
 
 	// AddPort adds an interface to the bridge, requesting the indicated port
 	// number, and optionally setting properties on it (as with "ovs-vsctl set
@@ -206,7 +205,7 @@ func validateColumns(columns ...string) error {
 }
 
 func (ovsif *ovsExec) AddBridge(properties ...string) error {
-	args := []string{"add-br", ovsif.bridge}
+	args := []string{"--may-exist", "add-br", ovsif.bridge}
 	if len(properties) > 0 {
 		if err := validateColumns(properties...); err != nil {
 			return err
@@ -218,12 +217,8 @@ func (ovsif *ovsExec) AddBridge(properties ...string) error {
 	return err
 }
 
-func (ovsif *ovsExec) DeleteBridge(ifExists bool) error {
-	args := []string{"del-br", ovsif.bridge}
-
-	if ifExists {
-		args = append([]string{"--if-exists"}, args...)
-	}
+func (ovsif *ovsExec) DeleteBridge() error {
+	args := []string{"--if-exists", "del-br", ovsif.bridge}
 	_, err := ovsif.exec(OVS_VSCTL, args...)
 	return err
 }
--- a/pkg/network/node/ovscontroller.go
+++ b/pkg/network/node/ovscontroller.go
@@ -71,7 +71,7 @@ func (oc *ovsController) AlreadySetUp(vxlanPort uint32) bool {
 }
 
 func (oc *ovsController) SetupOVS(clusterNetworkCIDR []string, serviceNetworkCIDR, localSubnetCIDR, localSubnetGateway string, mtu uint32, vxlanPort uint32) error {
-	err := oc.ovs.DeleteBridge(true)
+	err := oc.ovs.DeleteBridge()
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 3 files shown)


pkg/network/node/ovscontroller.go
```diff
@@ -71,11 +71,7 @@ func (oc *ovsController) AlreadySetUp(vxlanPort uint32) bool {
 }
 
 func (oc *ovsController) SetupOVS(clusterNetworkCIDR []string, serviceNetworkCIDR, localSubnetCIDR, localSubnetGateway string, mtu uint32, vxlanPort uint32) error {
-	err := oc.ovs.DeleteBridge(true)
```
abhat, line 74: Why do we do `DeleteBridge` in SetupOVS? This causes all the vports in ovsdb to get deleted. and then when GetPodNetworkInfo is called, there is nothing in the db. are we expected to "re-add" the vports as the API server sync happens for already existing pod?

abhat: `del-br br0` basically deletes the entire tree under the bridge.

danwinship: At this point we have already determined that something is wrong with `br0` and we can't use it as-is. We don't have code to synchronize it to a working state, so we have to destroy and recreate it. The ports get re-added later, in `reattachPods` in `pkg/network/node/node.go`

abhat: aha, the problem is `reattachPods` relies on the `existingOFPodNetworks` to be there, the cache for which should have been built before the `DeleteBridge()` then. But right now we nuke the db when we delete the bridge, so the `existingOFPodNetworks` is basically an empty slice

```
 if networkChanged && len(existingOFPodNetworks) > 0 {
                err := node.reattachPods(existingPodSandboxes, existingOFPodNetworks)
                if err != nil {
                        return err
                }
 }
```

alexanderConstantinescu:
Just, if we always want to do `--may-exist` why not remove the argument?

abhat:
> Just, if we always want to do `--may-exist` why not remove the argument from `AddBridge`?
I am simply supporting what the API can. I agree there is little value to providing the argument in the contexts in which we use it. But this is for the adventurous folks who want to avoid the non-zero overhead of first checking if the bridge exists. It makes sense when there is a single controller/script talking to ovsdb.

danwinship:
>  _No description provided._

:frowning_face: 

Why do we need this? What is failing? Where? Why is this the correct fix?

rcarrillocruz:
I don't understand why parameterizing a function and then always passing same value ?

abhat:
> > _No description provided._
> 
> ☹️
> 
> Why do we need this? What is failing? Where? Why is this the correct fix?

Updated the description. Sorry, I should have written that up prior to marking it ready for review.

abhat:
> I don't understand why parameterizing a function and then always passing same value ?

https://github.com/openshift/sdn/pull/158#issuecomment-656139734

anuragthehatter:
@abhat Does this change addresses situation when br0 gets re-created but flows are still missing? QE reproducer was able to repro two issues:

1) br0 never gets re-created
2) br0 gets re-created but flows are still missing.

rcarrillocruz:
> > I don't understand why parameterizing a function and then always passing same value ?
> 
> [#158 (comment)](https://github.com/openshift/sdn/pull/158#issuecomment-656139734)

Yeah, I still don't get it. Just because the API supports we don't have to open up the parameters to functions.
If this was a library or code that others would use I would support it, but in the end this code and repo is just for OpenShift consumption and we are the users.

alexanderConstantinescu:
> > > I don't understand why parameterizing a function and then always passing same value ?
> > 
> > 
> > [#158 (comment)](https://github.com/openshift/sdn/pull/158#issuecomment-656139734)
> 
> Yeah, I still don't get it. Just because the API supports we don't have to open up the parameters to functions.
> If this was a library or code that others would use I would support it, but in the end this code and repo is just for OpenShift consumption and we are the users.

I agree with this.

danielmellado:
> I agree with this.

Same here, I don't really see the use of open a new parameter if it would always have the same value, unless we'd be speaking about a library which could be used in that way in the future. Summarizing, is there any case where we'd expect that value to be false? If no, I'd just put that directly in the code and remove the parameter.

pkg/network/node/ovs/ovs.go
```diff
@@ -214,6 +214,9 @@ func (ovsif *ovsExec) AddBridge(properties ...string) error {
 		args = append(args, "--", "set", "Bridge", ovsif.bridge)
 		args = append(args, properties...)
 	}
+	if mayExist {
+		args = append([]string{"--may-exist"}, args...)
+	}
```
danielmellado, line 219: As commented before, I'd just add this before along the another args and drop the argument ;)

abhat:
> @abhat Does this change addresses situation when br0 gets re-created but flows are still missing? QE reproducer was able to repro two issues:
> 
>     1. br0 never gets re-created
> 
>     2. br0 gets re-created but flows are still missing.

Yes, it should address both of them.

pkg/network/node/ovs/ovs.go
abhat: ack

abhat:
> > > > I don't understand why parameterizing a function and then always passing same value ?
> > > 
> > > 
> > > [#158 (comment)](https://github.com/openshift/sdn/pull/158#issuecomment-656139734)
> > 
> > 
> > Yeah, I still don't get it. Just because the API supports we don't have to open up the parameters to functions.
> > If this was a library or code that others would use I would support it, but in the end this code and repo is just for OpenShift consumption and we are the users.
> 
> I agree with this.

Sure. If we as a client don't have a use for not passing `--may-exist`, might as well remove the patch.

PR #159, opened 2020-07-09
Retry all OVS executions, not only transactions

alexanderConstantinescu:
Introducing redundancy for OVS commands from the SDN. The previous PR I added only fixed `Transactions` called for flow setup mainly. It would be good to do this for anything and everything, to make sure nothing slips us by. 

/assign @abhat

### DIFF
--- a/pkg/network/node/ovs/ovs.go
+++ b/pkg/network/node/ovs/ovs.go
@@ -106,8 +106,13 @@ type Transaction interface {
 const (
 	OVS_OFCTL = "ovs-ofctl"
 	OVS_VSCTL = "ovs-vsctl"
-	RETRIES   = 10
 )
+
+var ovsBackoff utilwait.Backoff = utilwait.Backoff{
+	Duration: 500 * time.Millisecond,
+	Factor:   1.25,
+	Steps:    10,
+}
 
 // ovsExec implements ovs.Interface via calls to ovs-ofctl and ovs-vsctl
 type ovsExec struct {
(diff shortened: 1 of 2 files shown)


alexanderConstantinescu:
@openshift/openshift-team-sdn , could you please have a look at this? I would like this to get in so that I can back-port it together with: https://github.com/openshift/sdn/pull/154

pkg/network/node/ovs/ovs.go
```diff
@@ -109,6 +109,12 @@ const (
 	RETRIES   = 10
 )
 
+var OVSBackoff utilwait.Backoff = utilwait.Backoff{
+	Duration: 500 * time.Millisecond,
+	Factor:   1.25,
+	Steps:    RETRIES,
```
danwinship, line 115: No need to have a separate `RETRIES` any more

alexanderConstantinescu: Actually `RETRIES` is used here to make sure the tests passes: https://github.com/openshift/sdn/blob/master/pkg/network/node/ovs/ovs_test.go#L112

alexanderConstantinescu: Well, I am not sure if that's what you meant by "separate `RETRIES`". Could you clarify?

danwinship: ah... well, the test could just use `OVSBackoff.Steps` intead of `RETRIES`.
actually, `OVSBackoff` should be `ovsBackoff`; we don't want to export that variable

alexanderConstantinescu: I've addressed it, @danwinship . PTAL

PR #161, opened 2020-07-15
[wip] sdn: update to OVS 2.13

dcbw:
Soak test to see what (if anything) breaks.


dcbw:
Nevermind, we're using system OVS here hwich is 2.13 already

PR #164, opened 2020-07-17
update egress dns code for dual-stack...

danwinship:
This updates the egress DNS code to handle IPv6 and dual-stack. Um, not that we need that or anything, but we're going to need it in ovn-kubernetes and I happened to have this code lying around so let's just merge this before copying the code over to ovn-kubernetes...

/cc @JacobTanenbaum @juanluisvaladas

### DIFF
--- a/pkg/network/common/dns.go
+++ b/pkg/network/common/dns.go
@@ -36,6 +36,9 @@ type DNS struct {
 	nameservers []string
 	// DNS port
 	port string
+
+	// query timeout; overridden by tests
+	timeout time.Duration
 }
 
 func NewDNS(resolverConfigFile string) (*DNS, error) {
@@ -48,6 +51,7 @@ func NewDNS(resolverConfigFile string) (*DNS, error) {
 		dnsMap:      map[string]dnsValue{},
 		nameservers: filterIPv4Servers(config.Servers),
 		port:        config.Port,
+		timeout:     5 * time.Second,
 	}, nil
 }
 
@@ -136,7 +140,7 @@ func (d *DNS) getIPsAndMinTTL(domain string) ([]net.IP, time.Duration, error) {
 			dialServer = net.JoinHostPort(server, d.port)
 		}
 		c := new(dns.Client)
-		c.Timeout = 5 * time.Second
+		c.Timeout = d.timeout
 		in, _, err := c.Exchange(msg, dialServer)
 		if err != nil {
 			return nil, defaultTTL, err
--- a/pkg/network/common/dns.go
+++ b/pkg/network/common/dns.go
@@ -8,17 +8,17 @@ import (
 
 	"github.com/miekg/dns"
 
-	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
+	utilnet "k8s.io/utils/net"
 )
 
 const (
-	// defaultTTL is used if an invalid or zero TTL is provided.
-	defaultTTL = 30 * time.Minute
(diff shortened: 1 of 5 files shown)


pkg/network/common/dns.go
```diff
@@ -122,58 +133,110 @@ func (d *DNS) updateOne(dns string) (bool, error) {
...
+	if in == nil || err != nil {
+		return ips, ttl, err
+	}
+	if in.Rcode != dns.RcodeSuccess {
+		return ips, ttl, fmt.Errorf("failed to get a valid answer: %v", in)
+	}
+
+	for _, a := range in.Answer {
+		aTTL := int(a.Header().Ttl)
+		if aTTL < ttl && aTTL != 0 {
```
juanluisvaladas, line 155: This always returns the lowest TTL without considering if it's an A or AAAA record.

pkg/network/common/dns.go
```diff
@@ -122,58 +133,110 @@ func (d *DNS) updateOne(dns string) (bool, error) {
...
 	}
-	if ttl == 0 {
-		ttl = defaultTTL
+	// else dual stack
+	ips := []net.IP{}
+	ttl := defaultTTL
+	errs := make(chan error)
+	var mutex sync.Mutex
+
+	go func() {
```
juanluisvaladas, line 188: I'm a bit concerned about this blocking...
If one of the petition needs to wait 5 seconds for a retry then the other will be blocked regardless which means instead of having just one family outdated for 5 seconds you have both.

That's a pretty big change and I don't think it has a huge but could we consider add a fixme here?

juanluisvaladas:
/hold
Until Dan reads my comments, remove the hold at your will.

pkg/network/common/dns.go
danwinship: hm... all of the responses in `in.Answer` _should_ be the same type (`rtype`)... I don't remember if the server is allowed to return answers of the wrong type? Easy enough to be defensive here though.

pkg/network/common/dns.go
danwinship: Well... the really big problem would be if the DNS server is old and broken and ignores AAAA requests, in which case we'll time out every time. But for the use case here (egress firewall) we _have_ to wait for both answers so I'm not sure there's anything better we could do...

pkg/network/common/dns.go
danwinship: Actually, the code should probably be considered correct as-is; in particular, if there is a CNAME and an A, and the CNAME has a lower TTL than the A, then we need to re-query when the CNAME expires, not when the A expires. So it's best to just take the lowest TTL of all returned answers.

juanluisvaladas: You're right.

pkg/network/common/dns.go
juanluisvaladas: > But for the use case here (egress firewall) we have to wait for both answers so I'm not sure there's anything better we could do...

Small delays when updating the flows timely can cause a significant amount of drops: https://bugzilla.redhat.com/show_bug.cgi?id=1850060 and more often than not, A and AAAA records have different TTLs.

In other words, if we're doing the query because we need to update the A record and the AAAA query delays for whatever reason, we'll block the updates for the A record while we wait for an AAAA record which hasn't changed.

However, this involves modifying EgressDNS.Sync() and EgressDNS.GetNextQueryTime() (which already are getting rewritten for IPv4 alone https://github.com/openshift/sdn/pull/168) and that's a pretty big change for an issue that for all we know is only potential.

juanluisvaladas:
/hold cancel

@danwinship since you reviewed my comments there is no reason to hold this.
/test e2e-gcp

PR #165, opened 2020-07-22
Remove unused packages from Dockerfile

yselkowitz:


yselkowitz:
Is the test failure indicative of a problem, or just CI being flaky again?

PR #166, opened 2020-07-23
Bug 1855119: [release-4.4] Add support for --may-exist when adding the bridge in sdn

alexanderConstantinescu:
Back-port of #160 to 4.4

/assign @dcbw @abhat


alexanderConstantinescu:
/retitle Bug 1855119: [release-4.4] Add support for --may-exist when adding the bridge in sdn

PR #168, opened 2020-07-24
[WIP] Bug 1850060: Make egress network policy DNS queries parallel

juanluisvaladas:
Several customers reported timeouts after upgrading to 3.11.219 which
includes a backport of github.com/openshift/sdn/pull/72.

Because names are resolved sequentially, if a name takes a couple
seconds the names that should have been resolved during that interval
won't be updated.

By doing asynchronous name resolution we ensure that if a name takes
longer than usual the rest of dnsNames won't block.

### DIFF
--- a/pkg/network/common/dns.go
+++ b/pkg/network/common/dns.go
@@ -10,6 +10,9 @@ import (
 
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
+	utilvalidation "k8s.io/apimachinery/pkg/util/validation"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/klog"
 )
 
 const (
@@ -18,24 +21,27 @@ const (
 )
 
 type dnsValue struct {
+	// Domain name to be queried to the DNS server.
+	name string
 	// All IPv4 addresses for a given domain name
 	ips []net.IP
 	// Time-to-live value from non-authoritative/cached name server for the domain
 	ttl time.Duration
-	// Holds (last dns lookup time + ttl), tells when to refresh IPs next time
-	nextQueryTime time.Time
+	// Channel to notify the goroutine to exit
+	exit chan struct{}
 }
 
 type DNS struct {
 	// Protects dnsMap operations
 	lock sync.Mutex
 	// Holds dns name and its corresponding information
-	dnsMap map[string]dnsValue
-
+	dnsMap map[string]*dnsValue
 	// DNS resolvers
 	nameservers []string
 	// DNS port
 	port string
+	// Channel to notify the egress DNS component about changes in the ip address
+	Updates chan string
 }
 
 func NewDNS(resolverConfigFile string) (*DNS, error) {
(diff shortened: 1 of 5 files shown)


juanluisvaladas:
/hold
This will probably conflict with: https://github.com/openshift/sdn/pull/164 , since Dan filed it first let's allow that to merge earlier.

/assign @abhat @danwinship 
CC @JacobTanenbaum 

I already gave the IT folks a test build and they report that the timeouts are singnificantly lower. Still not perfect but a big improvement. The test is quite unrealistic so this may be good enough.

juanluisvaladas:
/retitle Bug 1850060: Make egress network policy DNS queries async

juanluisvaladas:
@danwinship I added a new separate commit with subject TMP for easier review, but that needs to be squashed with the first one before merging it.

There is a third commit because I found an entirely unrelated bug while making these changes, let me know if you want me to have that in a separate PR or something.

juanluisvaladas:
Also, I wonder if we should report an error at all when the dns isn't found in the map.
Since now this is expected we probably should discard reporting anything

pkg/network/common/dns.go
```diff
@@ -90,36 +103,77 @@ func (d *DNS) Delete(dns string) {
...
-	return d.updateOne(dnsName)
+	if err != nil {
+		res, ok := d.dnsMap[dns]
+		if !ok {
+			// Can happen if DNS.Delete was called in the middle of an operation.
+			// In such cases we want to abort the update operation.
+
+			// In this scenario we don't really need to report the error because
+			// the record is gone anyway.
+			return false, fmt.Errorf("DNS value not found in dnsMap for domain: %q", dns)
```
danwinship, line 133: you say you don't need to report the error, but then you report the error

juanluisvaladas: Oh, by the error I meant the`err` variable.

Other than the the comment are you OK with the rest of the changes? If so I'll squash the tmp commit and change the comment for: In this scenario we don't really need to report the getIPsAndMinTTL error because the record is gone anyway.

danwinship: I was hoping for a little more rewriting... if there was some way to not need to grab locks and then release them and then grab them again, then the code would all make a lot more sense. I just feel like if we merge this we're going to be back here again in a few weeks.

(I don't actually see any bugs in the current code, but then, I didn't notice any bugs in the previous code either.)

juanluisvaladas: You're making perfect sense.

I added yet another commit which makes things cleaner, not good yet, but they're a step (I think) in the right direction. Right now I removed the `d.blockName` function by setting the `dnsValue.updating` field to true in `d.GetNextQueryTime`, which means now the Update function doesn't need to lock, unlock and lock again.

There is still a problem with the locking, both `Update` and `Add` call `getIPsAndMinTTL` before locking, but I can't think of a way to avoid this and still make the query async.

PS: I'm adding additive commits to make review easier, or do you prefe just amended commits? do you have any preference?

juanluisvaladas: /hold
This is just wrong. After GetNextQueryTime is executed there is no guarnatee that Update is going to be called again. this is just wrong.

danwinship: > PS: I'm adding additive commits to make review easier, or do you prefe just amended commits? do you have any preference?

In this particular case I'd rather you just amended commits, because the problem is that the first version of the patch was confusing, so I'd rather compare the new less-confusing version against the original code than against the confusing intermediate version.

juanluisvaladas:
Unit tests are currently broken,  just pushed to check the e2e-test, don't review this just yet

juanluisvaladas:
/test e2e-aws-multitenant

I don't think there's any difference, but let's test it anyway

juanluisvaladas:
/close
We did async rather than parallel

PR #169, opened 2020-07-28
Bug 1848478:  Invalid egressCIDR value causes sdn pods to fail on startup

tssurya:
If the user sets an invalid egressCIDR value, the API accepts this
and sets it in the hostsubnet field making the user think all went
well. However in the background sdn emits a warning in the logs
saying it ignores the invalid hostsubnet value but the user does
not become aware of this. Upon restarting the sdn-node pod, the pod
fails to come up and barfs that the egressCIDR is invalid.

This patch fixes this by making the sdn-controller wipe out the
invalid egressCIDR/IP field once the validation fails. This is the
same mechanism that the controller uses upon start-up when it comes
across an invalid hostsubnet.

### DIFF
--- a/pkg/network/common/common_test.go
+++ b/pkg/network/common/common_test.go
@@ -7,6 +7,7 @@ import (
 
 	networkapi "github.com/openshift/api/network/v1"
 	kapi "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 )
 
@@ -267,6 +268,65 @@ func TestParseClusterNetwork(t *testing.T) {
 			}
 		} else {
 			if !strings.Contains(err.Error(), test.err) {
+				t.Fatalf("test %q: error did not match %q: %v", test.name, test.err, err)
+			}
+		}
+	}
+}
+
+func TestValidateHostSubnetEgress(t *testing.T) {
+	tests := []struct {
+		name string
+		hs   networkapi.HostSubnet
+		err  string
+	}{
+		{
+			name: "valid egress ip",
+			hs: networkapi.HostSubnet{
+				EgressIPs:   []networkapi.HostSubnetEgressIP{"10.0.0.10", "10.0.0.11"},
+				EgressCIDRs: []networkapi.HostSubnetEgressCIDR{"10.0.0.0/16"},
+				ObjectMeta:  metav1.ObjectMeta{Name: "any"},
+			},
+			err: "",
+		},
+		{
+			name: "valid egress cidr",
+			hs: networkapi.HostSubnet{
+				EgressIPs:   []networkapi.HostSubnetEgressIP{"10.0.0.10", "10.0.0.11"},
+				EgressCIDRs: []networkapi.HostSubnetEgressCIDR{"10.0.0.0/16"},
+				ObjectMeta:  metav1.ObjectMeta{Name: "any"},
+			},
+			err: "",
+		},
+		{
+			name: "invalid CIDR address",
+			hs: networkapi.HostSubnet{
+				EgressIPs:   []networkapi.HostSubnetEgressIP{"10.0.0.10", "10.0.0.11"},
+				EgressCIDRs: []networkapi.HostSubnetEgressCIDR{"10.139.125.80/27"},
+				ObjectMeta:  metav1.ObjectMeta{Name: "any"},
+			},
+			err: "Invalid",
+		},
+		{
+			name: "invalid egress ip",
+			hs: networkapi.HostSubnet{
+				EgressIPs:   []networkapi.HostSubnetEgressIP{"2001:0db8:85a3:0000:0000:8a2e:0370:7334"},
+				EgressCIDRs: []networkapi.HostSubnetEgressCIDR{"10.139.125.64/27"},
+				ObjectMeta:  metav1.ObjectMeta{Name: "any"},
+			},
+			err: "Invalid",
+		},
+	}
+	for _, test := range tests {
+		err := ValidateHostSubnetEgress(&test.hs)
+		if err == nil {
+			if len(test.err) > 0 {
+				t.Fatalf("test %q unexpectedly did not get an error", test.name)
+			}
+		} else {
+			if test.err != "" && !strings.Contains(err.Error(), test.err) {
+				t.Fatalf("test %q: error did not match %q: %v", test.name, test.err, err)
+			} else if test.err == "" {
 				t.Fatalf("test %q: error did not match %q: %v", test.name, test.err, err)
 			}
 		}
--- a/pkg/network/common/egressip.go
+++ b/pkg/network/common/egressip.go
@@ -179,6 +179,11 @@ func (eit *EgressIPTracker) handleAddOrUpdateHostSubnet(obj, _ interface{}, even
 	hs := obj.(*networkv1.HostSubnet)
 	klog.V(5).Infof("Watch %s event for HostSubnet %q", eventType, hs.Name)
 
+	if err := ValidateHostSubnetEgress(hs); err != nil {
+		utilruntime.HandleError(fmt.Errorf("Ignoring invalid HostSubnet %s: %v", HostSubnetToString(hs), err))
+		return
+	}
+
 	eit.UpdateHostSubnetEgress(hs)
 }
 
--- a/pkg/network/common/validation.go
+++ b/pkg/network/common/validation.go
@@ -152,6 +152,29 @@ func ValidateHostSubnet(hs *networkapi.HostSubnet) error {
 	}
 }
 
+// ValidateHostSubnetEgress checks if the user-maintained fields of hostsubnet are valid.
+func ValidateHostSubnetEgress(hs *networkapi.HostSubnet) error {
+	allErrs := validation.ValidateObjectMeta(&hs.ObjectMeta, false, path.ValidatePathSegmentName, field.NewPath("metadata"))
+
+	for i, egressIP := range hs.EgressIPs {
+		if _, err := validateIPv4(string(egressIP)); err != nil {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("egressIPs").Index(i), egressIP, err.Error()))
+		}
+	}
+
+	for i, egressCIDR := range hs.EgressCIDRs {
+		if _, err := validateCIDRv4(string(egressCIDR)); err != nil {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("egressCIDRs").Index(i), egressCIDR, err.Error()))
+		}
+	}
+
+	if len(allErrs) > 0 {
+		return allErrs.ToAggregate()
+	}
+
+	return nil
+}
+
 func cidrsOverlap(cidr1, cidr2 *net.IPNet) bool {
 	return cidr1.Contains(cidr2.IP) || cidr2.Contains(cidr1.IP)
 }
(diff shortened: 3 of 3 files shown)


aojea:
You should add unit tests in the corresponding files :smile:

pkg/network/common/validation.go
```diff
@@ -147,12 +147,18 @@ func ValidateHostSubnet(hs *networkapi.HostSubnet) error {
...
+			hs.EgressIPs = nil
+			break
 		}
 	}
 
 	for i, egressCIDR := range hs.EgressCIDRs {
 		if _, err := validateCIDRv4(string(egressCIDR)); err != nil {
 			allErrs = append(allErrs, field.Invalid(field.NewPath("egressCIDRs").Index(i), egressCIDR, err.Error()))
+			// if any one of the CIDRs is invalid, we reset the whole array
+			hs.EgressCIDRs = nil
```
aojea, line 160: I'm not familiar enough to have an opinion on this, is not too disruptive to reset everything, maybe filtering out the bad ones?
Anyway, validation does not seems the right place to do this

tssurya:
> You should add unit tests in the corresponding files

I know :) and I tried, but there are no corresponding files for subnets.go. I might have to start one or let me see if this stuff gets tested elsewhere.

pkg/network/common/validation.go
tssurya: > I'm not familiar enough to have an opinion on this, is not too disruptive to reset everything, maybe filtering out the bad ones?

I'm not familiar either, but the first thing I did was to reset just the one incorrect value, but because of the way we check for CIDR patterns, I cannot leave an empty string and wipe out just one CIDR. I'd have to leave the whole field empty.

> Anyway, validation does not seems the right place to do this

I agree that this is not the right place, but if I were to do this in the subnets file, I'd have to do the check again to specifically see if its egressCIDRs or IPs that are failing. I was hoping for more ideas in the review.

aojea: well, kube-proxy filters out the invalid IP addresses for some things
https://github.com/aojea/kubernetes/blob/90918a0a873f068b8be5d1e08f7eee8054d9c36e/pkg/proxy/service.go#L157-L184

it may be a possibility, filter the bad ones and log a warning in `pkg/network/master/subnets.go`

aojea:
> > You should add unit tests in the corresponding files
> 
> I know :) and I tried, but there are no corresponding files for subnets.go. I might have to start one or let me see if this stuff gets tested elsewhere.

don't worry for subnets_test.go by now, at least the ones that exist, `pkg/network/common/validation_test.go` and `pkg/network/common/egressip_test.go` to cover the new functionality and avoid regressions

pkg/network/common/validation.go
tssurya: hmm ok let me see if I push this back to `subnets.go` and stop making `validation.go` ugly

juanluisvaladas: > is not too disruptive to reset everything, maybe filtering out the bad ones?
Yes. you may define several egress cidrs per hostsubnet, reseting everything can provoke downtime in the whole namespace.

pkg/network/common/validation.go
```diff
@@ -147,12 +147,18 @@ func ValidateHostSubnet(hs *networkapi.HostSubnet) error {
 	for i, egressIP := range hs.EgressIPs {
 		if _, err := validateIPv4(string(egressIP)); err != nil {
 			allErrs = append(allErrs, field.Invalid(field.NewPath("egressIPs").Index(i), egressIP, err.Error()))
+			// if any one of the IPs is invalid, we reset the whole array
+			hs.EgressIPs = nil
```
juanluisvaladas, line 151: This can also cause downtime. Just remove the invalid ones.

tssurya: > This can also cause downtime. Just remove the invalid ones.

ack yea, I need to update this patch.

pkg/network/master/subnets.go
```diff
@@ -235,6 +235,13 @@ func (master *OsdnMaster) handleAddOrUpdateSubnet(obj, _ interface{}, eventType
 
 	if err := common.ValidateHostSubnet(hs); err != nil {
 		utilruntime.HandleError(fmt.Errorf("Ignoring invalid HostSubnet %s: %v", common.HostSubnetToString(hs), err))
+		// If it was either the egressCIDR or egressIP fields that was invalid, then running an update will restore
+		// the hostsubnet object by wiping out the invalid egressCIDR/IP fields.
+		hs, err = master.networkClient.NetworkV1().HostSubnets().Update(context.TODO(), hs, metav1.UpdateOptions{})
```
danwinship, line 240: In general we do not undo the user's changes, even when they're wrong. So remove this part.

pkg/network/common/validation.go
danwinship: It would be consistent with our behavior in other situations to just block all egress from the namespace in this case

pkg/network/master/subnets.go
tssurya: > In general we do not undo the user's changes, even when they're wrong. So remove this part.

but we do right ? On startup in the sdn-controller, we reset invalid hostsubnet values.

https://github.com/openshift/sdn/blob/100a929557f10fe3a00707fb4b8e4f677944571f/pkg/network/master/subnets.go#L109

That's why I was trying to do something similar for the sdn-worker pods as well.

tssurya:
> So the real fix here is that we need to have an admission controller...

hmm I'll have to dig in more to see the details of adding and admission controller.

Anyhow, in this particular case, we used to check the invalid egressCIDR values set by the users in 3.11. But in 4.x we simply don't do anything and let the values get set and then blowup when the sdn worker pods get restarted. So the user understandably feels like we caused a regression.

pkg/network/common/validation.go
tssurya: > It would be consistent with our behavior in other situations to just block all egress from the namespace in this case

hmm so I'll continue to just erase all the egress then.

juanluisvaladas: > It would be consistent with our behavior in other situations to just block all egress from the namespace in this case

 OK that's fair enough

pkg/network/master/subnets.go
tssurya: @danwinship : so if I don't undo the user changes, how do I fix this bug ? The `reconcileHostSubnet` function does not automatically update the hostsubnet even if I clear the field. So I'd have to do the update here or add a new if condition inside `reconcileHostSubnet`

danwinship: Hm... that code is probably wrong. Basically it's not talking into account the fact that HostSubnet is half system-maintained (subnet allocations) and half user-maintained (EgressIPs). We don't want to be reverting mistakes to the user-maintained part.

`reconcileHostSubnet` is only there to deal with reconciliation of the system-maintained part; we have to do extra work to allocate subnets for F5 routers in one case.

In general, the behavior of openshift-sdn egress IP handling is not well-defined if the input egress IP data is invalid. The failing-to-start is clearly wrong, and we need to fix that, but other than that (and the fact that we shouldn't try to "fix" the HostSubnets ourselves), it doesn't much matter exactly what we do.

So, fix whatever is making it not start up, and ensure we're logging an error _somewhere_ about the bad IP, but otherwise, if we aren't going to take the time to write an admission controller, then there's not much we can do.

tssurya: ack thanks Dan, I also spoke with @squeed , he also suggests to fix the failing-on-startup part specifically if its failing due to invalid user-maintained parts (egress-CIDRs/IPs).

danwinship: right; if the system-maintained parts are invalid then that would mean that we couldn't set up the local network correctly anyway. But if the user-maintained/egressIP parts are invalid, we can at least get up and running, we just won't be able to set up the expected egress IPs

tssurya: cool, I've updated this patch. So when the `ValidateHostSubnet` function fails it returns an ErrorList with invalid fields. If those errors contain either `host`/`hostIP`/`subnet` I allow the cluster to fail, else we just ignore it - log it and move forward.

Casey, Dan : let me know if this lgty .

pkg/network/node/subnets.go
```diff
@@ -148,7 +149,11 @@ func (node *OsdnNode) getLocalSubnet() (string, error) {
 		subnet, err = node.networkClient.NetworkV1().HostSubnets().Get(context.TODO(), node.hostName, metav1.GetOptions{})
 		if err == nil {
 			if err = common.ValidateHostSubnet(subnet); err != nil {
-				return false, err
+				// fail only if the system-maintained fields of hostsubnet are invalid.
+				if strings.Contains(err.Error(), "host") || strings.Contains(err.Error(), "hostIP") || strings.Contains(err.Error(), "subnet") {
```
danwinship, line 153: no... that's too fragile.

It's probably best to just make `ValidateHostSubnet` not validate the egress IP fields; the egress IP code doesn't use `ValidateHostSubnet` anyway, it does its own checking.

Hm... actually it mostly _doesn't_ do any checking. That needs to be fixed (maybe as a separate PR).

tssurya: ok yea, actually for me this whole `main hostsubnet watcher` and then the `egressip hostsubnet watcher` concepts are confusing (for me).

So here's what I'll do. I'll separate the validation functions into `ValidateHostSubnet` and `ValidateHostSubnetEgress`. I'll use the former in the `hostSubnetWatcher` and latter in the `EgressIPTracker` portions.

pkg/network/common/common_test.go
```diff
@@ -272,3 +272,56 @@ func TestParseClusterNetwork(t *testing.T) {
...
+			err: "Invalid",
+		},
+	}
+	for _, test := range tests {
+		err := ValidateHostSubnetEgress(&test.hs)
+		if err == nil {
+			if len(test.err) > 0 {
+				t.Fatalf("test %q unexpectedly did not get an error", test.name)
+			}
+		} else {
```
squeed, line 321: You have to explicitly test for the case where `test.err == "" && err != nil`, since `strings.Contains, "asdf", "") == true`

tssurya: ah good catch!

tssurya:
@danwinship : PTAL whenever you have some time.

pkg/network/common/common_test.go
```diff
@@ -272,3 +273,62 @@ func TestParseClusterNetwork(t *testing.T) {
 		}
 	}
 }
+
+func TestParseHostSubnetEgress(t *testing.T) {
```
danwinship, line 277: by convention this should be `Test` plus the name of the function it's testing, so `TestValidateHostSubnetEgress`

pkg/network/common/egressip.go
```diff
@@ -238,7 +243,12 @@ func (eit *EgressIPTracker) UpdateHostSubnetEgress(hs *networkv1.HostSubnet) {
 		node.requestedCIDRs = newRequestedCIDRs
 		node.parsedCIDRs = make(map[string]*net.IPNet)
 		for _, cidr := range hs.EgressCIDRs {
-			_, parsed, _ := net.ParseCIDR(string(cidr))
+			_, parsed, err := net.ParseCIDR(string(cidr))
+			if err != nil {
+				utilruntime.HandleError(fmt.Errorf("Could not parse HostSubnet %q CIDR: %v", hs.Name, err))
```
danwinship, line 248: This shouldn't be needed since you only call this function on validated HostSubnets.

tssurya: yea true, I added it in case someone decided to call `UpdateHostSubnetEgress` from elsewhere. But, I am sure that can be handled later on, this does not serve any purpose for now

pkg/network/common/common_test.go
tssurya: my bad, will change this as well.

PR #171, opened 2020-08-11
Bug 1859451: Add NodeInformer to EgressIP

danielmellado:
Currently, EgressIP falls out to another node if the node isn't pinging.
This commit also adds the ability of doing so by checking the node
status.

Signed-off-by: Daniel Mellado <dmellado@redhat.com>

### DIFF
--- a/pkg/network/master/egressip.go
+++ b/pkg/network/master/egressip.go
@@ -6,12 +6,14 @@ import (
 	"sync"
 	"time"
 
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/klog"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
 	utilwait "k8s.io/apimachinery/pkg/util/wait"
+	kcoreinformers "k8s.io/client-go/informers/core/v1"
 	"k8s.io/client-go/util/retry"
 
 	networkclient "github.com/openshift/client-go/network/clientset/versioned"
@@ -25,6 +27,7 @@ type egressIPManager struct {
 	tracker            *common.EgressIPTracker
 	networkClient      networkclient.Interface
 	hostSubnetInformer networkinformers.HostSubnetInformer
+	nodeInformer       kcoreinformers.NodeInformer
 
 	updatePending bool
 	updatedAgain  bool
@@ -35,6 +38,7 @@ type egressIPManager struct {
 
 type egressNode struct {
 	ip      string
+	name    string
 	offline bool
 	retries int
 }
@@ -100,7 +104,7 @@ func (eim *egressIPManager) maybeDoUpdateEgressCIDRs() (bool, error) {
 			if node := eim.monitorNodes[hs.HostIP]; node != nil {
 				monitorNodes[hs.HostIP] = node
 			} else {
-				monitorNodes[hs.HostIP] = &egressNode{ip: hs.HostIP}
+				monitorNodes[hs.HostIP] = &egressNode{ip: hs.HostIP, name: nodeName}
 			}
 
 			oldIPs := sets.NewString(common.HSEgressIPsToStrings(hs.EgressIPs)...)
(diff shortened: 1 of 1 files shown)


danielmellado:
Pull-request updated, HEAD is now 24cd5c4fea0bcaa7e82b02bf75e6c2c2cd579aa3

danwinship:
Is this in response to a bug report or something?

Node readiness tracks the state of kubelet on the node, but we're more concerned with the state of the SDN on that node, and in fact the state of kubelet doesn't matter at all as long as the SDN is up and running.

danielmellado:
Yep, this is in response to https://bugzilla.redhat.com/show_bug.cgi?id=1859451, so basically I'd like to track the node state as well, as it might happen the case where just like the bug

danielmellado:
Pull-request updated, HEAD is now 0e77b8fa08427e42d246a4d4f9398e3e416733a0

danielmellado:
Pull-request updated, HEAD is now 388effd46fe108f6731ff6bfff32cd8100b4660c

danielmellado:
Pull-request updated, HEAD is now 19491956e013a9827aa894059a68a32673d1c051

pkg/network/master/egressip.go
```diff
@@ -148,16 +152,17 @@ func (eim *egressIPManager) poll(stop chan struct{}) {
 		}
 
 		start := time.Now()
-		retry := eim.check(retry)
+		retry, _ := eim.check(retry)
```
squeed, line 155: If you're going to return an error, you should handle it. Don't just add it only to drop it on the floor.

pkg/network/master/egressip.go
```diff
@@ -170,6 +175,19 @@ func (eim *egressIPManager) check(retrying bool) bool {
...
 		}
 
+		nn, err := eim.nodeInformer.Lister().Get(node.name)
+		if err != nil {
+			return err
+		}
+
+		for _, cond := range nn.Status.Conditions {
+			if cond.Type == v1.TaintNodeNotReady {
+				klog.Warningf("Node %s is offline", node.ip)
```
squeed, line 185: suggestion: reword this to "Node %s is not Ready"

pkg/network/master/egressip.go
```diff
@@ -170,6 +175,19 @@ func (eim *egressIPManager) check(retrying bool) bool {
...
+				node.offline = true
+				eim.tracker.SetNodeOffline(node.ip, true)
```
squeed, line 187: I think you need to return here, or the node will go back online if the ping succeeds.

danielmellado:
Pull-request updated, HEAD is now 5e1cfe6059a1edadd6255998d797df1e00aca308

pkg/network/master/egressip.go
```diff
@@ -170,6 +177,21 @@ func (eim *egressIPManager) check(retrying bool) bool {
 			continue
 		}
 
+		nn, err := eim.nodeInformer.Lister().Get(node.name)
+		if err != nil {
+			return true, err
```
squeed, line 182: I would just warn on the error here and treat the node as up; if it's disappeared we will deal with that shortly.

danielmellado:
Pull-request updated, HEAD is now 265e89a3861bd2dc055da830344d08c818f5d861

pkg/network/master/egressip.go
```diff
@@ -25,6 +27,7 @@ type egressIPManager struct {
 	tracker            *common.EgressIPTracker
 	networkClient      networkclient.Interface
 	hostSubnetInformer networkinformers.HostSubnetInformer
+	nodeInformer       kcoreinformers.NodeInformer
```
aojea, line 30: this has to been initialized, plumbed down from the master initialization

danielmellado: This seems actually to be initalized in master, just check https://github.com/openshift/sdn/blob/master/pkg/network/master/master.go#L37. I'm wondering whether it'd be passed to the start function as well.

aojea: you are creating a new object that is using an informer of that type

```go
func newEgressIPManager() *egressIPManager {
	eim := &egressIPManager{}
	eim.tracker = common.NewEgressIPTracker(eim)
	return eim
}
func (eim *egressIPManager) Start(networkClient networkclient.Interface, hostSubnetInformer networkinformers.HostSubnetInformer, netNamespaceInformer networkinformers.NetNamespaceInformer) {
	eim.networkClient = networkClient
	eim.hostSubnetInformer = hostSubnetInformer
	eim.tracker.Start(hostSubnetInformer, netNamespaceInformer)
}
```
, but you are not cascading it down from the master
https://github.com/openshift/sdn/blob/578df042fa70429b3f40765513ee485f0dc49d60/pkg/network/master/master.go#L117-L118

danielmellado:
cherry-pick release-4.5

PR #172, opened 2020-08-11
Bug 1857743: Port stuck open when ep deleted before svc in unidling mode

tssurya:
When the ep is deleted before the svc in unidling proxy mode, the port
does not get deleted and remains stuck. This patch adds a new variable
pendingDeletion to hybrid proxier which keeps track of the svc if the
ep is the one that gets deleted first.

### DIFF
--- a/pkg/network/proxyimpl/hybrid/proxy.go
+++ b/pkg/network/proxyimpl/hybrid/proxy.go
@@ -55,6 +55,10 @@ type HybridProxier struct {
 
 	// A gate that prevents iptables from being exhausted by proxy calls.
 	syncRunner *async.BoundedFrequencyRunner
+
+	// This map is used in unidling proxy mode to ensure the service is correctly deleted
+	// if it's deletion occurs after the deletion of the endpoint.
+	pendingDeletion map[types.NamespacedName]bool
 }
 
 func NewHybridProxier(
@@ -71,6 +75,7 @@ func NewHybridProxier(
 
 		usingUserspace:      make(map[types.NamespacedName]bool),
 		switchedToUserspace: make(map[types.NamespacedName]bool),
+		pendingDeletion:     make(map[types.NamespacedName]bool),
 	}
 
 	p.syncRunner = async.NewBoundedFrequencyRunner("sync-runner", p.syncProxyRules, minSyncPeriod, time.Hour, 4)
@@ -140,6 +145,21 @@ func (p *HybridProxier) OnServiceUpdate(oldService, service *corev1.Service) {
 	}
 }
 
+// cleanupState handles the deletion of endpoints and service in any order.
+// svcName should be the NamespacedName of a service (or endpoint).
+func (p *HybridProxier) cleanupState(svcName types.NamespacedName) {
+	_, isPendingDeletion := p.pendingDeletion[svcName]
+	if isPendingDeletion {
+		klog.V(6).Infof("hybrid proxy: removing %s/%s entry from pendingDeletion", svcName.Namespace, svcName.Name)
+		delete(p.pendingDeletion, svcName)
+		delete(p.usingUserspace, svcName)
+		delete(p.switchedToUserspace, svcName)
+	} else {
+		klog.V(6).Infof("hybrid proxy: adding %s/%s entry to pendingDeletion", svcName.Namespace, svcName.Name)
+		p.pendingDeletion[svcName] = true
+	}
+}
+
 func (p *HybridProxier) OnServiceDelete(service *corev1.Service) {
 	svcName := types.NamespacedName{
 		Namespace: service.Namespace,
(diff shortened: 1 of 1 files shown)


tssurya:
/assign @squeed 

@squeed : I know this is not exactly what you had in mind, but I just wanted to check and see if this would work or not.

PS: I know I'd have to add tests at some point.

pkg/network/proxyimpl/hybrid/proxy.go
```diff
@@ -153,9 +158,21 @@ func (p *HybridProxier) OnServiceDelete(service *corev1.Service) {
...
 	defer p.switchedToUserspaceLock.Unlock()
 
-	if isUsingUserspace, ok := p.usingUserspace[svcName]; ok && isUsingUserspace {
+	isUsingUserspace, ok := p.usingUserspace[svcName]
+	_, isPendingDeletion := p.pendingDeletion[svcName]
+
+	if (ok && isUsingUserspace) || isPendingDeletion {
 		klog.V(6).Infof("hybrid proxy: del svc %s/%s in unidling proxy", service.Namespace, service.Name)
 		p.unidlingProxy.OnServiceDelete(service)
+		if isPendingDeletion {
```
tssurya, line 167: I guess I could chalk this out to a separate function since the code is duplicated here and below.

squeed:
I had a bit of a different fix in mind:

In both `OnServiceDelete` and `OnEndpointsDelete`, we put an entry in `pendingDeletions`. If the entry is already there, then we delete the values in `usingUserspace` and `switchedService`. Then we don't have to change the logic anywhere else.

tssurya:
> I had a bit of a different fix in mind:
> 
> In both `OnServiceDelete` and `OnEndpointsDelete`, we put an entry in `pendingDeletions`. 

Q: but you mean only in the `unidlingproxy` mode right ? 

> If the entry is already there, then we delete the values in `usingUserspace` and `switchedService`. Then we don't have to change the logic anywhere else.

hmm yea I wanted to do it the way you said but I got a bit confused on the local cache deletion parts since `OnServiceDelete` uses the `switchedService` map and  `OnEndpointsDelete` uses the `usingUserspace` map and then got confused. But... wait! ah I think I get it now. So what you meant was:

- when we go to `OnServiceDelete` we add  an entry in `pendingDeletions[svcName]`. 
- when we go to `OnEndpointsDelete` we check if the entry is already present and if yes we delete it from both caches since that would clear endpoints and svc. It doesn't really matter even if we deleted both caches which is what is expected.
- same for viceversa.


What I basically do now, is fix what was broken which is calling the svc deletion in  `OnServiceDelete` in `unidlingproxy` mode if the endpoint got deleted already. I think this should also work since technically all this `unidlingproxy.deletion `takes place only when you idle and unidle. But anyhow I guess I'd have to test this properly.

squeed:
> * when we go to `OnServiceDelete` we add  an entry in `pendingDeletions[svcName]`.
> 
> * when we go to `OnEndpointsDelete` we check if the entry is already present and if yes we delete it from both caches since that would clear endpoints and svc. It doesn't really matter even if we deleted both caches which is what is expected.
> 
> * same for viceversa.

Exactly. I'd just write a `cleanupState(nsName)` function that takes care of everything, then call it in `OnServiceDelete` and `OnEndpointsDelete`. That way we don't have to worry about ordering at all.

tssurya:
> Exactly. I'd just write a `cleanupState(nsName)` function that takes care of everything, then call it in `OnServiceDelete` and `OnEndpointsDelete`. That way we don't have to worry about ordering at all.

cool, I've tried it your way and also err I am not really sure how/where to add the tests for this bug.

pkg/network/proxyimpl/hybrid/proxy.go
```diff
@@ -140,6 +145,19 @@ func (p *HybridProxier) OnServiceUpdate(oldService, service *corev1.Service) {
 	}
 }
 
+func (p *HybridProxier) CleanupState(svcName types.NamespacedName) {
```
squeed, line 148: this should be a private function, so lowercase this name.

squeed: also, add a short docblock explaining what this does (handle endpoint and service deletions in any order)

squeed:
Two very minor things, then this is good to go.

pkg/network/proxyimpl/hybrid/proxy.go
tssurya: > also, add a short docblock explaining what this does (handle endpoint and service deletions in any order)

ack yea good point, I always forget to do this.

tssurya:
CI failing due to:
`Waiting for container artifacts in pod release-latest in namespace`

This failure is being investigated by the testplatform team.

PR #178, opened 2020-08-25
[release 4.2] Bug 1801742: Multiple iptables improvements

danwinship:
replacement for #112


danwinship:
> @danwinship: This pull request references [Bugzilla bug 1801742](https://bugzilla.redhat.com/show_bug.cgi?id=1801742), which is invalid:
> 
>     * expected the bug to be open, but it isn't

well fine then

PR #185, opened 2020-09-09
Add documentation to openshift/sdn

danielmellado:
This commit adds sphinx documentation, alongside a Dockerfile for
building it, to openshift/sdn repository.

It mainly has the current wiki as a source, and it aims to be blend
together with the code in order for it to be updated together.


docs/source/openshift_sdn/README.md
```diff
@@ -0,0 +1,66 @@
+This is openshift-sdn, the default network plugin for OpenShift (both
+OKD and OCP). It uses Open vSwitch to connect pods locally, with VXLAN
+tunnels to connect different nodes.
+
+OpenShift SDN is designed to be installed by the [OpenShift Network
+Operator](https://github.com/openshift/cluster-network-operator), and
```
juanluisvaladas, line 6: I think everything after the comma should be removed.

docs/source/openshift_sdn/README.md
```diff
@@ -0,0 +1,66 @@
...
+certain components of it (such as the Deployment and DaemonSet
+objects) are found there.
+
+This module defines two images: the sdn image, which contains OpenShift SDN
+(both controller and node components) and the kube-proxy image, which is
+deployed by the Network Operator for third-party network plugins that need it.
+(Kube-proxy is built from here rather than from the origin repo so
```
juanluisvaladas, line 13: If there previous sentence is finished with a "." I don't think we actually need the parenthesis

docs/source/openshift_sdn/README.md
```diff
@@ -0,0 +1,66 @@
...
+Network Operator creates the CRD definitions.
+
+## The OpenShift SDN Controller
+
+The network controller is run on
+the masters to handle cluster-level processing:
+
+  - Creating `NetNamespace` objects corresponding to `Namespace`s
+  - Creating `HostSubnet` objects corresponding to `Node`s
+  - Implementing high availability for egress IPs
```
juanluisvaladas, line 37: Don't you mean egressCIDRs ?

docs/source/openshift_sdn/workflow.rst
```diff
@@ -0,0 +1,115 @@
...
+Just query cluster-bot and send him a message with the following::
+
+  launch openshift/sdn#<PR_NUMBER> aws
+
+Update your CNO image without deleting your cluster
+---------------------------------------------------
+
+.. code-block:: shell-session
+
+  update_cno ()
```
juanluisvaladas, line 89: you're not updating it, you're making it unmanaged, a better name would be unmanage_cno.

juanluisvaladas (changes requested):
I'm not quite sure about this making sense here rather than the CNO. Or even if this should be part of CNO wiki...

danwinship:
I guess I don't get how this is supposed to be an improvement over just having the markdown files? Why would I want to build a container and run a web server locally when I can just read the files, or view them off github.com?

docs/source/openshift_sdn/README.md
danielmellado: This content is just migrated, please don't go over it yet.

docs/source/openshift_sdn/README.md
danielmellado: ditto

docs/source/openshift_sdn/README.md
danielmellado: ditto

docs/source/openshift_sdn/workflow.rst
danielmellado: ditto

danielmellado:
> I guess I don't get how this is supposed to be an improvement over just having the markdown files? Why would I want to build a container and run a web server locally when I can just read the files, or view them off github.com?

Hi Dan, so overall let me give you some context. 

1) This content will be published to github pages once we agree on a proper location.

2) Sphinx [1], which is used by a lot of open source projects allows for a lot of functionalities, besides just having the files over github.

As a matter of fact, you can just go over the files themselves even if you don't want to run the container, which is just being provided as a commodity in order for a developer to be able to trigger such builld. Enhanced funcionality goes from search to code It also does provide a way to organize the content, which IMHO is lacking in just 'bare' markdown.

3) Having the documentation in-tree also allows us to do a proper branching and tracking of it per-release.

[1] https://www.sphinx-doc.org/en/master/

PR #186, opened 2020-09-12
Bug 1878731: vxlan_monitor_test: fix flake

danwinship:
The test was trying to take advantage of the fact that the updates were guaranteed to happen in order, but forgot that they weren't guaranteed to be _returned_ in order...

/assign @juanluisvaladas

### DIFF
--- a/pkg/network/node/vxlan_monitor_test.go
+++ b/pkg/network/node/vxlan_monitor_test.go
@@ -254,20 +254,19 @@ func TestEgressVXLANMonitor(t *testing.T) {
 	}
 
 	update := peekUpdate(updates, evm)
-	// Because we're doing the node updates sequentially we can expect
-	// the nodes to be marked offline in a specific order.
-	// It doesn't really matter, but it's unexpected enough to raise an error.
 	if update == nil {
 		t.Fatalf("Nodes failed to go offline")
-	} else if update[0].nodeIP != "192.168.1.1" {
-		t.Fatalf("Expected update[0] to be 192.168.1.1: %#v", update)
-	} else if !update[0].offline {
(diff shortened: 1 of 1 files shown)


pkg/network/node/vxlan_monitor_test.go
```diff
@@ -254,20 +254,19 @@ func TestEgressVXLANMonitor(t *testing.T) {
...
 	if update == nil {
 		t.Fatalf("Nodes failed to go offline")
-	} else if update[0].nodeIP != "192.168.1.1" {
-		t.Fatalf("Expected update[0] to be 192.168.1.1: %#v", update)
+	} else if len(update) != 2 {
+		t.Fatalf("Check return wrong number of updates %#v", update)
+	}
+
+	// GetUpdates returns the updates in a random order
+	if !(update[0].nodeIP == "192.168.1.1" && update[1].nodeIP == "192.168.1.3") &&
```
aojea, line 264: should this be an OR? if the first is true the second is false, no?

danwinship: Not sure which `&&` you're saying should be an `||`, but the condition is correct. If the first parenthesized expression is `true` (update 0 is 192.168.1.1 and update 1 is 192.168.1.3) then we are ok and don't want to hit the `t.Fatalf`, and if the second parenthesized expression is `true` (update 0 is 192.168.1.3 and update 1 is 192.168.1.1) then we are ok and don't want to hit the `t.Fatalf`. Only if both parenthesized expressions are `false` do we want to fail the test. So `if !(X) && !(Y) { t.Fatalf(...) }`

aojea: I got it now sorry, thanks for the clarification

PR #189, opened 2020-09-14
 Bug 1762580: Enable conntrack for ovs-multitenant unless userspace proxy

juanluisvaladas:
Should I do the same for ovs-subnet as well? I think it makes sense but I doubt anyone actually cares about this in ovs-subnet.

/assign @danwinship


juanluisvaladas:
@danwinship can you PTAL?

PR #194, opened 2020-09-19
Bug 1877794: Emit a warning and ignore if empty cidr is passed to kubeproxy

tssurya:
Currently we are directly using the NewDetectLocalByCIDR function of the
kubeproxy to detect if the traffic is local or not. Kubeproxy does
not support multiple-cidr-ranges. Due to this if a user specifies
multiple-cidr-ranges, CNO simply leaves the kubeproxy.cluster-cidr field
empty. In 4.4 kubeproxy used to emit a warning if an empty cluster-cidr
was passed and ignore the field. In 4.5 kubeproxy still emits a warning
if an empty cluster-cidr is passed. It does this in the getLocalDetector
function. Since getLocalDetector is not exported, we cannot leverage that.
So when we use the NewDetectLocalByCIDR function directly, sdn pods go into
crashloopbackoff because of the empty cidr string.

This patch mimics what kubeproxy does in case of empty cidr string. It emits
a warning and uses the NewNoOpLocalDetector. This is the simplest backportable
fix that does not require us to rebase on the new vendor version of kubeproxy.

The long term plan is to replace detection of local traffic using cidr
with interfaces (upstream implementation in progress).

### DIFF
--- a/pkg/openshift-sdn/proxy.go
+++ b/pkg/openshift-sdn/proxy.go
@@ -109,9 +109,15 @@ func (sdn *OpenShiftSDN) runProxy(waitChan chan<- bool) {
 			klog.Fatalf("Unable to read IPTablesMasqueradeBit from config")
 		}
 
-		localDetector, err := proxyutiliptables.NewDetectLocalByCIDR(sdn.ProxyConfig.ClusterCIDR, iptInterface)
-		if err != nil {
-			klog.Fatalf("Unable to configure local traffic detector: %v", err)
+		var localDetector proxyutiliptables.LocalTrafficDetector
+		if sdn.ProxyConfig.ClusterCIDR == "" {
+			klog.Warningf("Kubeproxy does not support multiple cluster CIDRs, configuring no-op local traffic detector")
+			localDetector = proxyutiliptables.NewNoOpLocalDetector()
+		} else {
+			localDetector, err = proxyutiliptables.NewDetectLocalByCIDR(sdn.ProxyConfig.ClusterCIDR, iptInterface)
+			if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/openshift-sdn/proxy.go
```diff
@@ -109,9 +109,15 @@ func (sdn *OpenShiftSDN) runProxy(waitChan chan<- bool) {
 			klog.Fatalf("Unable to read IPTablesMasqueradeBit from config")
 		}
 
-		localDetector, err := proxyutiliptables.NewDetectLocalByCIDR(sdn.ProxyConfig.ClusterCIDR, iptInterface)
-		if err != nil {
-			klog.Fatalf("Unable to configure local traffic detector: %v", err)
+		var localDetector proxyutiliptables.LocalTrafficDetector
+		if sdn.ProxyConfig.ClusterCIDR == "" {
+			klog.Warningf("detect-local-mode set to ClusterCIDR, but no cluster CIDR defined")
```
tssurya, line 114: need to change this to something that makes sense like multiple cidrs not supported, blah...

tssurya:
/hold cancel.

tested on cluster, looks ok:

```
I0922 12:07:19.310073  135791 proxy.go:106] Using unidling+iptables Proxier.
W0922 12:07:19.310098  135791 proxy.go:114] Kubeproxy does not support multiple cluster CIDRs, configuring no-op local traffic detector
I0922 12:07:19.310299  135791 proxy.go:144] Tearing down userspace rules.
```

knobunc:
Should this have CI to test the case?

tssurya:
> Should this have unit tests?

I tried to write, but couldn't find any interface for testing the proxy file.

PR #196, opened 2020-09-24
Bug 1882071: Fix bug in reflector not recovering from "Too large resource version"

JacobTanenbaum:
created a new fork of openshift/kubernetes for sdn-4.5-kubernetes-1.18.6 based on the 1.18.6 release of kubernetes 

This pulls in the right version of client-go to fix the issue


JacobTanenbaum:
@danwinship @p0lyn0mial one of the operands of the CNO that needs updated client-go

PR #197, opened 2020-09-28
Bug 1878787: vxlan_monitor_test: fix flake

danwinship:
Replaces #190, adds a commit to fix "make verify-deps" which got broken by a disappeared upstream module. (The two commits have nothing to do with each other but I don't feel like filing a bz to let me fix the verify-deps bug separately.)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -29,6 +29,7 @@ require (
 )
 
 replace (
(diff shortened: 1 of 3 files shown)


JacobTanenbaum:
could you add the version on the title of the PR?

PR #198, opened 2020-10-03
Allow running openshift-sdn with standalone kube-proxy

danwinship:
Spinoff of https://github.com/openshift/cluster-network-operator/pull/819; along with an upcoming CNO PR this will allow us to run openshift-sdn with a standalone kube-proxy for testing


danwinship:
/retitle Allow running openshift-sdn with standalone kube-proxy
/retest

PR #201, opened 2020-10-10
Dockerfile: add tcpdump for debugging

dcbw:
We'll not infrequently try to debug, and this is easier and
less error prone than trying to 'oc debug node' and run toolbox.

@stevekuznetsov @danwinship @juanluisvaladas

### DIFF
--- a/images/sdn/Dockerfile.rhel
+++ b/images/sdn/Dockerfile.rhel
@@ -15,7 +15,7 @@ RUN INSTALL_PKGS=" \
       libmnl libnetfilter_conntrack conntrack-tools \
       libnfnetlink iproute procps-ng openssl \
       iputils binutils xz util-linux dbus nftables \
-      " && \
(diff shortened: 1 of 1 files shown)


images/sdn/Dockerfile.rhel
```diff
@@ -15,7 +15,7 @@ RUN INSTALL_PKGS=" \
       libmnl libnetfilter_conntrack conntrack-tools \
       libnfnetlink iproute procps-ng openssl \
       iputils binutils xz util-linux dbus nftables \
-      " && \
+      tcpdump conntrack-tools" && \
```
danwinship, line 18: `conntrack-tools` is already listed above. (kube-proxy depends on it)

dcbw: Good catch, removed.

dcbw:
@danwinship PTAL thanks

PR #204, opened 2020-10-21
Bug 1878845: Fix ruleversion

juanluisvaladas:
In commit a464217b we changed the flows but forgot to increase the ruleVersion. This breaks upgrades in multitenant.

Also since we're here another little bug that isn't breaking anything right now. Initially 300 was the highest priority but now we have a rule that is 400, make it 1000 so that it's guaranteed that the flow is always evaluated.

### DIFF
--- a/pkg/network/node/ovscontroller.go
+++ b/pkg/network/node/ovscontroller.go
@@ -36,7 +36,7 @@ const (
 	Vxlan0 = "vxlan0"
 
 	// rule versioning; increment each time flow rules change
-	ruleVersion = 7
+	ruleVersion = 8
 
 	ruleVersionTable = 253
 )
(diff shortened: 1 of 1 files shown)


juanluisvaladas:
I still need to test the upgrade between 4.5 and 4.6. I'm currently waiting for the 4.5 cluster to deploy.

Makefile
```diff
@@ -9,7 +9,6 @@ GO_BUILD_PACKAGES = \
     ./cmd/... \
     ./vendor/github.com/containernetworking/plugins/plugins/ipam/host-local \
     ./vendor/k8s.io/kubernetes/cmd/kube-proxy
-GO_BUILD_PACKAGES_EXPANDED =$(shell GO111MODULE=on $(GO) list $(GO_MOD_FLAGS) $(GO_BUILD_PACKAGES))
```
rcarrillocruz, line 12: bogus change, probably from git commit -a

juanluisvaladas: Fixed :sweat_smile: yep, git commit -a

rcarrillocruz:
Overriding since othre jobs passed and we need to expedite this for 4.6

PR #209, opened 2020-10-28
Bug 1890130: fix pod creation deadlock

squeed:
Commit b5f89a66 introduced a deadlock on pod creation, but only for multitenant mode.

The fix is to lock runningPods as late as possible.

Fixes: 1890130

### DIFF
--- a/pkg/network/node/pod.go
+++ b/pkg/network/node/pod.go
@@ -255,9 +255,6 @@ func (m *podManager) processCNIRequests() {
 }
 
 func (m *podManager) processRequest(request *cniserver.PodRequest) *cniserver.PodResult {
-	m.runningPodsLock.Lock()
-	defer m.runningPodsLock.Unlock()
-
 	pk := getPodKey(request.PodNamespace, request.PodName)
 	result := &cniserver.PodResult{}
 	switch request.Command {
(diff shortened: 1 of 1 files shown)


squeed:
Single test failed; GCP issue:

` Failed to setup provider config for "gce": Error building GCE/GKE provider: unexpected response listing zones: googleapi: Error 503: The service is currently unavailable.`

/retest

squeed:
More importantly, e2e-aws-multitenant is green (and we don't appear to have regressed networkpolicy).

pkg/network/node/pod.go
```diff
@@ -288,6 +289,8 @@ func (m *podManager) processRequest(request *cniserver.PodRequest) *cniserver.Po
 		}
 		result.Err = err
 	case cniserver.CNI_DEL:
+		m.runningPodsLock.Lock()
+		defer m.runningPodsLock.Unlock()
```
danwinship, line 293: for consistency should we drop the lock before calling `teardown`?

squeed: definitely, good catch.

squeed: fixed.

squeed:
CI all green, just needs lgtm from someone.

PR #188, opened 2020-09-14
Bug 1812333: Fix error handling from DNS nameservers

alexanderConstantinescu:
Egress network policy should be able to resolve DNS names. If
multiple DNS servers are listed in /etc/resolv.conf and one of these
returns an NXDOMAIN response, we, in turn, return an error and
skip testing any succeeding nameserver.

Instead of returning an error, just continue to the next one.

This also modifies the log output from the verbose:

```
E0914 09:07:04.465928  834879 dns.go:146] failed to get a valid answer: ;; opcode: QUERY, status: NXDOMAIN, id: 8469
;; flags: qr rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 0

;; QUESTION SECTION:
;pod2.ec2.internal.	IN	 A

;; AUTHORITY SECTION:
.	9454	IN	SOA	a.root-servers.net. nstld.verisign-grs.com. 2020030401 1800 900 604800 86400
E0304 10:00:10.168990 1004855 egress_dns.go:64] failed to get a valid answer: ;; opcode: QUERY, status: NXDOMAIN, id: 7927
;; flags: qr rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 0

;; QUESTION SECTION:
;pod2.ec2.internal.	IN	 A

;; AUTHORITY SECTION:
.	9454	IN	SOA	a.root-servers.net. nstld.verisign-grs.com. 2020030401 1800 900 604800 86400
```

to a more short and concrete output:

```
W0914 09:07:04.465928  834879 dns.go:146] failed to get a valid answer: 3 from nameserver: 10.11.5.19 for domain: pod2.ec2.internal
```
 

Signed-off-by: Alexander Constantinescu <aconstan@redhat.com>

PR #210, opened 2020-10-29
Bug 1891454: Fix error handling from DNS nameservers

alexanderConstantinescu:
This was done on master with PR #188. However I closed that, since PR #164 managed to slip in and fix the problem before 188. 164 is however a bit to big and not intended to be back-ported. So let's do this instead. 

/assign @danwinship

### DIFF
--- a/pkg/network/common/dns.go
+++ b/pkg/network/common/dns.go
@@ -10,6 +10,7 @@ import (
 
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/klog"
 )
 
 const (
@@ -139,10 +140,12 @@ func (d *DNS) getIPsAndMinTTL(domain string) ([]net.IP, time.Duration, error) {
 		c.Timeout = 5 * time.Second
 		in, _, err := c.Exchange(msg, dialServer)
 		if err != nil {
-			return nil, defaultTTL, err
+			klog.Warningf("failed to query nameserver: %s with address: %s for domain: %s, err: %v", server, dialServer, domain, err)
+			continue
 		}
 		if in != nil && in.Rcode != dns.RcodeSuccess {
-			return nil, defaultTTL, fmt.Errorf("failed to get a valid answer: %v", in)
+			klog.Warningf("failed to get a valid answer: %v from nameserver: %s for domain: %s", in.Rcode, server, domain)
+			continue
 		}
 
 		if in != nil && len(in.Answer) > 0 {


pkg/network/common/dns.go
```diff
@@ -142,7 +143,8 @@ func (d *DNS) getIPsAndMinTTL(domain string) ([]net.IP, time.Duration, error) {
 			return nil, defaultTTL, err
 		}
 		if in != nil && in.Rcode != dns.RcodeSuccess {
-			return nil, defaultTTL, fmt.Errorf("failed to get a valid answer: %v", in)
+			klog.Warningf("failed to get a valid answer: %v from nameserver: %s for domain: %s", in.Rcode, server, domain)
+			continue
```
danwinship, line 147: This is not the line that's returning an error in the bugzilla bug:

> `2020-10-15T08:20:43.03350351Z E1015 08:20:43.033470    3066 egress_dns.go:64] read udp 10.37.64.130:54962->10.37.64.109:53: read: connection refused`

There's no "`failed to get a valid answer`" in the logged error message.

In fact, if a DNS server actually returns `NXDOMAIN` then we're done, because that means it's conclusively saying that the name doesn't exist; if it means "you should ask another DNS server", it would return `NOERROR` but an empty response (which the code already deals with).

It looks like the failure reported in the bug must be from the previous `return` (which is the only one in `getIPsAndMinTTL` that doesn't prefix the returned error message with any other string). I'm not sure if we want to retry on _all_ errors from `c.Exchange` or just some. You'll have to look at the code/docs.

I think the code in master is actually wrong... `queryServer`/`getIPsAndMinTTL` ought to be correctly distinguishing "I couldn't talk to the server" from "I talked to the server and it says there's no match".

danwinship (changes requested):

pkg/network/common/dns.go
alexanderConstantinescu: So, there were multiple parts to this bug which can cause confusion. Essentially, the issue is related to bug; https://bugzilla.redhat.com/show_bug.cgi?id=1812333 (which I closed in favor of https://bugzilla.redhat.com/show_bug.cgi?id=1889943, since this one had a customer case attached). 

Problem with: https://bugzilla.redhat.com/show_bug.cgi?id=1812333

Whenever you specify a custom nameserver in `/etc/resolv.conf` and try to lookup a domain that only that nameserver knows about, all others will return an `NXDOMAIN` when you query them. If that happens we stop checking the preceding nameservers (read: our custom nameserver) and don't end up resolving the DNS name, even though the custom nameserver is aware of it. 

Problem with: https://bugzilla.redhat.com/show_bug.cgi?id=1889943 

For this bug what happened was that the customer had specified two custom nameservers, one which was not working (resulting in `read: connection refused` as you noted, and was being queried first) and another which was working (but which we never ended up querying because of this problem of stopping the iteration as soon as _any_ nameserver does not return an `NoError`) 

>In fact, if a DNS server actually returns NXDOMAIN then we're done, because that means it's conclusively saying that the name doesn't exist; if the DNS server wants to say "you should ask another DNS server", it would return NOERROR but an empty response (which the code already deals with).

Doesn't that apply to "normal" public nameservers? If a company uses a corporate and custom nameserver, other nameservers might not be aware of a domain that only that nameserver knows about, and thus - correctly - return an `NXDOMAIN`...ultimately causing us to never query that corporate nameserver. This is the problem here. 

>  I'm not sure if we want to retry on all errors from c.Exchange or just some.

I don't see what the problem is of retrying and continuing on to the next nameserver in the list for all errors? I don't expect a lot of nameservers to be listed, which might have resulted in performance issues. 

> I think the code in master is actually wrong...

Master works for what concerns the issue I am trying to fix, see: https://bugzilla.redhat.com/show_bug.cgi?id=1812333#c31 the reason is https://github.com/openshift/sdn/blob/master/pkg/network/common/dns.go#L226 where even if one `queryServer` fails we still proceed to check the next and thus end up resolving the domain.

alexanderConstantinescu: @danwinship , did you get time to have a look at my comment above?

danwinship: sorry. OK, but, so, if you want to fix the "read: connection refused" bug, you need to keep looping rather than returning when `c.Exchange` returns an error, right?

alexanderConstantinescu: Hmm, yes, you're right. I will update the PR now.

alexanderConstantinescu:
@danwinship : could you please take another look at this?

PR #211, opened 2020-10-29
Bug 1891455: Fix error handling from DNS nameservers

alexanderConstantinescu:
4.5 back-port of #210 

/assign @rcarrillocruz


alexanderConstantinescu:
/bugzilla refresh

@danwinship / @rcarrillocruz : could any of you lgtm and approve this?

PR #212, opened 2020-10-29
Bug 1891457: Fix error handling from DNS nameservers

alexanderConstantinescu:
4.4 back-port of #211 

### DIFF
--- a/pkg/network/common/dns.go
+++ b/pkg/network/common/dns.go
@@ -10,6 +10,7 @@ import (
 
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/klog"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


pkg/network/common/dns.go
```diff
@@ -142,7 +143,8 @@ func (d *DNS) getIPsAndMinTTL(domain string) ([]net.IP, time.Duration, error) {
 			return nil, defaultTTL, err
 		}
 		if in != nil && in.Rcode != dns.RcodeSuccess {
-			return nil, defaultTTL, fmt.Errorf("failed to get a valid answer: %v", in)
+			klog.Warningf("failed to get a valid answer: %v from nameserver: %s for domain: %s", in.Rcode, server, domain)
+			continue
```
JacobTanenbaum, line 147: Should we return an error if none of the nameservers get a valid answer?

alexanderConstantinescu: That's what will happen on L166, `len(ips)` will be 0, since we never add any in such a case

bparees:
@alexanderConstantinescu this is marked as a medium severity BZ and 4.4 is in maintenance support only (and is about to go EOL entirely).  Is there a reason this needs to go into 4.4.z at this point?  I don't currently intend to approve it without more info.

alexanderConstantinescu:
> Is there a reason this needs to go into 4.4.z at this point? 

No, _at this point_: no. I forgot to close it.

PR #214, opened 2020-11-02
Bug 1892376: Ignore if netns is already deleted while deleting ns

tssurya:
While deleting a namespace, we check and delete the netnamespace
and remove the corresponding netid from the masterVNIDMap cache.
However if the netns is deleted before the ns, we return an error
and do not remove it from the cache. Upon recreation of the same
ns/project in the future, a new netns does not get created because
according to the masterVNIDMap cache, that netns is already present.

In this patch we ignore, emit a warning and continue to remove the
netid from the masterVNIDMap cache if the netns is already deleted.
There is no point in returning an error if the objective in the end
is to have the netns deleted. The order of deletion between netns
and ns should not matter.

### DIFF
--- a/pkg/network/master/vnids.go
+++ b/pkg/network/master/vnids.go
@@ -8,6 +8,7 @@ import (
 	"k8s.io/klog/v2"
 
 	kapi "k8s.io/api/core/v1"
+	kapierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


tssurya:
I don't know where/if I should add unit tests, there doesn't seem to be a mock for the networkClient portion of things, so the function I touched doesn't have tests.

squeed:
These sorts of things are notoriously hard to mock out; that's OK.

/lgtm
/approve

tssurya:
thanks casey!

/cherry-pick release-4.6

PR #220, opened 2020-11-12
Bug 1897073: Don't throw an error for control plane VNID

juanluisvaladas:
Initially VNIDs 1-9 were forbidden, in OpenShift 4 we need the control
plane components to communicate with each other, so we assigned netid 1 to
it. This is logging an error incorrectly.

### DIFF
--- a/pkg/network/master/vnids.go
+++ b/pkg/network/master/vnids.go
@@ -80,7 +80,7 @@ func (vmap *masterVNIDMap) isAdminNamespace(nsName string) bool {
 
 func (vmap *masterVNIDMap) markAllocatedNetID(netid uint32) error {
 	// Skip GlobalVNID, not part of netID allocation range
-	if netid == network.GlobalVNID {
+	if netid < network.MinVNID {
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


juanluisvaladas:
/retitle  Bug 1897073: Don't throw an error for control plane VNID

pkg/network/master/vnids.go
```diff
@@ -80,7 +80,7 @@ func (vmap *masterVNIDMap) isAdminNamespace(nsName string) bool {
 
 func (vmap *masterVNIDMap) markAllocatedNetID(netid uint32) error {
 	// Skip GlobalVNID, not part of netID allocation range
-	if netid == network.GlobalVNID {
+	if netid == network.GlobalVNID || netid == network.ControlPlaneVNID {
```
danwinship, line 83: hm... we were previously using VNID 2 for something as well, and might want to do that again later... I'd say just change this to `if netid < network.MinVNID {` (and then don't define `ControlPlaneVNID`)

juanluisvaladas: Fair enough, changed

PR #222, opened 2020-11-17
Bug 1893067: Flag SDN setup dirty/clean for AlreadySetup to read

alexanderConstantinescu:
There are situations applicable (such as for this bug) where the SDN
can start and have a look in OVS for the existing flows and find a partial
setup for whatever reason.

Since we were only writing one type of `ruleVersion` flow indicating the setup
there can be incorrect comparisons between several restarts of the SDN. One
such example is given here:

-User creates cluster
-OVS and SDN start, SDN starts programming OVS
-SDN finishes programming OVS and writes the note flow
-Upgrade starts
-New SDN and new OVS start, new SDN starts programming new OVS
-New OVS crashes in the middle of the setup, bringing new SDN down with it. The old note flow is still there
-New+1 SDN starts and checks new+1 OVS, finds the note flow and thinks "all is good"

Instead, start by checking the last setup in `AlreadySetUp`. If last is dirty or
has a missing flow, write dirty and proceed with setup. If setup manages to
finish we write clean, so that at the next restart we know.

An example of this situation can be found here: https://bugzilla.redhat.com/show_bug.cgi?id=1893067#c15 

/cc @tssurya

### DIFF
--- a/pkg/network/node/ovscontroller.go
+++ b/pkg/network/node/ovscontroller.go
@@ -39,6 +39,7 @@ const (
 	ruleVersion = 8
 
 	ruleVersionTable = 253
+	podFlowsTable    = 40
 )
 
 func NewOVSController(ovsif ovs.Interface, pluginId int, useConnTrack bool, localIP string) *ovsController {
@@ -53,18 +54,25 @@ func (oc *ovsController) getVersionNote() string {
 }
 
 func (oc *ovsController) AlreadySetUp(vxlanPort uint32) bool {
-	flows, err := oc.ovs.DumpFlows("table=%d", ruleVersionTable)
-	if err != nil || len(flows) != 1 {
+	ruleVersionFlows, err := oc.ovs.DumpFlows("table=%d", ruleVersionTable)
+	if err != nil || len(ruleVersionFlows) != 1 {
 		return false
 	}
-
+	// Have a look in the table filled in by setupPodFlows, as to make
+	// sure that we don't have a partial setup remaining in other tables,
+	// but are actually missing pod setup from an earlier run and don't
+	// proceed to setting up any pods that we should be doing in this run.
+	podFlows, err := oc.ovs.DumpFlows("table=%d", podFlowsTable)
+	if err != nil || len(podFlows) == 0 {
+		return false
+	}
 	port, err := oc.ovs.Get("Interface", Vxlan0, "options:dst_port")
 	// the call to ovs.Get() returns the port number surrounded by double quotes
 	// so add them to the structs value for purposes of comparison
 	if err != nil || fmt.Sprintf("\"%d\"", vxlanPort) != port {
 		return false
 	}
-	if parsed, err := ovs.ParseFlow(ovs.ParseForDump, flows[0]); err == nil {
+	if parsed, err := ovs.ParseFlow(ovs.ParseForDump, ruleVersionFlows[0]); err == nil {
 		return parsed.NoteHasPrefix(oc.getVersionNote())
 	}
 	return false
--- a/pkg/network/node/ovscontroller_test.go
+++ b/pkg/network/node/ovscontroller_test.go
@@ -772,27 +772,43 @@ func TestOVSEgressNetworkPolicy(t *testing.T) {
 
 func TestAlreadySetUp(t *testing.T) {
 	testcases := []struct {
-		flow    string
+		flows   []string
 		success bool
 	}{
 		{
-			// Good note
-			flow:    fmt.Sprintf("cookie=0x0, duration=4.796s, table=253, n_packets=0, n_bytes=0, actions=note:00.%02x.00.00.00.00", ruleVersion),
+			// Good note and existing pod flows
+			flows: []string{
+				fmt.Sprintf("cookie=0x0, duration=4.796s, table=253, n_packets=0, n_bytes=0, actions=note:00.%02x.00.00.00.00", ruleVersion),
+				fmt.Sprintf("cookie=0x0, duration=11753.619s, table=40, n_packets=324, n_bytes=13608, priority=100,arp,arp_tpa=10.128.2.76 actions=output:10"),
+			},
 			success: true,
 		},
 		{
+			// Good note but missing pod flows
+			flows: []string{
+				fmt.Sprintf("cookie=0x0, duration=4.796s, table=253, n_packets=0, n_bytes=0, actions=note:00.%02x.00.00.00.00", ruleVersion),
+			},
+			success: false,
+		},
+		{
 			// Wrong version
-			flow:    fmt.Sprintf("cookie=0x0, duration=4.796s, table=253, n_packets=0, n_bytes=0, actions=note:00.%02x.00.00.00.00", ruleVersion-1),
+			flows: []string{
+				fmt.Sprintf("cookie=0x0, duration=4.796s, table=253, n_packets=0, n_bytes=0, actions=note:00.%02x.00.00.00.00", ruleVersion-1),
+			},
 			success: false,
 		},
 		{
 			// Wrong table
-			flow:    fmt.Sprintf("cookie=0x0, duration=4.796s, table=10, n_packets=0, n_bytes=0, actions=note:00.%02x.00.00.00.00", ruleVersion),
+			flows: []string{
+				fmt.Sprintf("cookie=0x0, duration=4.796s, table=10, n_packets=0, n_bytes=0, actions=note:00.%02x.00.00.00.00", ruleVersion),
+			},
 			success: false,
 		},
 		{
 			// No note
-			flow:    "cookie=0x0, duration=4.796s, table=253, n_packets=0, n_bytes=0, actions=goto_table:50",
+			flows: []string{
+				"cookie=0x0, duration=4.796s, table=253, n_packets=0, n_bytes=0, actions=goto_table:50",
+			},
 			success: false,
 		},
 	}
@@ -810,7 +826,9 @@ func TestAlreadySetUp(t *testing.T) {
 		}
 
 		otx := ovsif.NewTransaction()
-		otx.AddFlow(tc.flow)
+		for _, flow := range tc.flows {
+			otx.AddFlow(flow)
+		}
 		if err := otx.Commit(); err != nil {
 			t.Fatalf("(%d) unexpected error from AddFlow: %v", i, err)
 		}


pkg/network/node/ovscontroller.go
```diff
@@ -53,18 +54,25 @@ func (oc *ovsController) getVersionNote() string {
...
+	ruleVersionFlows, err := oc.ovs.DumpFlows("table=%d", ruleVersionTable)
+	if err != nil || len(ruleVersionFlows) != 1 {
+		return false
+	}
+	// Have a look in the table filled in by setupPodFlows, as to make
+	// sure that we don't have a partial setup remaining in other tables,
+	// but are actually missing pod setup from an earlier run and don't
+	// proceed to setting up any pods that we should be doing in this run.
+	podFlows, err := oc.ovs.DumpFlows("table=%d", podFlowsTable)
+	if err != nil || len(podFlows) == 0 {
```
tssurya, line 66: considering the many pods each worker will have on openshift, I guess this condition to check for at least one podFlow is ok, as in there will anyways never be a condition where we don't have at least one non-hostnetwork pod on the node.

tssurya (commented):
thanks for doing this Alex, looks ok to me, let's wait for CI.

pkg/network/node/ovscontroller.go
```diff
@@ -39,6 +39,7 @@ const (
 	ruleVersion = 8
 
 	ruleVersionTable = 253
+	podFlowsTable    = 40
```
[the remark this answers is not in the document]
alexanderConstantinescu: It's not any more or any less special than 20 / 70, `setupPodFlows` touches all three, see:https://github.com/openshift/sdn/blob/9ec55d64eb273325b39d7cae50cd90e8d651b915/pkg/network/node/ovscontroller.go#L314-L335 We need to only look in one here though, and I chose 40 (for no other reason than :four_leaf_clover: ).

pkg/network/node/ovscontroller.go
alexanderConstantinescu: > as in there will anyways never be a condition where we don't have at least one non-hostnetwork pod on the node.

Well, there will be when we start fresh, but in that case we want to set everything up anyways. I guess I could have re-worked the code to check that each individual pod that is returned in `GetPodNetworkInfo` is passed to `alreadySetup` and check each one...but I figured that reworks the code too much and might lead to more entropy and more cases of "full setup".

pkg/network/node/ovscontroller.go
```diff
@@ -53,18 +54,25 @@ func (oc *ovsController) getVersionNote() string {
 }
 
 func (oc *ovsController) AlreadySetUp(vxlanPort uint32) bool {
-	flows, err := oc.ovs.DumpFlows("table=%d", ruleVersionTable)
-	if err != nil || len(flows) != 1 {
+	ruleVersionFlows, err := oc.ovs.DumpFlows("table=%d", ruleVersionTable)
+	if err != nil || len(ruleVersionFlows) != 1 {
+		return false
+	}
+	// Have a look in the table filled in by setupPodFlows, as to make
```
danwinship, line 61: So this is a pre-existing weirdness, but the function calls `DumpFlows("table=%d", ruleVersionTable)` at the top, and then parses it at the bottom, after doing other checks in between for some reason? The two halves of the rule table check should be put back together.

pkg/network/node/ovscontroller.go
danwinship: If there are actually no (pod-network) pods then there's not really much downside to destroying and recreating the bridge anyway

danwinship (commented):
I don't really like this... A node having flows that are correct except that the pod flows are missing is a "can't happen" situation. So I wonder if this PR would really have fixed the node in the bz, or if it would only be fixing one symptom of some larger problem, but still leaving the node broken. :frowning_face:

tssurya:
> I don't really like this... A node having flows that are correct except that the pod flows are missing is a "can't happen" situation. So I wonder if this PR would really have fixed the node in the bz, or if it would only be fixing one symptom of some larger problem, but still leaving the node broken.

hmm really ? Can't there be situations where for some reason the existing pods on the node were not picked up and no podflows got created? But yea the situation is weird. However the way we do the check now, where we check only for the 253 table, is it always guaranteed that if that exists all is well in the setup? I mean is that an accurate metric for determining things are good ?

alexanderConstantinescu:
> A node having flows that are correct except that the pod flows are missing is a "can't happen" situation. 

Except it just happened, this bug is the proof. 

> So I wonder if this PR would really have fixed the node in the bz, or if it would only be fixing one symptom of some larger problem, but still leaving the node broken.

I agree that figuring out what caused the pod flows to disappear and fixing that would have been great. But now that we've seen that this _can_ happen I think that it's worthwhile guarding against it. In the ideal case we would have figured out the disappearing pod flows _and_ fixed this, but at least we can do one. 

> Can't there be situations where for some reason the existing pods on the node were not picked up in time and no podflows got created?

A hypothesis I had was that OVS crashed while the SDN was setting up networking, leading to a restart of the SDN and stuff in a half-done state. But we won't know for sure.

danwinship:
> > A node having flows that are correct except that the pod flows are missing is a "can't happen" situation.
> 
> Except it just happened, this bug is the proof.

My point was that if we have no good explanation for how the table=40 flows disappeared, then it's hard to be confident that it wasn't just one symptom of a larger bug which would still be causing problems for the node even after we fixed the table=40 flows. (It's hard to even be confident that the table=40 flows will _stay_ fixed after we fix them...)

(I'm not vetoing the PR, but I really wish we had a better explanation.)

> A hypothesis I had was that OVS crashed while the SDN was setting up networking, leading to a restart of the SDN and stuff in a half-done state. But we won't know for sure.

A good hypothesis, because that scenario used to be possible. But #52 should have fixed it; now adding the version note is the last step of initializing the bridge, so if the bridge was "in a half-done state", that would mean it wouldn't have the version note, and so we'd decide to recreate the bridge based on that, without needing to look at table 40.

alexanderConstantinescu:
> But #52 should have fixed it; now adding the version note is the last step of initializing the bridge, so if the bridge was "in a half-done state", that would mean it wouldn't have the version note, and so we'd decide to recreate the bridge based on that, without needing to look at table 40.

Hmm, `ruleVersion` is a `const`, so the comment in the code here below is not really correct. Were we basing ourselves on this to be true? 

https://github.com/openshift/sdn/blob/60437b0ab6bc01e04d90bc11f4d5a09c484472d4/pkg/network/node/ovscontroller.go#L38-L39

Moreover, the check here is useless it seems:

https://github.com/openshift/sdn/blob/60437b0ab6bc01e04d90bc11f4d5a09c484472d4/pkg/network/node/ovscontroller.go#L50-L52

I mean what happens if the following occurs:

1. User creates cluster
2. OVS and SDN start, SDN starts programming OVS 
3. SDN finishes programming OVS and writes the note flow
4. Upgrade starts 
5. New SDN and new OVS start, new SDN starts programming new OVS
6. New OVS crashes in the middle of the setup, bringing new SDN down with it. The old note flow is still there
7. New+1 SDN starts and checks new+1 OVS, finds the note flow and thinks "all is good"

Naturally this scenario is based on a lot of non-empirical hearsay, and in OCP < 4.6 we're not running OVS on the system which (maybe?) blows all the flows away on any OVS container restart (well maybe not, if it reads its DB from a mounted volume). MCO will restart the node at one point and definitely blow them away at one moment anyways. 

My point is however, `ruleVersion` is never incremented, which I think it should if we're trying to check state w.r.t a given SDN / OVS run. There might be corner cases where not incrementing `ruleVersion` leads to "half-done" states.

alexanderConstantinescu:
Expanding on my previous comment: we don't really need to increment the value, we can just use a boolean int to mark the flows "dirty" (last setup not completed) or "clean" (last setup completed). 

So, something along the lines of:

1. SDN starts, checks OVS note exists and equals clean
2. It does not exist, it writes dirty and proceeds with setup 
3. Finishes setup and writes clean
4. Upgrade 
5. SDN upgrades, checks OVS note exists and equals clean 
5. It does not exist (because the flows have been flushed somehow), it writes dirty and proceeds with setup 
6. OVS crashes, kills SDN 
7. SDN restarts, , checks OVS note exists and equals clean
8. It exists, but is dirty and proceeds with setup
9. Finishes setup and writes clean

That should do it, I believe...

alexanderConstantinescu:
/retitle Bug 1893067: Flag SDN setup dirty/clean for AlreadySetup to read

danwinship:
> My point is however, `ruleVersion` is never incremented

That's not what's being "version"ed; `ruleVersion` tracks, vaguely, the version of openshift-sdn that created the rules. The use of `ruleVersion` as "proof that we reached the end of setup" is a later addition to the system, and doesn't depend on the actual _content_ of the version note at all, just whether it's there.

(Basically, `ruleVersion` has to be increased any time we make a change to `ovscontroller.go` such that a version of openshift-sdn with the new version would not be able to run successfully against a set of OVS flows created by a version of openshift-sdn with the old version. eg, see #204.)

> 5. New SDN and new OVS start, new SDN starts programming new OVS
> 6. New OVS crashes in the middle of the setup, bringing new SDN down with it. The old note flow is still there
> 7. New+1 SDN starts and checks new+1 OVS, finds the note flow and thinks "all is good"

No. There are two different cases here, but both of them should work:

5a. New SDN and new OVS start, new SDN is happy with the existing flows in OVS and so declares "startup complete" and proceeds to its main loop, preserving all the existing flows
6a. New OVS crashes in the middle of the setup, bringing new SDN down with it.
7a. New+1 SDN starts and checks new+1 OVS, finds the note flow, thinks "all is good", _and it's right_: all _is_ good.

5b. New SDN and new OVS start. New SDN is unhappy with the flows in OVS, so it destroys `br0`, recreates it, and then starts programming new OVS
6b. New OVS crashes in the middle of the setup, bringing new SDN down with it; there is no note flow because the old one was destroyed with the old bridge and the new one hasn't been written yet
7b. New+1 SDN starts and checks new+1 OVS, finds no note flow, so it destroys `br0` and starts over again.

alexanderConstantinescu:
> (Basically, `ruleVersion` has to be increased any time we make a change to `ovscontroller.go` such that a version of openshift-sdn with the new version would not be able to run successfully against a set of OVS flows created by a version of openshift-sdn with the old version. eg, see #204.)

Oh! I'm mixing the concepts up in that case, sorry. 
 
> No. There are two different cases here, but both of them should work:
> 
> 5a. New SDN and new OVS start, new SDN is happy with the existing flows in OVS and so declares "startup complete" and proceeds to its main loop, preserving all the existing flows
> 6a. New OVS crashes, bringing new SDN down with it.
> 7a. New+1 SDN starts and checks new+1 OVS, finds the note flow, thinks "all is good", _and it's right_: all _is_ good.
> 
> 5b. New SDN and new OVS start. New SDN is unhappy with the flows in OVS, so it destroys `br0`, recreates it, and then starts programming new OVS
> 6b. New OVS crashes in the middle of the setup, bringing new SDN down with it; there is no note flow because the old one was destroyed with the old bridge and the new one hasn't been written yet
> 7b. New+1 SDN starts and checks new+1 OVS, finds no note flow, so it destroys `br0` and starts over again.

Isn't that the problem here though? That it determining if it's "happy with the flows" depends on something which doesn't really indicate that flows have been setup properly during run N-1, but instead that it happened sometime in run N-X. 

I'm mean, I think there's a third case here:

5c. New SDN and new OVS start, new SDN is happy with the existing flows in OVS because it saw the note - but that note was created by the SDN two runs ago, the last one crashed however and didn't manage to write them all (maybe also managed to flush some tables before getting to the re-creation part) - and so declares "startup complete" and proceeds to its main loop, preserving all the existing flows
6c. New OVS crashes, bringing new SDN down with it.
7c. New+1 SDN starts and checks new+1 OVS, finds the note flow, thinks "all is good", and it's wrong: all _is not_ good.

I think we need a dedicated indicator for the run N-1, right?

alexanderConstantinescu:
I've reverted back to the first version with the table 40 check, as to not touch `ruleVersion`

alexanderConstantinescu:
We don't have enough data to determine the root cause and should thus not implement a fix to a problem we don't completely understand. Moreover: the customer case has been closed out. I have thus closed the BZ and will thus close this. 

/close

PR #226, opened 2020-12-01
Bug 1896958: NetworkPolicy performance (pod caching)

danwinship:
Trying to improve NetworkPolicy performance/memory usage in large clusters with lots of policies and lots of changes...

/cc @squeed @juanluisvaladas @JacobTanenbaum


pkg/network/node/networkpolicy.go
```diff
@@ -657,32 +684,24 @@ func (np *networkPolicyPlugin) handleAddOrUpdatePod(obj, _ interface{}, eventTyp
 	// We don't want to grab np.Lock for every Pod.Status change...
```
squeed, line 684: can probably remove this commment

pkg/network/node/networkpolicy.go
```diff
@@ -657,32 +684,24 @@ func (np *networkPolicyPlugin) handleAddOrUpdatePod(obj, _ interface{}, eventTyp
...
-	if !podExisted {
-		return
-	}
-
 	np.lock.Lock()
 	defer np.lock.Unlock()
 
-	delete(np.pods, pod.UID)
-	np.refreshNetworkPolicies(refreshForPods)
+	np.refreshPodNetworkPolicies(pod)
```
squeed, line 704: Does this need to be synchronous?

pkg/network/node/networkpolicy.go
```diff
@@ -252,21 +252,32 @@ func (np *networkPolicyPlugin) syncNamespace(npns *npNamespace) {
...
 			npns.dirty = false
 		}
 	}
+	if err := otx.Commit(); err != nil {
+		utilruntime.HandleError(fmt.Errorf("Error syncing OVS flows: %v", err))
+	}
 }
 
-func (np *networkPolicyPlugin) syncNamespaceFlows(npns *npNamespace) {
+func (np *networkPolicyPlugin) syncNamespaceFlows(otx ovs.Transaction, npns *npNamespace) {
```
juanluisvaladas, line 291: This function doesn't actually sync anymore, probably should be renamed to generateNamespaceFlows or something like that instead of syncNamespaceFlows

juanluisvaladas (approved):
Looks really good, I have a few nitpicks but nothing really relevant  given the urgency.

pkg/network/node/networkpolicy.go
danwinship: do you have any more specific concern than that?

Note that this is necessarily completely asynchronous with respect to CNI pod creation/deletion anyway

pkg/network/node/networkpolicy.go
```diff
@@ -401,13 +398,17 @@ func (np *networkPolicyPlugin) selectPodsFromNamespaces(nsLabelSel, podLabelSel
...
-	namespaces := np.selectNamespacesInternal(nsSel)
-	for _, pod := range np.pods {
-		vnid, exists := namespaces[pod.Namespace]
-		if exists && podSel.Matches(labels.Set(pod.Labels)) {
+	nsLister := np.node.kubeInformers.Core().V1().Pods().Lister()
+	for namespace, vnid := range np.selectNamespacesInternal(nsSel) {
+		pods, err := nsLister.Pods(namespace).List(podSel)
+		if err != nil {
+			// Shouldn't happen
+			utilruntime.HandleError(fmt.Errorf("Could not find matching pods in namespace %q: %v", namespace, err))
```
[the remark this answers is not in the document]
danwinship: It can only happen if there's a bug in the code. (Like, on the level of "the informer contains the wrong kind of object".) Users shouldn't ever see this message.

pkg/network/node/networkpolicy.go
squeed: Right, but it blocks the pod informer from processing deltas.

squeed: oh, I see, refreshPodNetworkPolicies is already asynchronous - didn't quite grok it.

danwinship:
CI is a dumpster fire but I brought up a cluster with this PR and ran `openshift-tests` by hand. All the NetworkPolicy tests passed.

pkg/network/node/networkpolicy.go
```diff
@@ -65,7 +68,8 @@ type npNamespace struct {
 type npPolicy struct {
 	policy            networkingv1.NetworkPolicy
 	watchesNamespaces bool
-	watchesPods       bool
+	watchesAllPods    bool
```
squeed, line 71: Want to leave a docblock indicating what the `watchesFoo` variables mean?

pkg/network/node/networkpolicy.go
```diff
@@ -736,18 +758,28 @@ func (np *networkPolicyPlugin) handleDeleteNamespace(obj interface{}) {
...
+	podNs := np.namespacesByName[pod.Namespace]
+	for _, npns := range np.namespaces {
+		for _, npp := range npns.policies {
+			if (npp.watchesOwnPods && npns == podNs) || npp.watchesAllPods {
+				npns.mustRecalculate = true
 			}
 		}
-		if changed && npns.inUse {
+		if npns.mustRecalculate && npns.inUse {
 			np.syncNamespace(npns)
```
squeed, line 783: Does it make sense to just set `npns.mustSync` on all relevant namespaces, and trigger the runner once? I don't see much advantage to "splitting out" the transaction.

PR #219, opened 2020-11-11
Bug 1879607: bump-k8s-to-1.19.2

tssurya:
This PR rebases sdn on k8sv1.19.2.
Also includes the upstream PR https://github.com/openshift/kubernetes/pull/442/commits/6e071b22148cb267190af53d9bd2520fad282602

Part of [SDN-1346](https://issues.redhat.com/browse/SDN-1346)

PR #227, opened 2020-12-04
update to kube 1.20

danwinship:
Update to `s d n -4.7-kubernetes-1.20-rc.0`. Obsoletes #219 (which it's partially based on)

/cc @tssurya


danwinship:
So the problem here (and probably in #219 too) is that it's not creating any iptables service rules...

pkg/openshift-sdn/proxy.go
```diff
@@ -43,7 +45,16 @@ func readProxyConfig(filename string) (*kubeproxyconfig.KubeProxyConfiguration,
 	if err := o.Complete(); err != nil {
 		return nil, err
 	}
-	return o.GetConfig(), nil
+
+	config := o.GetConfig()
```
abhat, line 49: Is there a reason we can't disable the endpointslice featuregates in the config file itself, so we don't need the changes you added to `readProxyConfig`?

Otherwise LGTM.

danwinship: "because CI is on fire and I didn't want to try to get a CNO PR in too". But I'll fix that later...

PR #230, opened 2020-12-07
Bug 1906844: Handle unsupported EndpointSlice and EndpointSliceProxying feature gates

abhat:
This PR reverts commit 12b21f8339b04ebe429cdab71674cad13b70fdef.

It also validates that unsupported feature gates are not accidentally enabled by the daemonset/deployment that launches 
openshift-sdn.

This is gated by the CNO PR: openshift/cluster-network-operator#905 landing first.

### DIFF
--- a/pkg/openshift-sdn/proxy.go
+++ b/pkg/openshift-sdn/proxy.go
@@ -13,12 +13,10 @@ import (
 	utilwait "k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apiserver/pkg/server/mux"
 	"k8s.io/apiserver/pkg/server/routes"
-	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	"k8s.io/client-go/kubernetes/scheme"
 	kv1core "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/record"
 	kubeproxyoptions "k8s.io/kubernetes/cmd/kube-proxy/app"
-	"k8s.io/kubernetes/pkg/features"
 	proxy "k8s.io/kubernetes/pkg/proxy"
 	kubeproxyconfig "k8s.io/kubernetes/pkg/proxy/apis/config"
 	pconfig "k8s.io/kubernetes/pkg/proxy/config"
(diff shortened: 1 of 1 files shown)


danwinship:
/lgtm
/hold
though I kind of feel like maybe it should check the gate and error out if you somehow managed to start it with the gate enabled

abhat:
do we really think that's possible given that we have validation now to make sure that feature-gates can't be over-ridden by user-config? or are we just trying to be defensive about it?

danwinship:
Just being defensive. _In theory_ someone could launch openshift-sdn via a non-CNO-managed DaemonSet

pkg/openshift-sdn/proxy.go
```diff
@@ -45,16 +44,13 @@ func readProxyConfig(filename string) (*kubeproxyconfig.KubeProxyConfiguration,
...
-	config.FeatureGates[string(features.EndpointSlice)] = false
-	config.FeatureGates[string(features.EndpointSliceProxying)] = false
-	utilfeature.DefaultMutableFeatureGate.SetFromMap(config.FeatureGates)
-
-	return config, nil
+	if config.FeatureGates[string(features.EndpointSlice)] ||
+		config.FeatureGates[string(features.EndpointSliceProxying)] {
+		return nil, fmt.Errorf("Unsupported feature gate(s) enabled in proxy config. Feature gates: %v",
+			config.FeatureGates)
+	}
```
danwinship, line 52: hm... I was thinking more that `runProxy` would check `if utilfeature.DefaultFeatureGate.Enabled(features.EndpointSliceProxying)` and bail out there.

(That will also protect us by making things fail immediately when the feature gate becomes on-by-default.)

abhat: the `readProxyConfig` is happening in `ValidateAndParse()` before `Start()` though. Don't we want to fail fast?

danwinship: It's still at startup time. The point is that we want to fail "if EndpointSliceProxing is enabled", not "if EndpointSliceProxying was enabled via command-line options"

abhat: Are there other paths besides command line args by which the gate could be enabled?

abhat: I guess I see your point. But I was hoping we will actually add support for EndpointSlices in o-sdn by the time the FeatureGate is turned on-by-default.

PR #233, opened 2020-12-14
Bug 1905761: Fix IP list for empty Egress network policy

vpickard:
For a policy with only Egress Type, the selectedIPs was being
set to an empty IP address. This resulted in attempting to program
a rule in OVS with an empty IP, resulting in an error programming
the flow(s)

Signed-off-by: vpickard <vpickard@redhat.com>

### DIFF
--- a/pkg/network/node/networkpolicy.go
+++ b/pkg/network/node/networkpolicy.go
@@ -71,8 +71,9 @@ type npPolicy struct {
 	watchesAllPods    bool
 	watchesOwnPods    bool
 
-	flows       []string
-	selectedIPs []string
+	flows         []string
+	selectedIPs   []string
+	selectsAllIPs bool
 }
 
 // npCacheEntry caches information about matches for a LabelSelector
@@ -303,7 +304,7 @@ func (np *networkPolicyPlugin) generateNamespaceFlows(otx ovs.Transaction, npns 
 			for _, flow := range npp.flows {
 				otx.AddFlow("table=80, priority=150, reg1=%d, %s actions=output:NXM_NX_REG2[]", npns.vnid, flow)
 			}
-			if npp.selectedIPs == nil {
+			if npp.selectsAllIPs {
 				allPodsSelected = true
 			}
 		}
@@ -493,7 +494,8 @@ func (np *networkPolicyPlugin) parseNetworkPolicy(npns *npNamespace, policy *net
 		// The rest of this function assumes that all policies affect ingress: a
 		// policy that only affects egress is, for our purposes, equivalent to one
 		// that affects ingress but does not select any pods.
-		npp.selectedIPs = []string{""}
+		npp.selectedIPs = nil
+		npp.selectsAllIPs = true
 		return npp
 	}
 
@@ -506,6 +508,7 @@ func (np *networkPolicyPlugin) parseNetworkPolicy(npns *npNamespace, policy *net
 		}
 	} else {
 		npp.selectedIPs = nil
+		npp.selectsAllIPs = true
 		destFlows = []string{""}
 	}
 
@@ -610,7 +613,7 @@ func (np *networkPolicyPlugin) updateNetworkPolicy(npns *npNamespace, policy *ne
 	oldNPP, existed := npns.policies[policy.UID]
 	npns.policies[policy.UID] = npp
 
-	changed := !existed || !reflect.DeepEqual(oldNPP.flows, npp.flows) || !reflect.DeepEqual(oldNPP.selectedIPs, npp.selectedIPs)
+	changed := !existed || !reflect.DeepEqual(oldNPP.flows, npp.flows) || !reflect.DeepEqual(oldNPP.selectedIPs, npp.selectedIPs) || oldNPP.selectsAllIPs != npp.selectsAllIPs
 	if !changed {
 		klog.V(5).Infof("NetworkPolicy %s/%s is unchanged", policy.Namespace, policy.Name)
 	}


pkg/network/node/networkpolicy.go
```diff
@@ -493,7 +493,7 @@ func (np *networkPolicyPlugin) parseNetworkPolicy(npns *npNamespace, policy *net
 		// The rest of this function assumes that all policies affect ingress: a
 		// policy that only affects egress is, for our purposes, equivalent to one
 		// that affects ingress but does not select any pods.
-		npp.selectedIPs = []string{""}
+		npp.selectedIPs = nil
```
danwinship, line 496: This won't work; `selectedIPs == nil` means something else ("all pods in the namespace are selected"; see `generateNamespaceFlows`).

We _could_ distinguish `selectedIPs = nil` from `selectedIPs = []string{}`, but that's a little sketchy. It would probably be better to have a separate `matchesAll bool` in `npPolicy`. Then if that's `true`, `selectedIPs` is ignored. (And then also the fix you have here now on top of that.)

You'll need to update the check in `updateNetworkPolicy` to check the new field too

danwinship (changes requested):

pkg/network/node/networkpolicy.go
vpickard: Thanks for the review. Working on updating this now. Just to make sure, your suggestion to add` matchesAll bool` in `npPolicy `would represent if the policy matchesAllIPs, right, so maybe a better name would be `matchesAllIPs`.

vpickard: I have updated the PR based on your comments (much appreciated!). PTAL and let me know if this is what you were thinking.

pkg/network/node/networkpolicy.go
```diff
@@ -71,8 +71,9 @@ type npPolicy struct {
 	watchesAllPods    bool
 	watchesOwnPods    bool
 
-	flows       []string
-	selectedIPs []string
+	flows         []string
+	selectedIPs   []string
+	matchesAllIPs bool
```
danwinship, line 76: sorry, didn't see your question yesterday until just now.. so actually, maybe "`selectsAllIPs`" would be better, to go with `selectedIPs`?

pkg/network/node/networkpolicy.go
```diff
@@ -303,7 +304,7 @@ func (np *networkPolicyPlugin) generateNamespaceFlows(otx ovs.Transaction, npns
 			for _, flow := range npp.flows {
 				otx.AddFlow("table=80, priority=150, reg1=%d, %s actions=output:NXM_NX_REG2[]", npns.vnid, flow)
 			}
-			if npp.selectedIPs == nil {
+			if npp.selectedIPs == nil || npp.matchesAllIPs {
 				allPodsSelected = true
```
danwinship, line 308: You don't want `||` here; the idea is that we're changing it so that `npp.selectedIPs == nil` no longer means "select all IPs". It means "select _no_ IPs". So you'd want just

    if npp.matchesAllIPs {
            allPodsSelected = true
    }

pkg/network/node/networkpolicy.go
```diff
@@ -494,6 +495,7 @@ func (np *networkPolicyPlugin) parseNetworkPolicy(npns *npNamespace, policy *net
 		// policy that only affects egress is, for our purposes, equivalent to one
 		// that affects ingress but does not select any pods.
 		npp.selectedIPs = []string{""}
+		npp.matchesAllIPs = true
```
danwinship, line 498: You need to keep your original change here too; `npp.selectedIPs = nil`

pkg/network/node/networkpolicy.go
```diff
@@ -610,7 +612,7 @@ func (np *networkPolicyPlugin) updateNetworkPolicy(npns *npNamespace, policy *ne
 	oldNPP, existed := npns.policies[policy.UID]
 	npns.policies[policy.UID] = npp
 
-	changed := !existed || !reflect.DeepEqual(oldNPP.flows, npp.flows) || !reflect.DeepEqual(oldNPP.selectedIPs, npp.selectedIPs)
+	changed := !existed || !reflect.DeepEqual(oldNPP.flows, npp.flows) || !reflect.DeepEqual(oldNPP.selectedIPs, npp.selectedIPs) || !reflect.DeepEqual(oldNPP.matchesAllIPs, npp.matchesAllIPs)
```
danwinship, line 615: right, though you don't need to use `reflect.DeepEqual` for the new check because it's not an array. You can just say `oldNPP.matchesAllIPs != npp.matchesAllIPs`

pkg/network/node/networkpolicy.go
vpickard: Sounds good. Done.

pkg/network/node/networkpolicy.go
vpickard: Done.

pkg/network/node/networkpolicy.go
vpickard: I don't quite follow this, I'll ping you on slack to follow up.

vpickard: ok, followup from slack chat, the idea is that we wouldn't be looking at `npp.selectedIPs` when `npp.matchesAllIPs` is true. And, also, we shouldn't be putting an empty ip address in `npp.selectedIPs `ever (invalid, empty ip address)

pkg/network/node/networkpolicy.go
vpickard: done

pkg/network/node/networkpolicy.go
vpickard: done

danwinship:
/lgtm cancel
NetworkPolicy tests are failing; there must be something wrong with the patch

vpickard:
> /lgtm cancel
> NetworkPolicy tests are failing; there must be something wrong with the patch

Spinning up a cluster with the patch now to run the NetworkPolicy tests to see what is going on.

PR #236, opened 2021-01-05
Bug 1903414: Do not use egressIP on reply packets

juanluisvaladas:
EgressIP namespaces should only force the egressIP when the pod is the
client. If the pod is the server we want to reply normally.

### DIFF
--- a/pkg/network/node/ovscontroller.go
+++ b/pkg/network/node/ovscontroller.go
@@ -203,6 +203,7 @@ func (oc *ovsController) SetupOVS(clusterNetworkCIDR []string, serviceNetworkCID
 	otx.AddFlow("table=100, priority=300,udp,udp_dst=%d,actions=drop", vxlanPort)
 	otx.AddFlow("table=100, priority=200,tcp,tcp_dst=53,nw_dst=%s,actions=output:2", oc.localIP)
 	otx.AddFlow("table=100, priority=200,udp,udp_dst=53,nw_dst=%s,actions=output:2", oc.localIP)
+	otx.AddFlow("table=100, priority=150,ct_state=+rpl,actions=goto_table:101")
 	// eg, "table=100, priority=100, reg0=${tenant_id}, ip, actions=set_field:${tun0_mac}->eth_dst,set_field:${egress_mark}->pkt_mark,goto_table:101"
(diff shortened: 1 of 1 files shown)


juanluisvaladas:
/Hold
I need to test this properly

abhat:
do we need an e2e to capture this bug?

pkg/network/node/ovscontroller.go
```diff
@@ -203,6 +203,7 @@ func (oc *ovsController) SetupOVS(clusterNetworkCIDR []string, serviceNetworkCID
 	otx.AddFlow("table=100, priority=300,udp,udp_dst=%d,actions=drop", vxlanPort)
 	otx.AddFlow("table=100, priority=200,tcp,tcp_dst=53,nw_dst=%s,actions=output:2", oc.localIP)
 	otx.AddFlow("table=100, priority=200,udp,udp_dst=53,nw_dst=%s,actions=output:2", oc.localIP)
+	otx.AddFlow("table=100, priority=150,ct_state=+rpl,actions=goto_table:101")
```
abhat, line 206: does this need +trk too?

abhat:
I am a bit unsure if this needs to get in to 4.7 so late in the cycle unless we have enough test coverage for egress IP as a feature. Alternatively, if we can get QE to validate that this doesn't break egress IP for openshift-sdn, we can get this in. But I would think at the very least we should have an e2e test to capture the bug itself.

huiran0826:
Verified this PR, see steps in  https://bugzilla.redhat.com/show_bug.cgi?id=1903414#c30.   Did EgressIP regression testing, no issue found.  Test run. https://polarion.engineering.redhat.com/polarion/#/project/OSE/testrun?id=OS-20210204-0117&tab=records&result=passed

abhat:
@danwinship can you PTAL as well?

pkg/network/node/ovscontroller.go
danwinship: > does this need +trk too?

no, if any other bit is set then `trk` logically has to be set, so `ct_state=+rpl+trk` would be the same as `ct_state=+rpl`

danwinship (approved):

PR #240, opened 2021-01-11
Bug 1914284: Don't try to generate NetworkPolicy flows for non-pod-network pods

danwinship:
When switching from our pod cache to the informer's, we forgot to filter out the pods that would have been filtered out before. In particular, we forgot to filter out pods that didn't have an IP address assigned yet, causing us to generate invalid flows like

> `flow add table=80, priority=150, reg1=8083932, ip, nw_dst=10.128.2.8, reg0=8083932, ip, nw_src=, tcp, tp_dst=2181,  actions=output:NXM_NX_REG2[]`

(note empty `nw_src=`)

/assign @squeed

### DIFF
--- a/pkg/network/node/networkpolicy.go
+++ b/pkg/network/node/networkpolicy.go
@@ -432,7 +432,9 @@ func (np *networkPolicyPlugin) selectPodsFromNamespaces(nsLabelSel, podLabelSel 
 			continue
 		}
 		for _, pod := range pods {
-			peerFlows = append(peerFlows, fmt.Sprintf("reg0=%d, ip, nw_src=%s, ", vnid, pod.Status.PodIP))
+			if isOnPodNetwork(pod) {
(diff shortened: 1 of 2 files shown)


pkg/network/node/networkpolicy.go
```diff
@@ -674,16 +678,18 @@ func (np *networkPolicyPlugin) watchPods() {
...
+	}
+	return pod.Status.PodIP != ""
+}
+
 func (np *networkPolicyPlugin) handleAddOrUpdatePod(obj, old interface{}, eventType watch.EventType) {
 	pod := obj.(*corev1.Pod)
 	klog.V(5).Infof("Watch %s event for Pod %q", eventType, getPodFullName(pod))
 
-	// Ignore pods with HostNetwork=true, SDN is not involved in this case
-	if pod.Spec.SecurityContext != nil && pod.Spec.HostNetwork {
```
danwinship, line 682: (The `pod.Spec.SecurityContext != nil` part here was (harmless) leftover cruft from when the code was ported from `k8s.io/kubernetes/pkg/apis/core` to `k8s.io/api/core/v1`)

PR #243, opened 2021-01-11
Bug 1915027: Fix MCS-blocking iptables rules

danwinship:
The MCS-blocking rules were too aggressive, and would block connections using 22623/22624 as a source port in some cases.

`--syn` means "SYN and not ACK", so this should make it only block attempts to establish connections to port 22623/4, and not any other traffic involving those ports.

/assign @dcbw


danwinship:
/hold
@trozet pointed out a problem

danwinship:
/hold cancel
never mind, doesn't apply to the openshift-sdn version

PR #249, opened 2021-01-21
Bug 1904455: NetworkPolicy performance fixes, v2

danwinship:
Un-reverts #228, with the addition of #240 to fix the bug.


danwinship:
/assign @dcbw @knobunc 
ok, I worked with Anurag to add some more detailed testing to the QE NetworkPolicy tests, so I think this is ready to go now

PR #251, opened 2021-01-27
Bug 1850060: Make DNS queries for egress network policy async

juanluisvaladas:
Prior to this PR we have exactly one goroutine to query all the DNS names synchronously.
If a DNS query is slow we'll block the whole synchronization goroutine causing the IP addresses of flows for the rest of the dnsNames to be potentially outdated.

### DIFF
--- a/pkg/network/common/dns.go
+++ b/pkg/network/common/dns.go
@@ -9,6 +9,7 @@ import (
 	"github.com/miekg/dns"
 
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/klog/v2"
 	utilnet "k8s.io/utils/net"
 )
 
@@ -24,6 +25,9 @@ type dnsValue struct {
 	ttl time.Duration
 	// Holds (last dns lookup time + ttl), tells when to refresh IPs next time
 	nextQueryTime time.Time
+	// Used to know if DNS.Update or DNS.Add are modifying it, so that
+	// DNS.GetNextQueryTime can ignore it
+	updating bool
 }
 
 type DNS struct {
@@ -41,6 +45,11 @@ type DNS struct {
 
 	// query timeout; overridden by tests
 	timeout time.Duration
+}
+
+type DNSResponseNotification struct {
+	Name    string
+	Changed bool
 }
 
 func NewDNS(resolverConfigFile string, ipv4, ipv6 bool) (*DNS, error) {
@@ -84,15 +93,20 @@ func (d *DNS) Get(dns string) dnsValue {
 }
 
 func (d *DNS) Add(dns string) error {
+	// This is a blocking operation, therefore must be done before acquring
+	// the lock
+	ips, ttl, err := d.getIPsAndMinTTL(dns)
+	if err != nil {
+		return err
+	}
+
 	d.lock.Lock()
 	defer d.lock.Unlock()
-
-	d.dnsMap[dns] = dnsValue{}
-	_, err := d.updateOne(dns)
-	if err != nil {
-		delete(d.dnsMap, dns)
+	d.dnsMap[dns] = dnsValue{
+		updating: true,
 	}
-	return err
+	d.updateDNSValue(dns, ips, ttl)
+	return nil
 }
 
 func (d *DNS) Delete(dns string) {
@@ -101,25 +115,56 @@ func (d *DNS) Delete(dns string) {
 	delete(d.dnsMap, dns)
 }
 
-func (d *DNS) Update(dnsName string) (bool, error) {
+func (d *DNS) SetUpdating(dns string) error {
 	d.lock.Lock()
 	defer d.lock.Unlock()
-
-	return d.updateOne(dnsName)
-}
-
-func (d *DNS) updateOne(dns string) (bool, error) {
 	res, ok := d.dnsMap[dns]
 	if !ok {
 		// Should not happen, all operations on dnsMap are synchronized by d.lock
-		return false, fmt.Errorf("DNS value not found in dnsMap for domain: %q", dns)
+		return fmt.Errorf("DNS value not found in dnsMap for domain: %q", dns)
 	}
 
+	res.updating = true
+	d.dnsMap[dns] = res
+
+	return nil
+}
+
+func (d *DNS) Update(dns string) (bool, error) {
+	// This is a blocking operation, therefore must be done before acquring
+	// the lock
 	ips, ttl, err := d.getIPsAndMinTTL(dns)
+
+	d.lock.Lock()
+	defer d.lock.Unlock()
+
 	if err != nil {
-		res.nextQueryTime = time.Now().Add(defaultTTL * time.Second)
-		d.dnsMap[dns] = res
+		d.updateNextQueryTime(dns)
 		return false, err
+	}
+
+	changed := d.updateDNSValue(dns, ips, ttl)
+	return changed, nil
+}
+
+func (d *DNS) updateNextQueryTime(dns string) {
+	res, ok := d.dnsMap[dns]
+	if !ok {
+		// Should not happen, all operations on dnsMap are synchronized by d.lock
+		klog.Errorf("DNS value not found in dnsMap for domain: %q", dns)
+		return
+	}
+	res.nextQueryTime = time.Now().Add(res.ttl)
+	res.updating = false
+	d.dnsMap[dns] = res
+}
+
+func (d *DNS) updateDNSValue(dns string, ips []net.IP, ttl time.Duration) bool {
+	res, ok := d.dnsMap[dns]
+	if !ok {
+		// Should not happen, all operations on dnsMap are synchronized by d.lock
+		klog.Errorf("DNS value not found in dnsMap for domain: %q", dns)
+		return false
 	}
 
 	changed := false
@@ -129,8 +174,9 @@ func (d *DNS) updateOne(dns string) (bool, error) {
 	res.ips = ips
 	res.ttl = ttl
 	res.nextQueryTime = time.Now().Add(res.ttl)
+	res.updating = false
 	d.dnsMap[dns] = res
-	return changed, nil
+	return changed
 }
 
 func (d *DNS) doOneQuery(server, domain string, rtype uint16) ([]net.IP, int, error) {
@@ -248,7 +294,7 @@ func (d *DNS) GetNextQueryTime() (time.Time, string, bool) {
 	var dns string
 
 	for dnsName, res := range d.dnsMap {
-		if (timeSet == false) || res.nextQueryTime.Before(minTime) {
+		if !res.updating && (timeSet == false || res.nextQueryTime.Before(minTime)) {
 			timeSet = true
 			minTime = res.nextQueryTime
 			dns = dnsName
--- a/pkg/network/common/dns.go
+++ b/pkg/network/common/dns.go
@@ -11,11 +11,16 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
 	utilnet "k8s.io/utils/net"
+	utiltrace "k8s.io/utils/trace"
 )
 
 const (
 	// defaultTTL the time (in seconds) used as a TTL if an invalid or zero TTL is provided.
 	defaultTTL = 30 * 60
+	// dnsMapTraceThreshold the grace period before warning about a slow operation
+	dnsMapTraceThreshold = 100 * time.Millisecond
+	// dnsQueryTraceThreshold the grace period before warning about a slow operation
+	dnsQueryTraceThreshold = 350 * time.Millisecond
 )
 
 type dnsValue struct {
@@ -100,6 +105,9 @@ func (d *DNS) Add(dns string) error {
 		return err
 	}
 
+	trace := utiltrace.New(fmt.Sprintf("Update resolved DNS record %q", dns))
+	defer trace.LogIfLong(dnsMapTraceThreshold)
+
 	d.lock.Lock()
 	defer d.lock.Unlock()
 	d.dnsMap[dns] = dnsValue{
@@ -110,12 +118,18 @@ func (d *DNS) Add(dns string) error {
 }
 
 func (d *DNS) Delete(dns string) {
+	trace := utiltrace.New(fmt.Sprintf("Delete DNS record %q", dns))
+	defer trace.LogIfLong(dnsMapTraceThreshold)
+
 	d.lock.Lock()
 	defer d.lock.Unlock()
 	delete(d.dnsMap, dns)
 }
 
 func (d *DNS) SetUpdating(dns string) error {
+	trace := utiltrace.New(fmt.Sprintf("SetUpdating DNS record %q", dns))
+	defer trace.LogIfLong(dnsMapTraceThreshold)
+
 	d.lock.Lock()
 	defer d.lock.Unlock()
 	res, ok := d.dnsMap[dns]
@@ -134,6 +148,9 @@ func (d *DNS) Update(dns string) (bool, error) {
 	// This is a blocking operation, therefore must be done before acquring
 	// the lock
 	ips, ttl, err := d.getIPsAndMinTTL(dns)
+
+	trace := utiltrace.New(fmt.Sprintf("Update resolved DNS record %q", dns))
+	defer trace.LogIfLong(dnsMapTraceThreshold)
 
 	d.lock.Lock()
 	defer d.lock.Unlock()
@@ -264,6 +281,9 @@ func (d *DNS) queryServer(nameserver, domain string) ([]net.IP, int, error) {
 }
 
 func (d *DNS) getIPsAndMinTTL(domain string) ([]net.IP, time.Duration, error) {
+	trace := utiltrace.New(fmt.Sprintf("DNS resolution for %q", domain))
+	defer trace.LogIfLong(dnsQueryTraceThreshold)
+
 	var ips []net.IP
 	var ttl int
 	var err error
--- a/pkg/network/common/dns.go
+++ b/pkg/network/common/dns.go
@@ -35,6 +35,16 @@ type dnsValue struct {
 	updating bool
 }
 
+type DNSInterface interface {
+	Add(dns string) error
+	Size() int
+	Get(dns string) dnsValue
+	Delete(dns string)
+	SetUpdating(dns string) error
+	Update(dns string) (bool, error)
+	GetNextQueryTime() (time.Time, string, bool)
+}
+
 type DNS struct {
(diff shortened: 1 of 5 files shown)


juanluisvaladas:
/retitle [WIP]Bug1850060: Make DNS queries for egress network policy async

pkg/network/common/egress_dns.go
```diff
@@ -117,43 +121,34 @@ func (e *EgressDNS) Sync() {
...
-				changed, err := e.Update(dnsName)
-				if err != nil {
-					utilruntime.HandleError(err)
-				}
-
-				if changed {
-					e.Updates <- updates
-				}
-				continue
+				go e.Update(dnsName)
```
abhat, line 124: why not just run go routine for doing the dns query?

juanluisvaladas: Prior to this PR we have the function `egressDNS.Sync()`, which calls `egressDNS.Update()`, which calls `DNS.Update()`. which gets blocked on `DNS.getMinIPsAndMinTTL()`.

Perhaps I'm missing something; but in order to do one goroutine per query the only two functions that can be ran in a separate goroutine are `egressDNS.Update()` or `DNS.Update()` because if we call `DNS.getMinIPsAndMinTTL()` in a separate goroutine `DNS.Update()` will still need to wait for it prior to doing the return. So instead of blocking in the function we'll block on waiting for it to write to the channel.

We could call `DNS.Update()` instead but then it would need to write to the `EgressDNS.dnsResponse` channel directly and I don't see any actual benefit. It would be harder to test and we would need to modify the tests which I managed to get without modifications.

juanluisvaladas:
Moving back to WIP.
After some manual testing I've identified a bottleneck:

```
W0128 11:57:36.753447   97626 dns.go:293] Writing to egressDNS Updates channel took too long: 5.542399362s. Expected less than 100ms
W0128 11:57:47.305604   97626 dns.go:293] Writing to egressDNS Updates channel took too long: 10.551092731s. Expected less than 100ms
W0128 11:57:57.884220   97626 dns.go:293] Writing to egressDNS Updates channel took too long: 10.576496838s. Expected less than 100ms
W0128 11:58:08.464444   97626 dns.go:293] Writing to egressDNS Updates channel took too long: 10.575839282s. Expected less than 100ms
W0128 11:58:19.042831   97626 dns.go:293] Writing to egressDNS Updates channel took too long: 10.576980075s. Expected less than 100ms
W0128 11:58:29.623615   97626 dns.go:293] Writing to egressDNS Updates channel took too long: 10.57840838s. Expected less than 100ms
W0128 11:58:40.200012   97626 dns.go:293] Writing to egressDNS Updates channel took too long: 10.57425184s. Expected less than 100ms
W0128 11:58:50.777478   97626 dns.go:293] Writing to egressDNS Updates channel took too long: 10.390045228s. Expected less than 100ms
W0128 11:59:01.355494   97626 dns.go:293] Writing to egressDNS Updates channel took too long: 10.576549115s. Expected less than 100ms
W0128 11:59:11.933716   97626 dns.go:293] Writing to egressDNS Updates channel took too long: 10.577143096s. Expected less than 100ms
W0128 11:59:22.512446   97626 dns.go:293] Writing to egressDNS Updates channel took too long: 10.578068683s. Expected less than 100ms
W0128 11:59:33.107216   97626 dns.go:293] Writing to egressDNS Updates channel took too long: 10.591565244s. Expected less than 100ms
W0128 11:59:43.696383   97626 dns.go:293] Writing to egressDNS Updates channel took too long: 10.586992445s. Expected less than 100ms
W0128 11:59:54.273864   97626 dns.go:293] Writing to egressDNS Updates channel took too long: 10.576080032s. Expected less than 100ms
W0128 12:00:04.851594   97626 dns.go:293] Writing to egressDNS Updates channel took too long: 10.576523652s. Expected less than 100ms
```
I don't know why yet.

juanluisvaladas:
I determined the previous cluster had an issue with most of the CPU stolen and everything was unreasonably slow.

I'm not able to reproduce anymore.

pkg/network/common/dns.go
```diff
@@ -24,6 +27,9 @@ type dnsValue struct {
 	ttl time.Duration
 	// Holds (last dns lookup time + ttl), tells when to refresh IPs next time
 	nextQueryTime time.Time
+	// Used to know if DNS.Update or DNS.Add are modifying it, so that
+	// DNS.GetNextQueryTime can ignore it
+	updating bool
```
abhat, line 32: should this be atomic?

pkg/network/common/dns.go
```diff
@@ -84,14 +99,23 @@ func (d *DNS) Get(dns string) dnsValue {
...
 func (d *DNS) Add(dns string) error {
+	// This is a blocking operation, therefore must be done before acquring
+	// the lock
+	ips, ttl, err := d.getIPsAndMinTTL(dns)
+	if err != nil {
+		return err
+	}
+
+	defer func(start time.Time) {
+		warnSlowOperation("Add resolved egress DNS record", start)
```
abhat, line 110: shouldn't the defer func be declared before calling getIPsAndMinTTL() if you want to warn of slow operation when making the dns query?

pkg/network/common/dns.go
```diff
@@ -101,34 +125,69 @@ func (d *DNS) Delete(dns string) {
 	delete(d.dnsMap, dns)
 }
 
-func (d *DNS) Update(dnsName string) (bool, error) {
+func (d *DNS) LockForUpdate(dns string) error {
```
abhat, line 128: `s/LockForUpdate/SetUpdating`

pkg/network/common/dns.go
```diff
@@ -101,34 +125,69 @@ func (d *DNS) Delete(dns string) {
...
+	}
+
 	changed := false
 	if !ipsEqual(res.ips, ips) {
 		changed = true
 	}
 	res.ips = ips
 	res.ttl = ttl
 	res.nextQueryTime = time.Now().Add(res.ttl)
+	res.updating = false
```
abhat, line 190: why do it twice if we are also doing it as a part of `updateNextQueryTime`?

pkg/network/common/dns.go
juanluisvaladas: What I wanted to log was precisely the bits that we can control and optimize (if the DNS is slow, we can't do anything about it).

Said so I agree with logging slow DNS queries is useful as well, but I'm not quite sure what's an appropriate threshold for that. I tried doing a few DNS queries to 8.8.8.8 and 1.1.1.1 and I'm getting somewhere between 20 and 80 ms most of the time. Do 300 milliseconds sound like a reasonable

pkg/network/common/dns.go
juanluisvaladas: That's definitely a better name

pkg/network/common/dns.go
juanluisvaladas: We shouldn't do it twice for the same update. If `getIPsAndMinTTL` returns an error we call `updateNextQueryTime`, else we call `updateDNSValue`.

pkg/network/common/dns.go
```diff
@@ -70,6 +81,10 @@ func (d *DNS) Size() int {
 }
 
 func (d *DNS) Get(dns string) dnsValue {
+	defer func(start time.Time) {
+		warnSlowOperation("Get egress DNS record", start)
```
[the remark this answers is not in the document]
juanluisvaladas: Good idea.

pkg/network/common/dns.go
juanluisvaladas: I don't think that's necessary because we're only modifying this variable when `DNS.lock` is held. So I think this is thread safe as is.

juanluisvaladas:
/retitle Bug 1850060: Make DNS queries for egress network policy async

danwinship:
> _No description provided._

please fix

pkg/network/common/egress_dns.go
```diff
@@ -98,17 +102,19 @@ func (e *EgressDNS) Delete(policy networkv1.EgressNetworkPolicy) {
 	}
 }
 
-func (e *EgressDNS) Update(dns string) (bool, error) {
-	e.lock.Lock()
-	defer e.lock.Unlock()
+func (e *EgressDNS) Update(dns string) {
```
danwinship, line 105: if this is only used internally (which it seems like it is?) it should have a lowercase name

pkg/network/common/dns.go
```diff
@@ -347,3 +361,10 @@ func removeDuplicateIPs(ips []net.IP) []net.IP {
 
 	return uniqueIPs
 }
+
+func warnSlowOperation(operation string, start time.Time) {
```
danwinship, line 387: dunno if you saw my response on slack the other day, but the thing you were looking for is `k8s.io/utils/trace`

pkg/network/common/egress_dns.go
```diff
@@ -133,7 +133,9 @@ func (e *EgressDNS) Sync() {
 		case response := <-e.dnsResponse:
 			if response.Changed {
 				// TODO: Should this be non-blocking?
+				start := time.Now()
 				e.Updates <- e.getEgressDNSUpdates(response.Name)
+				warnSlowOperation("Writing to egressDNS Updates channel", start)
```
danwinship: You could also rewrite this to use a non-blocking channel that's just used to signal the presence of updates, and a `GetUpdates()` method to get the actual changes. Like in #139.

pkg/network/common/dns.go
danwinship: > I don't think that's necessary because we're only modifying this variable when `DNS.lock` is held. So I think this is thread safe as is.

It's not necessary because you're only modifying _or reading_ the variable when the lock is held. It wouldn't be safe to read `updating` from `GetNextQueryTime` if you weren't holding the lock there.

danwinship (commented):
Some initial review but it would be good to have a description of exactly what is going wrong in 1850060, and how this fixes it

pkg/network/common/dns.go
abhat: In fact, if the only reason we hold the lock is to update the "updating" bool, we should instead make it atomic and get rid of the locks.

juanluisvaladas: > In fact, if the only reason we hold the lock is to update the "updating" bool, we should instead make it atomic and get rid of the locks.

The problem is that this variable is part of a struct in a map and we cannot guarantee that the map entry won't be deleted while we update it without locking the whole map.

pkg/network/common/dns.go
juanluisvaladas: Note to myself: Not  outdated TODO

pkg/network/common/egress_dns.go
juanluisvaladas: Note to myself: Not  outdated TODO

juanluisvaladas:
There is some kind of race condition in the unit test or the code itself, sometimes it passes. Trying to figure out the problem.

pkg/network/common/dns.go
abhat: I think that's fair. You cannot control the object's map membership by making the variable atomic.

pkg/network/common/egress_dns.go
juanluisvaladas: Now this is done on a separate goroutine, technically it's not the same thing but it achieves the same.

juanluisvaladas:
@danwinship I addressed all your previous concerns, added a test that forges out of order answers proving the egressDNS is handling the slow async queries correctly. Do you mind having another look at this?

pkg/network/common/dns.go
```diff
@@ -84,53 +107,102 @@ func (d *DNS) Get(dns string) dnsValue {
...
 
 func (d *DNS) Add(dns string) error {
+	// This is a blocking operation, therefore must be done before acquring
+	// the lock
+	ips, ttl, err := d.getIPsAndMinTTL(dns)
+	if err != nil {
+		return err
+	}
+
+	trace := utiltrace.New("Update resolved DNS record")
```
abhat, line 117: add the dns string name to the trace?

pkg/network/common/dns.go
```diff
@@ -84,53 +107,102 @@ func (d *DNS) Get(dns string) dnsValue {
...
+	trace := utiltrace.New("Delete DNS ")
+	defer trace.LogIfLong(dnsMapTraceThreshold)
+
 	d.lock.Lock()
 	defer d.lock.Unlock()
 	delete(d.dnsMap, dns)
 }
 
-func (d *DNS) Update(dnsName string) (bool, error) {
+func (d *DNS) SetUpdate(dns string) error {
```
abhat, line 138: s/SetUpdate/SetUpdating

pkg/network/common/dns.go
```diff
@@ -84,53 +107,102 @@ func (d *DNS) Get(dns string) dnsValue {
...
+	trace := utiltrace.New("SetUpdate DNS record")
```
abhat, line 139: add dns name to the trace

juanluisvaladas:
All the comments have been addressed.

pkg/network/common/dns.go
```diff
@@ -139,31 +139,32 @@ func (d *DNS) Update(dns string) (bool, error) {
...
 }
 
-func (d *DNS) updateNextQueryTime(dns string) error {
+func (d *DNS) updateNextQueryTime(dns string) {
 	res, ok := d.dnsMap[dns]
 	if !ok {
 		// Should not happen, all operations on dnsMap are synchronized by d.lock
-		return fmt.Errorf("DNS value not found in dnsMap for domain: %q", dns)
+		klog.Errorf("DNS value not found in dnsMap for domain: %q", dns)
+	} else {
```
danwinship, line 181: would be better to `return` rather than putting the rest of the function in an `else`

pkg/network/common/dns.go
```diff
@@ -264,6 +281,10 @@ func (d *DNS) queryServer(nameserver, domain string) ([]net.IP, int, error) {
 }
 
 func (d *DNS) getIPsAndMinTTL(domain string) ([]net.IP, time.Duration, error) {
+	trace := utiltrace.New(fmt.Sprintf("DNS resolution for %q", domain))
+	// Probably a bit too excessive.
+	defer trace.LogIfLong(dnsQueryTraceThreshold)
```
danwinship, line 295: we shouldn't be adding it if we think it's going to be excessive...

pkg/network/common/dns.go
```diff
@@ -34,6 +34,15 @@ type dnsValue struct {
 	// DNS.GetNextQueryTime can ignore it
 	updating bool
 }
+type DNSInterface interface {
```
danwinship, line 37: blank line before that. also, maybe use `DNS` as the name of the interface, and rename the struct? (The struct type doesn't need to be exported any more now...)

danwinship (changes requested):
Can you squash the first three commits? The second and third are basically fixups to the first.

pkg/network/common/dns.go
abhat: +1

pkg/network/common/egress_dns.go
```diff
@@ -117,43 +131,46 @@ func (e *EgressDNS) Sync() {
...
+	if response.Changed {
+		updates := e.getEgressDNSUpdates(response.Name)
+		trace.Step("getEgressDNSUpdates")
+		e.Updates <- updates
 	}
 
+}
+
+func (e *EgressDNS) getEgressDNSUpdates(dnsName string) []EgressDNSUpdate {
+	policyUpdates := make([]EgressDNSUpdate, 0)
```
abhat, line 165: This probably needs to grab a lock since it iterates over the dnsNamesToPolicies map which Add/Delete may muck around with?

abhat (changes requested):

pkg/network/common/dns.go
juanluisvaladas: Actually what I meant is the threshold is probably too high and should be lower. I removed the comment.

pkg/network/common/egress_dns.go
juanluisvaladas: Good catch! Added the lock to it.

juanluisvaladas:
> Can you squash the first three commits? The second and third are basically fixups to the first.
Yep that's a good point. Also done.

pkg/network/common/dns.go
```diff
@@ -218,6 +291,9 @@ func (d *DNS) queryServer(nameserver, domain string) ([]net.IP, int, error) {
 }
 
 func (d *DNS) getIPsAndMinTTL(domain string) ([]net.IP, time.Duration, error) {
```
abhat, line 293: `d.getIPsAndMinTTL` uses `d.nameservers` and `d.queryServer`. But it is not protected by a lock. It is probably OK since `d.nameservers` is not something that gets mucked around at all once the DNS object is created. And `d.queryServer` is not using any other shared resource.

juanluisvaladas: > d.getIPsAndMinTTL uses d.nameservers and d.queryServer. But it is not protected by a lock

If we locked the whole function we wouldn't be able to query the DNS in parallel, making the whole patch useless.  If there were a bit unsafe to call concurrently we should lock only that bit;  the rest must be unlocked so that we can call it concurrently.

> It is probably OK since d.nameservers is not something that gets mucked around at all once the DNS object is created. And d.queryServer is not using any other shared resource.

I agree, I don't think we need to lock it at all.

pkg/network/common/egress_dns_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+	update = <-egressDNS.Updates
+	if len(update) != 1 {
+		t.Errorf("Expected exactly one element in the update: %v", update)
+		// Exit the function to avoid a nil pointer dereference
+		return
+	}
+	u0 = update[0]
+	if u0.Namespace != "fake-ns-2" {
+		t.Errorf("Expecting an update for fake-ns-1 and fake-ns-2. Got: %v", update)
```
abhat, line 120: nit: update the error statement that the update is expected only for `fake-ns-2`

pkg/network/common/dns.go
```diff
@@ -84,53 +115,102 @@ func (d *DNS) Get(dns string) dnsValue {
...
+	res, ok := d.dnsMap[dns]
+	if !ok {
+		// Should not happen, all operations on dnsMap are synchronized by d.lock
+		klog.Errorf("DNS value not found in dnsMap for domain: %q", dns)
+		return false
+	}
+
 	changed := false
-	if !ipsEqual(res.ips, ips) {
+	if !d.ipsChangeGd(res.ips, dns string) {
```
abhat, line 205: typo `d.ipsChanged()`

PR #252, opened 2021-01-28
Bug 1900989: Unidling use Services annotations, switch away from Endpoints

aojea:
The openshift-sdn unidling logic was checking for the idling
annotations in the Endpoints object.
However, due to the consequences of the EndpointSlice migration
and the difference behavior of Slices with annotations, Openshift
is moving the Idling feature to use annotations on Services instead.

Signed-off-by: Antonio Ojea <aojea@redhat.com>

xref: https://github.com/openshift/openshift-controller-manager/pull/165

### DIFF
--- a/pkg/network/proxyimpl/hybrid/proxy.go
+++ b/pkg/network/proxyimpl/hybrid/proxy.go
@@ -201,8 +201,17 @@ func (p *HybridProxier) shouldEndpointsUseUserspace(endpoints *corev1.Endpoints)
 		}
 	}
 
+	// Get the Endpoints corresponding Service and checks if it has the IdledAt annotation
+	// oc idle annotates the Service object, it used to annotate the Endpoints object
+	svc, err := p.serviceLister.Services(endpoints.Namespace).Get(endpoints.Name)
+	// don´t use userspace if we are not able to find the corresponding Service
+	if err != nil {
+		return false
+	}
+
+	// use the Userspace proxy if the Service doesn´t have endpoints and has been annotated
 	if !hasEndpoints {
-		if _, ok := endpoints.Annotations[unidlingapi.IdledAtAnnotation]; ok {
+		if _, ok := svc.Annotations[unidlingapi.IdledAtAnnotation]; ok {
 			return true
 		}
 	}


aojea:
/retest
I can´t reference the other bug, the bot is unauthorised to do that

> JSONRPC error 32000: There was an error reported for the RPC call to Jira: There was an error reported for a GitHub REST call. URL: https://api.github.com/repos/openshift/sdn/pulls/252 Error: 403 Forbidden a

aojea:
/hold
I´ve talked with @danwinship  offline and he is right, 
this is missing the scenario when the Service is annotated for idling(generates an update) but there are no Endpoints already. so there is no Endpoints update. It doesn´t seem a common scenarios but is a legit one

pkg/network/proxyimpl/hybrid/proxy.go
```diff
@@ -201,8 +201,17 @@ func (p *HybridProxier) shouldEndpointsUseUserspace(endpoints *corev1.Endpoints)
 		}
 	}
 
+	// Get the Endpoints corresponding Service and checks if it has the IdledAt annotation
+	// oc idle annotates the Service object, it used to annotate the Endpoints object
+	svc, err := p.serviceLister.Services(endpoints.Namespace).Get(endpoints.Name)
```
frobware, line 206: Does this need a role binding update? I needed `Get` rights on `services` for the o-c-m change:
- https://github.com/openshift/openshift-apiserver/pull/180

aojea: it access the cache from the servicelister

aojea:
> /hold
> I´ve talked with @danwinship offline and he is right,
> this is missing the scenario when the Service is annotated for idling(generates an update) but there are no Endpoints already. so there is no Endpoints update. It doesn´t seem a common scenarios but is a legit one

@danwinship I think that this is the less disruptive way to check for the annotations in Services and the scenario not covered doesn´t seems realistic.
The option to move all the logic to sync by services instead of endpoints is more complicated, and we´ll always need to check for endpoints, until a scale down of resources, the trigger to switch the kube-proxy mode will be always endpoints=0

danwinship:
It looks like `oc idle` has been changed to annotate both the Service and the Endpoints, and thus we should not need to make any change to sdn in 4.7 (and especially, "we should not need to make any change to sdn _less than a week before final 4.7 code freeze_")

frobware:
> It looks like `oc idle` has been changed to annotate both the Service and the Endpoints, and thus we should not need to make any change to sdn in 4.7 (and especially, "we should not need to make any change to sdn _less than a week before final 4.7 code freeze_")

The router only watches endpointslices and services. When we get add/delete updates from endpointslices wee look up the service to see if it has been idled.

aojea:
> It looks like `oc idle` has been changed to annotate both the Service and the Endpoints, and thus we should not need to make any change to sdn in 4.7 (and especially, "we should not need to make any change to sdn _less than a week before final 4.7 code freeze_")

yeah, until the endpoint annotation is removed ... but I honestly lost track on which change goes when

frobware:
> yeah, until the endpoint annotation is removed we don´t need to switch to services... but I honestly lost track on which change goes when

As I understand things "today" we won't need this change but will continue with remove idle annotations when unidling on services and endpoints; https://github.com/openshift/openshift-controller-manager/pull/165

aojea:
closing, as this will be needed only when endpoint stop to be annotated

PR #262, opened 2021-02-23
Support allow-from-router feature using openshift-host-network namespace

abhat:
This commit treats `openshift-host-network` namespace as having vnid 0 for
network policy match purposes.

Signed-off-by: Aniket Bhat <anbhat@redhat.com>

### DIFF
--- a/pkg/network/node/networkpolicy.go
+++ b/pkg/network/node/networkpolicy.go
@@ -30,6 +30,8 @@ import (
 	"github.com/openshift/sdn/pkg/network/node/ovs"
 )
 
+const HostNetworkNamespace = "openshift-host-network"
+
 type networkPolicyPlugin struct {
 	node   *OsdnNode
 	vnids  *nodeVNIDMap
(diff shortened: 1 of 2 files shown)


pkg/network/master/vnids.go
```diff
@@ -39,8 +41,11 @@ func newMasterVNIDMap(allowRenumbering bool) *masterVNIDMap {
 	}
 
 	return &masterVNIDMap{
-		netIDManager:     pnetid.NewInMemory(netIDRange),
-		adminNamespaces:  sets.NewString(metav1.NamespaceDefault),
+		netIDManager: pnetid.NewInMemory(netIDRange),
+		// the HostNetworkNamespace should also be assigned vnid 0 so that it can
+		// be treated the same as default namespace to allow specifying
+		// network policies for host-network traffic.
+		adminNamespaces:  sets.NewString(metav1.NamespaceDefault, HostNetworkNamespace),
```
danwinship, line 48: This won't work. As mentioned in the enhancement, the NetworkPolicy code does not allow two namespaces to have the same VNID, and so `openshift-host-network` can't _actually_ have VNID 0. (eg, `networkPolicyPlugin.namespaces` is a map from VNID to namespace; if you have two namespaces with the same VNID, then only one of them would get added to `namespaces` and it's not even well-defined which one).

I _think_ what you want to do is hack `selectNamespacesInternal` in `networkpolicy.go` so that if `HostNetworkNamespace` matches, it sets `match.matches[npns.name] = 0` instead of `match.matches[npns.name] = vnid`. I _think_ that should have the right effect, and not have any side effects...

danwinship (changes requested):

pkg/network/master/vnids.go
abhat: @danwinship updated the PR with your suggested change. PTAL.

danwinship:
looks good but please add something to `networkpolicy_test.go` too.

PR #263, opened 2021-02-24
Bug 1933711: EgressDNS: Keep short lived records at most 30s

juanluisvaladas:
The TTL is guaranteed to be valid for its duration, however, it's not
guaranteed that the DNS server will cache it and always return the same
list of IP addreses for the whole TTL.

For records with large TTL (>30 min) we assume they will always return
the same list of addresses and query them only once every 30 min. We'll
assume rest of the records a TTL of at most 30 seconds.

### DIFF
--- a/pkg/network/common/dns.go
+++ b/pkg/network/common/dns.go
@@ -16,7 +16,7 @@ import (
 
 const (
 	// defaultTTL the time (in seconds) used as a TTL if an invalid or zero TTL is provided.
-	defaultTTL = 30 * 60
+	defaultTTL = 30
 	// dnsMapTraceThreshold the grace period before warning about a slow operation
 	dnsMapTraceThreshold = 100 * time.Millisecond
 	// dnsQueryTraceThreshold the grace period before warning about a slow operation
@@ -199,7 +199,7 @@ func (d *DNS) updateDNSValue(dns string, ips []net.IP, ttl time.Duration) bool {
 		changed = true
 	}
 	res.ips = ips
-	res.ttl = ttl
+	res.ttl = normalizeTTL(ttl)
 	res.nextQueryTime = time.Now().Add(res.ttl)
 	res.updating = false
 	d.dnsMap[dns] = res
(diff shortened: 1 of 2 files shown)


juanluisvaladas:
/retitle Bug 1933711: EgressDNS: Keep short lived records at most 30s

pkg/network/common/dns.go
```diff
@@ -400,3 +400,20 @@ func removeDuplicateIPs(ips []net.IP) []net.IP {
 
 	return uniqueIPs
 }
+
+func normalizeTTL(ttl time.Duration) time.Duration {
+
+	// The TTL is guaranteed to be valid for its duration, however, it's not
```
danwinship, line 406: (extra blank line)

pkg/network/common/dns.go
```diff
@@ -400,3 +400,20 @@ func removeDuplicateIPs(ips []net.IP) []net.IP {
...
+	// guaranteed that the DNS server will cache it and always return the same
+	// list of IP addreses for the whole TTL.
+	// For records with large TTL (>30 min) we assume they will always return
+	// the same list of addresses and query them only once every 30 min. We'll
+	// assume rest of the records a TTL of at most 30 seconds.
+
+	if ttl < 30*time.Second {
+		return ttl
+	} else if ttl >= 1800*time.Second {
+		return 1800 * time.Second
```
danwinship, line 416: seriously, this is what gofmt expects? No wonder I can never figure out if it wants spaces or not...

anyway, `30 * time.Minute` would be clearer

pkg/network/common/dns_test.go
```diff
@@ -160,8 +160,9 @@ func TestAddDNS(t *testing.T) {
 			if !ipsEqual(d.ips, test.ips) {
 				t.Fatalf("Test case: %s failed, expected IPs: %v, got: %v for the domain %q", test.testCase, test.ips, d.ips, test.domainName)
 			}
-			if d.ttl.Seconds() != test.ttl {
-				t.Fatalf("Test case: %s failed, expected TTL: %g, got: %g for the domain %q", test.testCase, test.ttl, d.ttl.Seconds(), test.domainName)
+			normalizedTTL := normalizeTTL(time.Duration(test.ttl) * time.Second)
```
danwinship, line 163: If you made `test.ttl` a `time.Duration` you could get rid of a lot of casting/multiplying

danwinship (commented):
looks good. just nitpicks

pkg/network/common/dns.go
juanluisvaladas: Yes, this is generated by gofmt :sweat_smile:

pkg/network/common/dns_test.go
```diff
@@ -508,6 +513,29 @@ func TestUpdateDNS(t *testing.T) {
...
+	tests[31] = 30
+	tests[1799] = 30
+	tests[1800] = 1800
+	tests[1801] = 1800
+	tests[3600] = 1800
+	for k, v := range tests {
+		originalTTL := time.Duration(k) * time.Second
+		expectedTTL := time.Duration(v) * time.Second
+		normalizedTTL := normalizeTTL(originalTTL)
+		if normalizedTTL != time.Duration(v)*time.Second {
```
abhat, line 533: Use `expectedTTL` for comparison.

abhat:
Mostly looks good. One minor comment.

danwinship:
/lgtm
i guess the `time.Duration` change didn't really save much, it just moved the messiness around

PR #264, opened 2021-02-24
Bug 1910378: networkpolicy: pass traffic through NAT to handle possible tuple collisions

dcbw:
If a pod makes connections to the same destination pod via a service VIP and
directly to the pod's IP address, it may be the case that both connections
choose the same source port. After the service VIP connection is DNAT-ed to
the destination pod's IP address, both connections will have the same 4-tuple.

When conntrack processes the second connection's TCP SYN and attempts to commit
the conntrack state, the commit fails due to the tuple collision.

To prevent the commit failure, make sure SNAT is done on the source port
during the commit in case of a collision. The use of a fake "0.0.0.0"
address (an undocumented "feature" of OVS that will soon be formalized)
ensures that only the source port will be touched if necessary.

Note that the packets won't be manipulated if they don't need to be, even
though they are passed through NAT.

@danwinship @tssurya @juanluisvaladas

### DIFF
--- a/pkg/network/node/networkpolicy.go
+++ b/pkg/network/node/networkpolicy.go
@@ -111,7 +111,10 @@ func (np *networkPolicyPlugin) Start(node *OsdnNode) error {
 
 	otx := node.oc.NewTransaction()
 	for _, cn := range np.node.networkInfo.ClusterNetworks {
-		otx.AddFlow("table=21, priority=200, ip, nw_dst=%s, actions=ct(commit,table=30)", cn.ClusterCIDR.String())
+		// Must pass packets through CT NAT to ensure NAT state is handled
+		// correctly by OVS when NAT-ed packets have tuple collisions.
+		// https://bugzilla.redhat.com/show_bug.cgi?id=1910378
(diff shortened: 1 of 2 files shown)


danwinship:
@dcbw can you grab https://github.com/danwinship/sdn/commit/30e534e8ba0ad179e0e583e599652ab1c1c990ee ?

The problem is that currently the rules:

	otx.AddFlow("table=30, priority=300, ip, nw_dst=%s, actions=output:2", localSubnetGateway)

and

	otx.AddFlow("table=30, priority=300, ip, nw_dst=%s, ct_state=+rpl, actions=ct(nat,table=70)", localSubnetCIDR)

have the same priority, so a reply packet to `localSubnetGateway` might match either one, but only the first flow will route it correctly. I'm not sure why this was never a problem before. (Packets that can match multiple flows have undefined behavior, so it's possible that for some reason, the latter rule gets hit reliably when you add Eelco's dummy SNAT, but the former rule got hit reliably when we didn't?) Anyway, this is specifically breaking hairpin service traffic (where the packet gets SNATted to the node IP, aka `localSubnetGateway`).

dcbw:
@danwinship grabbed your fixup, squashed, added some credits at the end of the commit message. PTAL?

dcbw:
gcp failure is quota exceeded
/retest

danwinship:
replaced by #269 
/close

PR #269, opened 2021-03-04
Bug 1910378: networkpolicy: pass traffic through NAT to handle possible tuple collisions

danwinship:
If a pod makes connections to the same destination pod via a service VIP and
directly to the pod's IP address, it may be the case that both connections
choose the same source port. After the service VIP connection is DNAT-ed to
the destination pod's IP address, both connections will have the same 4-tuple.

When conntrack processes the second connection's TCP SYN and attempts to commit
the conntrack state, the commit fails due to the tuple collision.

To prevent the commit failure, make sure SNAT is done on the source port
during the commit in case of a collision. The use of a fake "0.0.0.0"
address (an undocumented "feature" of OVS that will soon be formalized)
ensures that only the source port will be touched if necessary.

Note that the packets won't be manipulated if they don't need to be, even
though they are passed through NAT.

Flow change suggested by Eelco Chaudron and Flavio Leitner
Additional fixes by Dan Winship

### DIFF
--- a/pkg/network/node/networkpolicy.go
+++ b/pkg/network/node/networkpolicy.go
@@ -111,7 +111,10 @@ func (np *networkPolicyPlugin) Start(node *OsdnNode) error {
 
 	otx := node.oc.NewTransaction()
 	for _, cn := range np.node.networkInfo.ClusterNetworks {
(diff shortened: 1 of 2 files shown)


danwinship:
/retest
(looks like not our problem; the failing test is flaky in https://testgrid.k8s.io/redhat-openshift-ocp-release-4.8-blocking#release-openshift-ocp-installer-e2e-aws-4.8)

PR #270, opened 2021-03-04
DO NOT MERGE: UPSTREAM: 97238: Handle terminating endpoints

smarterclayton:
This is a test cherry-pick based on the current vendor state containing
upstream 97238, which allows the proxier to handle terminating endpoints.
This is not sufficient by itself because we need to test endpoint slices,
but ensures the right code is in place.


smarterclayton:
With the latest changes (including openshift/kubernetes#607, green https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp/1368691225044979712

Still seeing some disruption, I think this is that during a rolling update of the deployment the last previous running is deleted seconds after the new one goes running, which I believe is causing the cloud LB to drop all endpoints (and thus serve dead endpoints) until the new pod passes a health check.  I think we may need to set `minReadySeconds > lb health check max interval for ready` on all pods that are behind service load balancers.  At least in basic testing, a 10s min ready was enough to take a 1/3 flake on unready oauth when the router was rolled out to not happen, so I need to look in a bit more detail and construct a timeline then test the theory.

danwinship:
/close
sdn is now rebased to 1.22-rc.0, which includes 97238

PR #271, opened 2021-03-04
WIP: Support EndpointSlice in sdn and test handling terminating endpoints

smarterclayton:
86ceaca1b (Clayton Coleman, 5 minutes ago)
   DO NOT MERGE: Force EndpointSliceProxying on

bf1fd9bd2 (Clayton Coleman, 12 minutes ago)
   DO NOT MERGE: UPSTREAM: 97238: Handle terminating endpoints

   This is a test cherry-pick based on the current vendor state containing 
   upstream 97238, which allows the proxier to handle terminating endpoints. 
   This is not sufficient by itself because we need to test endpoint slices, 
   but ensures the right code is in place.


smarterclayton:
@Miciah this i where i'm testing it

smarterclayton:
```
E0304 23:42:19.517393    2983 reflector.go:138] k8s.io/client-go@v0.18.6/tools/cache/reflector.go:167: Failed to watch *v1beta1.EndpointSlice: failed to list *v1beta1.EndpointSlice: endpointslices.discovery.k8s.io is forbidden: User "system:serviceaccount:openshift-sdn:sdn" cannot list resource "endpointslices" in API group "discovery.k8s.io" at the cluster scope
```

https://github.com/openshift/cluster-network-operator/pull/1003

smarterclayton:
Panic from https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp/1367709628095270912

```
E0305 06:11:58.595222   21469 runtime.go:78] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 311 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic(0x1c0e9c0, 0x2e01a50)
	k8s.io/apimachinery@v0.18.6/pkg/util/runtime/runtime.go:74 +0x95
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
	k8s.io/apimachinery@v0.18.6/pkg/util/runtime/runtime.go:48 +0x89
panic(0x1c0e9c0, 0x2e01a50)
	runtime/panic.go:969 +0x1b9
k8s.io/kubernetes/pkg/proxy.(*EndpointSliceCache).updatePending(0x0, 0xc0006e1000, 0x0, 0xc000608200)
	k8s.io/kubernetes@v1.20.0-rc.0/pkg/proxy/endpointslicecache.go:167 +0x22d
k8s.io/kubernetes/pkg/proxy.(*EndpointChangeTracker).EndpointSliceUpdate(0xc000437f80, 0xc0006e1000, 0x0, 0xc000520600)
	k8s.io/kubernetes@v1.20.0-rc.0/pkg/proxy/endpoints.go:260 +0x29f
k8s.io/kubernetes/pkg/proxy/iptables.(*Proxier).OnEndpointSliceAdd(0xc000873c80, 0xc0006e1000)
	k8s.io/kubernetes@v1.20.0-rc.0/pkg/proxy/iptables/proxier.go:628 +0x3d
github.com/openshift/sdn/pkg/network/proxyimpl/hybrid.(*HybridProxier).OnEndpointSliceAdd(0xc0006073b0, 0xc0006e1000)
	github.com/openshift/sdn/pkg/network/proxyimpl/hybrid/proxy.go:268 +0x522
github.com/openshift/sdn/pkg/network/proxy.(*OsdnProxy).OnEndpointSliceAdd(0xc0007d0360, 0xc0006e1000)
	github.com/openshift/sdn/pkg/network/proxy/proxy.go:332 +0x142
k8s.io/kubernetes/pkg/proxy/config.(*EndpointSliceConfig).handleAddEndpointSlice(0xc00050f120, 0x1e79cc0, 0xc0006e1000)
	k8s.io/kubernetes@v1.20.0-rc.0/pkg/proxy/config/config.go:244 +0x13b
k8s.io/client-go/tools/cache.ResourceEventHandlerFuncs.OnAdd(...)
	k8s.io/client-go@v0.18.6/tools/cache/controller.go:231
```

Looks ilke 

```
	endpointSlicesEnabled := utilfeature.DefaultFeatureGate.Enabled(features.WindowsEndpointSliceProxying)
```

is assumed to equal feature gate enablement, because NewEndpointChangeTracker() doesn't get created.  Definitely an upstream bug, but won't block testing here.

smarterclayton:
https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp/1367988272344600576

clean install it looks like

smarterclayton:
well, test went green except for an unidling test (i may have broken it just because of the use of the slice for annotations instead of endpoints).  will kick some upgrade jobs and see how it plays out as well as a network stress test

smarterclayton:
Ok, got one run of e2e two runs of upgrade with endpoint slices on, but the termination gate off, and behavior was similar to endpoints except for the idling bug.  Now testing with termination gate on in https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-aws/1368251013236002816 as an upgrade and https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-aws/1368251449502339072 as an e2e

smarterclayton:
Passes on those (for the factors that matter) without EndpointSliceTerminating set on the server side (so we're safe to roll this out first to Kube-proxy).

pkg/network/proxyimpl/hybrid/proxy.go
```diff
@@ -244,11 +246,11 @@ func (p *HybridProxier) switchService(name types.NamespacedName) {
 	p.switchedToUserspace[name] = p.usingUserspace[name]
 }
 
-func (p *HybridProxier) OnEndpointsAdd(endpoints *corev1.Endpoints) {
+func (p *HybridProxier) OnEndpointSliceAdd(endpoints *discoveryv1beta1.EndpointSlice) {
 	// we track all endpoints in the unidling endpoints handler so that we can succesfully
 	// detect when a service become unidling
 	klog.V(6).Infof("hybrid proxy: (always) add ep %s/%s in unidling proxy", endpoints.Namespace, endpoints.Name)
-	p.unidlingProxy.OnEndpointsAdd(endpoints)
+	p.unidlingProxy.OnEndpointSliceAdd(endpoints)
```
aojea, line 253: I'm not familiar at all with the sdn/proxy implementation, maybe this information is redundant, but can be multiple slices for the same service, and each slice can have duplicate endpoints, kube-proxy uses a cache
https://github.com/kubernetes/kubernetes/blob/2bcbc527a760106ec89647fcf6852f37c804f4ed/pkg/proxy/endpointslicecache.go#L43-L49

aojea: the limit of endpoints per slice is 100, so if you have more than 100 endpoints,  let's say 110 for service X you'll receive two slices Y1 and Y2, maybe with 100 and 10 endpoints each

danwinship (changes requested):
Idling is broken because this flips us to only using EndpointSlice, but the userspace proxy (which is used by idling) doesn't support EndpointSlice.

vendor/k8s.io/kubernetes/cmd/kube-proxy/app/server.go
```diff
@@ -245,6 +245,11 @@ func (o *Options) Complete() error {
 		return err
 	}
 
+	// DO NOT MERGE: hack endpoint slice on
```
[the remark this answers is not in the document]
danwinship: oh, wait, no, Aniket already reverted that a while back; we'll need to revert the relevant part of https://github.com/openshift/cluster-network-operator/pull/905

pkg/network/proxyimpl/hybrid/proxy.go
```diff
@@ -27,11 +28,28 @@ type RunnableProxy interface {
...
+var _ proxyconfig.EndpointsHandler = &NoopEndpointsHandler{}
+
 // HybridProxier runs an unidling proxy and a primary proxy at the same time,
 // delegating idled services to the unidling proxy and other services to the
 // primary proxy.
 type HybridProxier struct {
-	proxyconfig.NoopEndpointSliceHandler
+	// TODO implement https://github.com/kubernetes/enhancements/pull/640
+	proxyconfig.NoopNodeHandler
+	NoopEndpointsHandler
```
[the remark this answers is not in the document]
smarterclayton: I updated userspace proxier to use EndpointSlice, I thought we were already going to have to switch to use Service instead of Endpoints for idling.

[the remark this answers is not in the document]
smarterclayton: Oops, I see what you mean.  Why wasn't userspace proxier updated?  Just no one signed up for it?

pkg/network/proxyimpl/hybrid/proxy.go
smarterclayton: We really should have an e2e test in upstream that creates a service with > 100 endpoints then to exercise this.  Is there one you know of I can crib?

smarterclayton: Actually, for idling, wondering whether we even need to support > three or four endpoints.  The only time the user space proxy should be in play is on an idle service which has no endpoints.

aojea: > her we even need to support > three or four endpoints.

as I said, I'm not familiar with this code, just raising some points that I think may be taking into consideration, if that is the case, it seems we should't worry about this

> We really should have an e2e test in upstream that creates a service with > 100 endpoints

this is well tested upstream

pkg/network/proxyimpl/hybrid/proxy.go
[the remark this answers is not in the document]
danwinship: > Why wasn't userspace proxier updated? Just no one signed up for it?

Upstream doesn't care about the userspace proxier any more (Tim would probably have already deleted it if OCP wasn't using it for unidling) and Red Hat had thought we weren't going to have to use EndpointSlice in openshift-sdn, so we didn't care about updating it either.

At any rate, I think we don't actually need to update userspace to use EndpointSlice; we just need to make `HybridProxier` and `OsdnProxy` pass both endpoint events and endpointslice events down to their wrapper proxiers, and then eventually the iptables proxy will act on the endpointslice events and the userspace proxy will act on the endpoint events.

pkg/network/proxyimpl/hybrid/proxy.go
danwinship: The unidling proxy can just ignore the endpointslices and work with the endpoints like it always did. Endpoints objects always contain the full set of endpoints, even in cases where the EndpointSlice controller would start splitting things up; it just means that code working with the Endpoints objects doesn't get the efficiency wins that code working with the EndpointSlice objects would get.

danwinship:
FYI #296 is a more complete EndpointSlice PR

danwinship:
/close
sdn now uses EndpointSlices, and is rebased to 1.22-rc.0, which has 97238

PR #277, opened 2021-03-17
egress IP traffic load-balancing 

JacobTanenbaum:
By using ovs groups transition openshift-sdn from using one egressIP at a time to load-balancing between all available egressIPs

### DIFF
--- a/pkg/network/common/egressip.go
+++ b/pkg/network/common/egressip.go
@@ -34,10 +34,16 @@ type nodeEgress struct {
 }
 
 type namespaceEgress struct {
-	vnid         uint32
-	requestedIPs []string
+	vnid              uint32
+	requestedIPs      []string
+	shouldDropTraffic bool
 
-	activeEgressIP string
+	activeEgressIPs []EgressIPAssignment
+}
+
+type EgressIPAssignment struct {
+	NodeIP   string
+	EgressIP string
 }
 
 type egressIPInfo struct {
@@ -59,7 +65,7 @@ type EgressIPWatcher interface {
 
 	SetNamespaceEgressNormal(vnid uint32)
 	SetNamespaceEgressDropped(vnid uint32)
-	SetNamespaceEgressViaEgressIP(vnid uint32, egressIP, nodeIP string)
+	SetNamespaceEgressViaEgressIPs(vnid uint32, activeEgressIPs []EgressIPAssignment)
 
 	UpdateEgressCIDRs()
 }
@@ -420,43 +426,44 @@ func (eit *EgressIPTracker) syncEgressNodeState(eg *egressIPInfo, active bool) {
 
 func (eit *EgressIPTracker) syncEgressNamespaceState(ns *namespaceEgress) {
 	if len(ns.requestedIPs) == 0 {
-		if ns.activeEgressIP != "" {
-			ns.activeEgressIP = ""
+		if len(ns.activeEgressIPs) != 0 || ns.shouldDropTraffic {
+			ns.activeEgressIPs = []EgressIPAssignment{}
+			ns.shouldDropTraffic = false
 			eit.watcher.SetNamespaceEgressNormal(ns.vnid)
 		}
 		return
 	}
 
-	var active *egressIPInfo
+	activeEgressIPs := make([]EgressIPAssignment, 0, len(ns.requestedIPs))
 	for _, ip := range ns.requestedIPs {
 		eg := eit.egressIPs[ip]
 		if eg == nil {
 			continue
 		}
 		if len(eg.namespaces) > 1 {
-			active = nil
 			klog.V(4).Infof("VNID %d gets no egress due to multiply-assigned egress IP %s", ns.vnid, eg.ip)
+			activeEgressIPs = nil
 			break
 		}
-		if active == nil {
-			if eg.assignedNodeIP == "" {
-				klog.V(4).Infof("VNID %d cannot use unassigned egress IP %s", ns.vnid, eg.ip)
-			} else if len(ns.requestedIPs) > 1 && eg.nodes[0].offline {
-				klog.V(4).Infof("VNID %d cannot use egress IP %s on offline node %s", ns.vnid, eg.ip, eg.assignedNodeIP)
-			} else {
-				active = eg
-			}
+		if eg.assignedNodeIP == "" {
+			klog.V(4).Infof("VNID %d cannot use unassigned egress IP %s", ns.vnid, eg.ip)
+		} else if len(ns.requestedIPs) > 1 && eg.nodes[0].offline {
+			klog.V(4).Infof("VNID %d cannot use egress IP %s on offline node %s", ns.vnid, eg.ip, eg.assignedNodeIP)
+		} else {
+			activeEgressIPs = append(activeEgressIPs, EgressIPAssignment{NodeIP: eg.assignedNodeIP, EgressIP: eg.ip})
 		}
 	}
 
-	if active != nil {
-		if ns.activeEgressIP != active.ip {
-			ns.activeEgressIP = active.ip
-			eit.watcher.SetNamespaceEgressViaEgressIP(ns.vnid, active.ip, active.assignedNodeIP)
+	if len(activeEgressIPs) > 0 {
+		if !activeEgressIPsTheSame(ns.activeEgressIPs, activeEgressIPs) {
+			ns.activeEgressIPs = activeEgressIPs
+			ns.shouldDropTraffic = false
+			eit.watcher.SetNamespaceEgressViaEgressIPs(ns.vnid, ns.activeEgressIPs)
 		}
 	} else {
-		if ns.activeEgressIP != "dropped" {
-			ns.activeEgressIP = "dropped"
+		if !ns.shouldDropTraffic {
+			ns.activeEgressIPs = []EgressIPAssignment{}
+			ns.shouldDropTraffic = true
 			eit.watcher.SetNamespaceEgressDropped(ns.vnid)
 		}
 	}
@@ -677,4 +684,26 @@ func (eit *EgressIPTracker) ReallocateEgressIPs() map[string][]string {
 	}
 
 	return allocation
+}
+
+func activeEgressIPsTheSame(oldEIPs, newEIPs []EgressIPAssignment) bool {
+	if len(oldEIPs) != len(newEIPs) {
+		return false
+	}
+
+	for _, olderEIPAssignment := range oldEIPs {
+		found := false
+		for _, newerEIPAssignment := range newEIPs {
+			if newerEIPAssignment == olderEIPAssignment {
+				found = true
+				break
+			}
+		}
+		if !found {
+			return false
+		}
+	}
+
+	return true
+
 }
--- a/pkg/network/common/egressip_test.go
+++ b/pkg/network/common/egressip_test.go
@@ -34,8 +34,10 @@ func (w *testEIPWatcher) SetNamespaceEgressDropped(vnid uint32) {
 	w.changes = append(w.changes, fmt.Sprintf("namespace %d dropped", int(vnid)))
 }
 
-func (w *testEIPWatcher) SetNamespaceEgressViaEgressIP(vnid uint32, egressIP, nodeIP string) {
-	w.changes = append(w.changes, fmt.Sprintf("namespace %d via %s on %s", int(vnid), egressIP, nodeIP))
+func (w *testEIPWatcher) SetNamespaceEgressViaEgressIPs(vnid uint32, activeEgressIPs []EgressIPAssignment) {
+	for _, activeEgressIP := range activeEgressIPs {
+		w.changes = append(w.changes, fmt.Sprintf("namespace %d via %s on %s", int(vnid), activeEgressIP.EgressIP, activeEgressIP.NodeIP))
+	}
 }
 
 func (w *testEIPWatcher) UpdateEgressCIDRs() {
@@ -362,38 +364,38 @@ func TestMultipleNamespaceEgressIPs(t *testing.T) {
 		t.Fatalf("%v", err)
 	}
 
-	// Now assigning that IP to a node should switch OVS to use that since it's first in the list
+	// Now assigning that IP to a node should cause ovs to add both of them
 	updateHostSubnetEgress(eit, &networkv1.HostSubnet{
 		HostIP:    "172.17.0.4",
 		EgressIPs: []networkv1.HostSubnetEgressIP{"172.17.0.101"},
 	})
 	err = w.assertChanges(
 		"claim 172.17.0.101 on 172.17.0.4 for namespace 42",
+		"namespace 42 via 172.17.0.100 on 172.17.0.3",
 		"namespace 42 via 172.17.0.101 on 172.17.0.4",
 	)
 	if err != nil {
 		t.Fatalf("%v", err)
 	}
 
-	// Swapping the order in the NetNamespace should swap back
+	// Swapping the order in the NetNamespace should do nothing since we are using both of them
 	updateNetNamespaceEgress(eit, &networkv1.NetNamespace{
 		NetID:     42,
 		EgressIPs: []networkv1.NetNamespaceEgressIP{"172.17.0.100", "172.17.0.101"},
 	})
-	err = w.assertChanges(
-		"namespace 42 via 172.17.0.100 on 172.17.0.3",
-	)
+	err = w.assertNoChanges()
 	if err != nil {
 		t.Fatalf("%v", err)
 	}
 
-	// Removing the inactive egress IP from its node should have no effect
+	// Removing the 1 egress IP from its node should cause it to be removed
 	updateHostSubnetEgress(eit, &networkv1.HostSubnet{
 		HostIP:    "172.17.0.4",
 		EgressIPs: []networkv1.HostSubnetEgressIP{"172.17.0.200"},
 	})
 	err = w.assertChanges(
 		"release 172.17.0.101 on 172.17.0.4",
+		"namespace 42 via 172.17.0.100 on 172.17.0.3",
 	)
 	if err != nil {
 		t.Fatalf("%v", err)
@@ -412,7 +414,7 @@ func TestMultipleNamespaceEgressIPs(t *testing.T) {
 		t.Fatalf("%v", err)
 	}
 
-	// Now add the egress IPs back...
+	// Now add the egress IPs back...this will cause them to all be added
 	updateHostSubnetEgress(eit, &networkv1.HostSubnet{
 		HostIP:    "172.17.0.3",
 		EgressIPs: []networkv1.HostSubnetEgressIP{"172.17.0.100"},
@@ -425,13 +427,15 @@ func TestMultipleNamespaceEgressIPs(t *testing.T) {
 		"claim 172.17.0.100 on 172.17.0.3 for namespace 42",
 		"claim 172.17.0.101 on 172.17.0.4 for namespace 42",
 		"namespace 42 via 172.17.0.100 on 172.17.0.3",
+		"namespace 42 via 172.17.0.100 on 172.17.0.3",
+		"namespace 42 via 172.17.0.101 on 172.17.0.4",
 	)
 	if err != nil {
 		t.Fatalf("%v", err)
 	}
 
-	// Assigning either the used or the unused Egress IP to another namespace should
-	// break this namespace
+	// Assigning either of the used Egress IP to another namespace should break that
+	// specific Egress IP on both namespaces
 	updateNetNamespaceEgress(eit, &networkv1.NetNamespace{
 		NetID:     43,
 		EgressIPs: []networkv1.NetNamespaceEgressIP{"172.17.0.100"},
@@ -449,6 +453,7 @@ func TestMultipleNamespaceEgressIPs(t *testing.T) {
 	err = w.assertChanges(
 		"claim 172.17.0.100 on 172.17.0.3 for namespace 42",
 		"namespace 42 via 172.17.0.100 on 172.17.0.3",
+		"namespace 42 via 172.17.0.101 on 172.17.0.4",
 		"namespace 43 normal",
 	)
 	if err != nil {
@@ -472,6 +477,7 @@ func TestMultipleNamespaceEgressIPs(t *testing.T) {
 	err = w.assertChanges(
 		"claim 172.17.0.101 on 172.17.0.4 for namespace 42",
 		"namespace 42 via 172.17.0.100 on 172.17.0.3",
+		"namespace 42 via 172.17.0.101 on 172.17.0.4",
 		"namespace 44 normal",
 	)
 	if err != nil {
@@ -744,12 +750,13 @@ func TestOfflineEgressIPs(t *testing.T) {
 		"claim 172.17.0.100 on 172.17.0.3 for namespace 42",
 		"claim 172.17.0.101 on 172.17.0.4 for namespace 42",
 		"namespace 42 via 172.17.0.100 on 172.17.0.3",
+		"namespace 42 via 172.17.0.101 on 172.17.0.4",
 	)
 	if err != nil {
 		t.Fatalf("%v", err)
 	}
 
-	// If the primary goes offline, fall back to the secondary
+	// If the primary goes offline, just drop that one
 	eit.SetNodeOffline("172.17.0.3", true)
 	err = w.assertChanges(
 		"namespace 42 via 172.17.0.101 on 172.17.0.4",
@@ -776,18 +783,21 @@ func TestOfflineEgressIPs(t *testing.T) {
 		t.Fatalf("%v", err)
 	}
 
-	// If the primary comes back, use it in preference to the secondary
+	// If the primary comes back, use both
 	eit.SetNodeOffline("172.17.0.3", false)
 	err = w.assertChanges(
 		"namespace 42 via 172.17.0.100 on 172.17.0.3",
+		"namespace 42 via 172.17.0.101 on 172.17.0.4",
 	)
 	if err != nil {
 		t.Fatalf("%v", err)
 	}
 
-	// If the secondary goes offline now, we don't care
+	// If the secondary goes offline now we do care
 	eit.SetNodeOffline("172.17.0.4", true)
-	err = w.assertNoChanges()
+	err = w.assertChanges(
+		"namespace 42 via 172.17.0.100 on 172.17.0.3",
+	)
 	if err != nil {
 		t.Fatalf("%v", err)
 	}
--- a/pkg/network/node/egressip.go
+++ b/pkg/network/node/egressip.go
@@ -34,6 +34,11 @@ type egressIPWatcher struct {
 	testModeChan chan string
 }
 
+type egressIPMetaData struct {
+	nodeIP     string
+	packetMark string
+}
+
 func newEgressIPWatcher(oc *ovsController, localIP string, masqueradeBit *int32) *egressIPWatcher {
 	eip := &egressIPWatcher{
 		oc:      oc,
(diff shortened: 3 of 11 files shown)


pkg/network/node/ovs/ovs.go
```diff
@@ -131,11 +141,11 @@ func New(execer exec.Interface, bridge string) (Interface, error) {
 }
 
 func (ovsif *ovsExec) execWithStdin(cmd string, stdinArgs []string, args ...string) (string, error) {
-	logLevel := klog.Level(4)
+	//logLevel := klog.Level(4)
```
squeed, line 144: did you mean to commit these comments?

squeed:
how does the load-balancing work? is it simple ECMP with some kind of 5-tuple hashing? is it stateful?

pkg/network/node/ovs/ovs.go
JacobTanenbaum: I did not, I took them out.

JacobTanenbaum:
@squeed it is simple ECMP with 5-tuple hashing

alexanderConstantinescu (changes requested):
I tested this patch in a KIND environment and it seems to work. I will do another pass tomorrow probably, but for now I had these remarks.

pkg/network/node/ovscontroller.go
```diff
@@ -767,35 +767,51 @@ func (oc *ovsController) ensureTunMAC() error {
...
-			return err
+	_ = oc.ovs.DeleteGroup(vnid)
+
+	var buildBuckets []string
+	for nodeIP, markedEgressIP := range marks {
+		if nodeIP == oc.localIP {
+			if err := oc.ensureTunMAC(); err != nil {
+				return err
+			}
+			buildBuckets = append(buildBuckets, fmt.Sprintf("actions=set_field:%s->eth_dst,set_field:%s->pkt_mark,output:tun0", oc.tunMAC, markedEgressIP))
```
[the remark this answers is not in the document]
JacobTanenbaum: should this be something that we allow the user to specify or in the future allow users to set the weights?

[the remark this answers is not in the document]
alexanderConstantinescu: > should this be something we allow the users to set

No. 

>  would this lead to unexpected behavior to the cluster/project admin if the traffic the sdn shapes the traffic in a way that is unexpected?

Well, if people think that the load balancing will be equal between egress nodes, then that would not hold if we do this. We would _sometimes_ send packets to other nodes, but the local node would be preferred. 

Anyways, don't implement that yet. I meant for it to be a discussion between all of us and giving people reviewing this going forwards, a chance to express their ideas.

pkg/network/node/ovs/ovs.go
```diff
@@ -268,6 +278,23 @@ func (ovsif *ovsExec) DeletePort(port string) error {
 	return err
 }
 
+func (ovsif *ovsExec) AddGroup(groupID uint32, groupType string, buckets []string) error {
+	_, err := ovsif.exec(OVS_OFCTL, "add-group", ovsif.bridge, fmt.Sprintf("group_id=%d,type=%s,bucket=%s", groupID, groupType, strings.Join(buckets[:], "bucket=")))
```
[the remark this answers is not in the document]
JacobTanenbaum: done

pkg/network/common/egressip.go
```diff
@@ -420,43 +421,44 @@ func (eit *EgressIPTracker) syncEgressNodeState(eg *egressIPInfo, active bool) {
...
-			ns.activeEgressIP = ""
+		if ns.activeEgressIP != nil {
+			ns.activeEgressIP = nil
 			eit.watcher.SetNamespaceEgressNormal(ns.vnid)
 		}
 		return
 	}
 
-	var active *egressIPInfo
+	//var active *egressIPInfo
```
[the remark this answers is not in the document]
JacobTanenbaum: done

pkg/network/common/egressip.go
```diff
@@ -420,43 +421,44 @@ func (eit *EgressIPTracker) syncEgressNodeState(eg *egressIPInfo, active bool) {
...
-				klog.V(4).Infof("VNID %d cannot use egress IP %s on offline node %s", ns.vnid, eg.ip, eg.assignedNodeIP)
-			} else {
-				active = eg
-			}
+		if eg.assignedNodeIP == "" {
+			klog.V(4).Infof("VNID %d cannot use unassigned egress IP %s", ns.vnid, eg.ip)
+			continue
+		} else if len(ns.requestedIPs) > 1 && eg.nodes[0].offline {
+			klog.V(4).Infof("VNID %d cannot use egress IP %s on offline node %s", ns.vnid, eg.ip, eg.assignedNodeIP)
+			continue
```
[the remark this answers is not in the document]
JacobTanenbaum: done

pkg/network/common/egressip.go
```diff
@@ -420,43 +421,44 @@ func (eit *EgressIPTracker) syncEgressNodeState(eg *egressIPInfo, active bool) {
...
-			if eg.assignedNodeIP == "" {
-				klog.V(4).Infof("VNID %d cannot use unassigned egress IP %s", ns.vnid, eg.ip)
-			} else if len(ns.requestedIPs) > 1 && eg.nodes[0].offline {
-				klog.V(4).Infof("VNID %d cannot use egress IP %s on offline node %s", ns.vnid, eg.ip, eg.assignedNodeIP)
-			} else {
-				active = eg
-			}
+		if eg.assignedNodeIP == "" {
+			klog.V(4).Infof("VNID %d cannot use unassigned egress IP %s", ns.vnid, eg.ip)
+			continue
```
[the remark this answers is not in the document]
JacobTanenbaum: done

pkg/network/node/egressip.go
```diff
@@ -151,9 +151,15 @@ func (eip *egressIPWatcher) SetNamespaceEgressDropped(vnid uint32) {
 	}
 }
 
-func (eip *egressIPWatcher) SetNamespaceEgressViaEgressIP(vnid uint32, egressIP, nodeIP string) {
-	mark := eip.iptablesMark[egressIP]
-	if err := eip.oc.SetNamespaceEgressViaEgressIP(vnid, nodeIP, mark); err != nil {
+//func (eip *egressIPWatcher) SetNamespaceEgressViaEgressIP(vnid uint32, egressIP, nodeIP string) {
```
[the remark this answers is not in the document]
JacobTanenbaum: done

pkg/network/node/egressip.go
```diff
@@ -151,9 +151,15 @@ func (eip *egressIPWatcher) SetNamespaceEgressDropped(vnid uint32) {
...
+func (eip *egressIPWatcher) SetNamespaceEgressViaEgressIP(vnid uint32, activeEgressIPs map[string]string) {
+	marks := make(map[string]string)
+	for egressIP, nodeIP := range activeEgressIPs {
+		//mark := eip.iptablesMark[egressIP]
```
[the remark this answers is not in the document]
JacobTanenbaum: done

pkg/network/node/ovs/ovs_test.go
```diff
@@ -336,3 +336,18 @@ ofport              : 3
...
 	}
 }
+
+func TestAddGroup(t *testing.T) {
+	fexec := normalSetup()
+	ovsif, err := New(fexec, "br0")
+	if err != nil {
+		t.Fatalf("Unexpected error from command: %v", err)
+	}
+	addTestResult(t, fexec, "ovs-ofctl -O OpenFlow13 add-group br0 group_id=32,type=select,bucket=weight=100,actions=\"ct(commit),move:NXM_NX_REG0[]->NXM_NX_TUN_ID[0..31],set_field:172.21.0.4->tun_dst,output:vxlan0\"bucket=bucket=weight=100,actions=\"set_field:8a:6e:23:3b:ad:3e->eth_dst,set_field:0x124a61c->pkt_mark,output:tun0\"", "", nil)
```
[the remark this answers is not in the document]
JacobTanenbaum: fixed the test so that the implementation matches the test

pkg/network/common/egressip.go
```diff
@@ -37,7 +38,7 @@ type namespaceEgress struct {
 	vnid         uint32
 	requestedIPs []string
 
-	activeEgressIP string
+	activeEgressIP map[string]string
```
danwinship, line 41: so you use maps for this everywhere, but no one ever uses it actually as a _map_... it's more like an array of node-ip/egress-ip pairs. It might be better to make that explicit, and have

    type EgressIPAssignment {
            NodeIP   string
            EgressIP string
    }
    ...
    activeEgressIPs []EgressIPAssignment

(in particular, when it's a `map[string]string` it's really not obvious whether it's egressIP→nodeIP or nodeIP→egressIP)

pkg/network/common/egressip.go
```diff
@@ -420,43 +421,41 @@ func (eit *EgressIPTracker) syncEgressNodeState(eg *egressIPInfo, active bool) {
...
+	if len(activeEgressIPs) > 0 {
+		if !reflect.DeepEqual(ns.activeEgressIP, activeEgressIPs) {
+			ns.activeEgressIP = activeEgressIPs
+			eit.watcher.SetNamespaceEgressViaEgressIP(ns.vnid, activeEgressIPs)
 		}
 	} else {
-		if ns.activeEgressIP != "dropped" {
-			ns.activeEgressIP = "dropped"
+		if ns.activeEgressIP["dropped"] != "dropped" {
+			ns.activeEgressIP = map[string]string{"dropped": "dropped"}
```
danwinship, line 458: meh, this was an ugly hack before and it's an uglier hack now. should probably have a separate field for this... 

(The important thing going on here is that we need to be able to distinguish "there are `requestedIPs` but no `activeEgressIPs` and we haven't called `SetNamespaceEgressDropped` yet" from "there are `requestedIPs` but no `activeEgressIPs` and we _have_ already called `SetNamespaceEgressDropped`", because we don't want to call it again every time this function runs.)

pkg/network/common/egressip.go
```diff
@@ -59,7 +60,7 @@ type EgressIPWatcher interface {
 
 	SetNamespaceEgressNormal(vnid uint32)
 	SetNamespaceEgressDropped(vnid uint32)
-	SetNamespaceEgressViaEgressIP(vnid uint32, egressIP, nodeIP string)
+	SetNamespaceEgressViaEgressIP(vnid uint32, activeEgressIPs map[string]string)
```
danwinship, line 63: probably "SetNamespaceEgressViaEgressIPs" now (plural)

pkg/network/common/egressip_test.go
```diff
@@ -362,38 +364,38 @@ func TestMultipleNamespaceEgressIPs(t *testing.T) {
...
-	err = w.assertChanges(
-		"namespace 42 via 172.17.0.100 on 172.17.0.3",
-	)
+	err = w.assertNoChanges()
 	if err != nil {
 		t.Fatalf("%v", err)
 	}
 
-	// Removing the inactive egress IP from its node should have no effect
+	// Removing the 1 egress IP from its node should cause it to be removed
```
danwinship, line 391: hm... the new rules are kind of ruining the logic of the unit test... It may be worth going through it again and thinking about if there are any cases we could be testing better now. And if not, then at least some parts of this are basically redundant now and could be dropped.

pkg/network/node/egressip.go
```diff
@@ -151,9 +151,13 @@ func (eip *egressIPWatcher) SetNamespaceEgressDropped(vnid uint32) {
 	}
 }
 
-func (eip *egressIPWatcher) SetNamespaceEgressViaEgressIP(vnid uint32, egressIP, nodeIP string) {
-	mark := eip.iptablesMark[egressIP]
-	if err := eip.oc.SetNamespaceEgressViaEgressIP(vnid, nodeIP, mark); err != nil {
+func (eip *egressIPWatcher) SetNamespaceEgressViaEgressIP(vnid uint32, activeEgressIPs map[string]string) {
+	marks := make(map[string]string)
```
danwinship, line 155: (same comment about map-vs-array-of-struct)

pkg/network/node/egressip_test.go
```diff
@@ -64,10 +65,6 @@ type egressOVSChange struct {
 	remote string
 }
 
-// Takes the previous set of egress OVS flows, then fetches the current set and checks
-// that the expected changes have occurred. Each namespace whose egress has changed should
-// have an egressOVSChange struct describing the expected new state. On success, returns
-// the new/current set of flows in flows.
 func assertOVSChanges(eip *egressIPWatcher, flows *[]string, changes ...egressOVSChange) error {
```
danwinship, line 68: why did you remove the comment?

pkg/network/node/egressip_test.go
```diff
@@ -130,6 +127,54 @@ func assertNoOVSChanges(eip *egressIPWatcher, flows *[]string) error {
 	return assertOVSChanges(eip, flows)
 }
 
+func compareGroups(groups []string, expected map[uint32][]string) bool {
+	returnedGroups := make(map[uint32][]string)
+
+	for _, group := range groups {
+		var groupID uint32
+		// it is difficult to separate the returned groups into a meaningful form
+		splitGroup := strings.SplitN(group, "bucket=", 2)
```
danwinship, line 136: there should probably be something in `pkg/network/node/ovs/parse.go` to parse the group lines, like we have with `ParseFlow`

pkg/network/node/ovs/ovs.go
```diff
@@ -35,6 +35,16 @@ type Interface interface {
 	// error if the interface is not currently a bridge port.)
 	DeletePort(port string) error
 
+	// AddGroup adds a group to the bridge
+	AddGroup(groupID uint32, groupType string, buckets []string) error
```
danwinship, line 39: `AddGroup` and `DeleteGroup` can and should be done as part of `Transaction`

pkg/network/node/ovscontroller.go
danwinship, line 796: If an egress IP is available locally, you should _always_ process it locally.

In particular, the way this is written now, if an egress IP exists on Nodes A and B, then it's possible that Node C might send traffic to Node A for the egress IP, but then Node A will decide to forward it to Node B, and then Node B will forward it back to Node A, and then Node A will emit it locally, etc. The more backup egress IPs you have, the less likely it is the packet will ever actually egress :joy:

pkg/network/common/egressip.go
JacobTanenbaum: done

pkg/network/common/egressip.go
JacobTanenbaum: done

pkg/network/node/egressip.go
JacobTanenbaum: done

pkg/network/node/egressip_test.go
JacobTanenbaum: I thought I would need to make many more changes and have to rewrite the comment but that is not the case

pkg/network/node/ovscontroller.go
JacobTanenbaum: done

pkg/network/common/egressip.go
JacobTanenbaum: done

alexanderConstantinescu (changes requested):

pkg/network/node/ovscontroller.go
```diff
@@ -767,35 +767,52 @@ func (oc *ovsController) ensureTunMAC() error {
...
+			commit := ""
+			if oc.useConnTrack {
+				commit = "ct(commit),"
+			}
+			buildBuckets = append(buildBuckets, fmt.Sprintf("actions=%smove:NXM_NX_REG0[]->NXM_NX_TUN_ID[0..31],set_field:%s->tun_dst,vxlan0", commit, egressIPMetaData.nodeIP))
 		}
-		otx.AddFlow("table=100, priority=100, reg0=%d, ip, actions=set_field:%s->eth_dst,set_field:%s->pkt_mark,goto_table:101", vnid, oc.tunMAC, mark)
+
+	}
+	oc.ovs.AddGroup(vnid, "select", buildBuckets)
```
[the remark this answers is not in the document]
JacobTanenbaum: done

pkg/network/common/egressip.go
```diff
@@ -34,10 +34,16 @@ type nodeEgress struct {
 }
 
 type namespaceEgress struct {
-	vnid         uint32
-	requestedIPs []string
+	vnid             uint32
+	requestedIPs     []string
+	isTrafficDropped bool
```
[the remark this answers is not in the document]
JacobTanenbaum: done

pkg/network/node/egressip_test.go
JacobTanenbaum: done

pkg/network/node/ovs/ovs.go
JacobTanenbaum: done

pkg/network/node/ovs/ovs.go
```diff
@@ -409,8 +423,9 @@ func (ovsif *ovsExec) UpdateOVSMetrics() {
 
 // ovsExecTx implements ovs.Transaction and maintains current flow context
 type ovsExecTx struct {
-	ovsif *ovsExec
-	flows []string
+	ovsif  *ovsExec
+	flows  []string
+	groups []string
```
danwinship, line 428: There should be a single array, because order matters (ie, you can't add a flow that outputs to a group if the group doesn't exist yet). Call it `mods` (the man page talks about "flow mods" and "group mods").

pkg/network/node/ovs/parse.go
```diff
@@ -397,3 +414,131 @@ func UnparseExternalIDs(externalIDs map[string]string) string {
...
+		if end == -1 {
+			end = len(group)
+		}
+		eq := strings.Index(group, "=")
+		if eq == -1 || eq > end {
+			field = group[:end]
+		} else {
+			field = group[:eq]
+			if field == "bucket" {
+				//find the index of the next bucket to determine the end
```
danwinship, line 439:
```suggestion
				// find the index of the next bucket to determine the end
```

pkg/network/node/ovs/parse.go
```diff
@@ -397,3 +414,131 @@ func UnparseExternalIDs(externalIDs map[string]string) string {
...
+				end = strings.Index(group[eq+1:], "bucket")
```
danwinship, line 440: It's confusing to use `end` for this, since everywhere else in the loop, `end` is a length relative to `group`, but here it's a length relative to `group[eq+1:]`. Should use a different temporary variable instead (`nextBucket`?)

pkg/network/node/ovs/parse.go
```diff
@@ -397,3 +414,131 @@ func UnparseExternalIDs(externalIDs map[string]string) string {
...
+			}
+			parsed.GroupID = uint32(converted)
+		case "type":
+			if value == "select" {
+				parsed.Type = SelectGroup
+			} else {
+				return nil, fmt.Errorf("groups of type %s are not supported", value)
+			}
+		case "bucket":
+			//remove actions either hear or above
```
danwinship, line 472:
```suggestion
			// remove actions either here or above
```
I don't understand what this means though...

pkg/network/node/ovs/parse.go
```diff
@@ -397,3 +414,131 @@ func UnparseExternalIDs(externalIDs map[string]string) string {
...
+			actions := strings.SplitAfter(value, "actions=")
+			if len(actions) != 2 {
+				return nil, fmt.Errorf("improperly formated bucket in ovs group %s", value)
```
danwinship, line 475:
```suggestion
				return nil, fmt.Errorf("improperly formatted bucket in ovs group %s", value)
```

(and again below)

pkg/network/node/ovs/parse.go
```diff
@@ -397,3 +414,131 @@ func UnparseExternalIDs(externalIDs map[string]string) string {
...
+
+		for end < len(group) && (group[end] == ',' || group[end] == ' ') {
+			end++
+		}
+		group = group[end:]
+	}
+	return parsed, nil
+}
+
+//GroupMatches checks if two Groups match returns true
```
danwinship, line 493:
```suggestion
// GroupMatches checks if two Groups match. Returns true
```

pkg/network/node/ovscontroller.go
```diff
@@ -769,33 +769,50 @@ func (oc *ovsController) ensureTunMAC() error {
...
+	otx.DeleteGroup(vnid)
+
+	var buildBuckets []string
+	for _, egressIPMetaData := range egressIPsMetaData {
+		if egressIPMetaData.nodeIP == oc.localIP {
+			if err := oc.ensureTunMAC(); err != nil {
+				return err
+			}
+			buildBuckets = []string{fmt.Sprintf("actions=set_field:%s->eth_dst,set_field:%s->pkt_mark,output:tun0", oc.tunMAC, egressIPMetaData.packetMark)}
+			//if one of the egressIPs is the IP of the node always use that egress otherwise it might not actually egress and get sent between nodes
```
danwinship, line 796:
```suggestion
			// if one of the egressIPs is the IP of the node always use that egress otherwise it might not actually egress and get sent between nodes
```

pkg/network/node/ovscontroller.go
```diff
@@ -769,33 +769,50 @@ func (oc *ovsController) ensureTunMAC() error {
...
 	} else {
-		commit := ""
-		if oc.useConnTrack {
-			commit = "ct(commit),"
-		}
-		otx.AddFlow("table=100, priority=100, reg0=%d, ip, actions=%smove:NXM_NX_REG0[]->NXM_NX_TUN_ID[0..31],set_field:%s->tun_dst,output:1", vnid, commit, nodeIP)
+		// there is at least one egressIP hosted by one other node. Use a group
+		// to load balance between the egressIPs
+		otx.AddFlow("table=100, priority=100,ip,reg0=%d, actions=group:%d", vnid, vnid)
+		otx.AddGroup(vnid, "select", buildBuckets)
```
danwinship, line 815: you need to call `AddGroup` before any `AddFlow` that references that group

pkg/network/node/ovscontroller_test.go
```diff
@@ -1138,7 +1138,7 @@ var expectedFlows = []string{
 	" cookie=0, table=100, priority=200, tcp, tcp_dst=53, nw_dst=172.17.0.4, actions=output:2",
 	" cookie=0, table=100, priority=200, udp, udp_dst=53, nw_dst=172.17.0.4, actions=output:2",
 	" cookie=0, table=100, priority=150, ct_state=+rpl, actions=goto_table:101",
-	" cookie=0, table=100, priority=100, reg0=37, ip, actions=ct(commit),move:NXM_NX_REG0[]->NXM_NX_TUN_ID[0..31],set_field:10.0.12.34->tun_dst,output:1",
+	" cookie=0, table=100, priority=100, ip, reg0=37, actions=group:37",
```
danwinship, line 1141: :frowning_face: read the comment before this array
(you need to increment `ruleVersion` by 1)

danwinship (commented):
a few bugs (particularly add-group/add-flow ordering), a bunch of easy nitpicks, and a few comments on things that don't have to be fixed right away

pkg/network/node/ovs/fake_ovs.go
```diff
@@ -124,6 +127,34 @@ func (fake *ovsFake) DeletePort(port string) error {
 	return nil
 }
 
+func (fake *ovsFake) DumpGroups() ([]string, error) {
+	if err := fake.ensureExists(); err != nil {
+		return nil, err
+	}
+
+	var groups []string
+	for _, group := range fake.groups {
```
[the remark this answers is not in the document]
JacobTanenbaum: that is the reason why the compare stuff is more complicated, I check if all the elements are the same regardless of order

pkg/network/common/egressip.go
```diff
@@ -420,43 +426,44 @@ func (eit *EgressIPTracker) syncEgressNodeState(eg *egressIPInfo, active bool) {
...
+		if len(ns.activeEgressIPs) != 0 {
+			ns.activeEgressIPs = []EgressIPAssignment{}
 		}
+		ns.shouldDropTraffic = false
+		eit.watcher.SetNamespaceEgressNormal(ns.vnid)
 		return
 	}
 
-	var active *egressIPInfo
+	activeEgressIPs := []EgressIPAssignment{}
```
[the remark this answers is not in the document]
JacobTanenbaum: done

pkg/network/node/ovs/fake_ovs.go
```diff
@@ -24,12 +25,14 @@ type ovsFake struct {
 	bridge string
 
 	ports map[string]ovsPortInfo
-	flows ovsFlows
+	//map of groupID to OVS group, makes it easier to add and delete groups
```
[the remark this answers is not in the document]
JacobTanenbaum: done

pkg/network/node/ovs/fake_ovs.go
```diff
@@ -297,15 +329,49 @@ func (fake *ovsFake) deleteFlowsHelper(flow string) error {
...
 	}
 
+	for _, group := range tx.groups {
+		origGroup := group
+		if strings.HasPrefix(group, "group add") {
+			group = strings.TrimPrefix(group, "group add ")
+			id := strings.TrimPrefix(strings.Split(group, ",")[0], "group_id=")
+			converted, err := strconv.Atoi(id)
+			if err != nil {
+				return fmt.Errorf("cannot convert %s to int for group add %s: %v", id, origGroup, err)
```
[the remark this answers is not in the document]
JacobTanenbaum: done

pkg/network/node/ovs/ovs.go
JacobTanenbaum: done I think I did this right

pkg/network/node/ovs/parse.go
JacobTanenbaum: done

pkg/network/node/ovs/parse.go
JacobTanenbaum: left a message to myself while I was coding it. removed the message

pkg/network/node/ovs/parse.go
JacobTanenbaum: done

pkg/network/node/ovscontroller.go
JacobTanenbaum: done

pkg/network/node/ovs/parse.go
JacobTanenbaum: done

pkg/network/node/ovscontroller_test.go
JacobTanenbaum: done

pkg/network/node/ovscontroller.go
JacobTanenbaum: done

pkg/network/node/ovs/parse.go
JacobTanenbaum: done

pkg/network/common/egressip.go
```diff
@@ -420,43 +426,44 @@ func (eit *EgressIPTracker) syncEgressNodeState(eg *egressIPInfo, active bool) {
...
 func (eit *EgressIPTracker) syncEgressNamespaceState(ns *namespaceEgress) {
 	if len(ns.requestedIPs) == 0 {
-		if ns.activeEgressIP != "" {
-			ns.activeEgressIP = ""
-			eit.watcher.SetNamespaceEgressNormal(ns.vnid)
+		if len(ns.activeEgressIPs) != 0 {
+			ns.activeEgressIPs = []EgressIPAssignment{}
 		}
+		ns.shouldDropTraffic = false
+		eit.watcher.SetNamespaceEgressNormal(ns.vnid)
```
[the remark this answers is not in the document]
JacobTanenbaum: done

pkg/network/node/ovs/fake_ovs.go
```diff
@@ -297,15 +329,47 @@ func (fake *ovsFake) deleteFlowsHelper(flow string) error {
...
 		return err
 	}
 
+	for _, group := range tx.groups {
+		origGroup := group
+		if strings.HasPrefix(group, "group add") {
+			group = strings.TrimPrefix(group, "group add ")
+			id := strings.TrimPrefix(strings.Split(group, ",")[0], "group_id=")
+			if err != nil {
+				return fmt.Errorf("cannot convert %s to int for group add %s: %v", id, origGroup, err)
```
danwinship, line 352: you removed the `strconv.Atoi` but you left the error check if it failed
(likewise below)

pkg/network/node/ovs/ovs.go
```diff
@@ -410,29 +424,37 @@ func (ovsif *ovsExec) UpdateOVSMetrics() {
...
 }
 
 func (tx *ovsExecTx) Commit() error {
 	defer func() {
-		tx.flows = []string{}
+		tx.mods = []string{}
 	}()
 	return utilwait.ExponentialBackoff(ovsBackoff, func() (bool, error) {
-		err := tx.ovsif.bundle(tx.flows)
+		err := tx.ovsif.bundle(append(tx.mods))
```
danwinship, line 457: don't need the `append` call

pkg/network/node/ovs/fake_ovs.go
JacobTanenbaum: done

pkg/network/node/ovs/ovs.go
JacobTanenbaum: done

PR #280, opened 2021-03-25
Bug 1937194: Routes not accessible from projects with egress IP configred on cloud providers

fgiloux:
This pull request avoids S-NAT done for egress IP when the destination is in the pod network range, which happens when a pod tries to access a service type load balancer, like the router in a cloud environment.
Packets with the egress IP as source otherwise get dropped by OVS.

### DIFF
--- a/pkg/network/node/iptables.go
+++ b/pkg/network/node/iptables.go
@@ -210,6 +210,7 @@ func (n *NodeIPTables) getNodeIPTablesChains() []Chain {
 		{"-m", "mark", "--mark", n.masqueradeBitHex + "/" + n.masqueradeBitHex, "-j", "RETURN"},
 	}
 	var masq2Rules [][]string
+	var masqERules [][]string
 	var filterRules [][]string
 	for _, cidr := range n.clusterNetworkCIDR {
 		if n.masqueradeServices {
@@ -218,6 +219,8 @@ func (n *NodeIPTables) getNodeIPTablesChains() []Chain {
 			masqRules = append(masqRules, []string{"-s", cidr, "-m", "comment", "--comment", "masquerade pod-to-external traffic", "-j", "OPENSHIFT-MASQUERADE-2"})
 			masq2Rules = append(masq2Rules, []string{"-d", cidr, "-m", "comment", "--comment", "masquerade pod-to-external traffic", "-j", "RETURN"})
 		}
+		masqRules = append(masqRules, []string{"-s", cidr, "-m", "comment", "--comment", "egress S-NAT pod-to-external traffic", "-j", "OPENSHIFT-MASQUERADE-EGRESS"})
+		masqERules = append(masqERules, []string{"-d", cidr, "-m", "comment", "--comment", "no egress S-NAT for traffic to pod", "-j", "RETURN"})
 
 		filterRules = append(filterRules, []string{"-s", cidr, "-m", "comment", "--comment", "attempted resend after connection close", "-m", "conntrack", "--ctstate", "INVALID", "-j", "DROP"})
 		filterRules = append(filterRules, []string{"-d", cidr, "-m", "comment", "--comment", "forward traffic from SDN", "-j", "ACCEPT"})
@@ -231,6 +234,12 @@ func (n *NodeIPTables) getNodeIPTablesChains() []Chain {
 			srcChain: "POSTROUTING",
 			srcRule:  []string{"-m", "comment", "--comment", "rules for masquerading OpenShift traffic"},
 			rules:    masqRules,
+		},
+		Chain{
+			table:   "nat",
+			name:    "OPENSHIFT-MASQUERADE-EGRESS",
+			srcRule: []string{"-m", "comment", "--comment", "rules for masquerading OpenShift traffic with EgressIP"},
+			rules:   masqERules,
 		},
 		Chain{
 			table:    "filter",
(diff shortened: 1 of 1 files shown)


pkg/network/node/iptables.go
```diff
@@ -227,7 +227,7 @@ func (n *NodeIPTables) getNodeIPTablesChains() []Chain {
 
 func (n *NodeIPTables) ensureEgressIPRules(egressIP, mark string) error {
 	for _, cidr := range n.clusterNetworkCIDR {
-		_, err := n.ipt.EnsureRule(iptables.Prepend, iptables.TableNAT, iptables.Chain("OPENSHIFT-MASQUERADE"), "-s", cidr, "-m", "mark", "--mark", mark, "-j", "SNAT", "--to-source", egressIP)
+		_, err := n.ipt.EnsureRule(iptables.Prepend, iptables.TableNAT, iptables.Chain("OPENSHIFT-MASQUERADE"), "-s", cidr, "!", "-d", cidr, "-m", "mark", "--mark", mark, "-j", "SNAT", "--to-source", egressIP)
```
alexanderConstantinescu, line 230: Hmm, this does fix the bug but I am not sure it's the proper way to fix it. We shouldn't be sending packets to egress nodes if those packets are east <-> west related (which seems to be the case, given that this bug exists). I think the proper way to fix it would be to add a default OVS flow in `SetupOVS` in table 100 that takes precedence over the egress IP related flows that are later setup, similar to what is done here: https://github.com/openshift/sdn/blob/master/pkg/network/node/ovscontroller.go#L203-L206

As such we'll avoid additional hops to egress node for packets that are not actually "egress-ing" the cluster.

alexanderConstantinescu (changes requested):

pkg/network/node/iptables.go
fgiloux: @alexanderConstantinescu you can find more background in the bugzilla referenced in the title of the PR.

I am not sure what you mean by "egress nodes". Egress IP is implemented by S-NAT in iptables on all the nodes.

First the packets have the load balancer IP as destination, outside of the pod range and as such it is not identified in OVS that it is east <-> west traffic. It is only after the D-NAT is applied in the load balancer implementation that it is clear that it is east <-> west traffic. This is again implemented in iptables not in OVS. This PR is only aiming at fixing the issue so that egress IP plays well with services of type load balancer not to change the implementation strategy of one or the other.

With the proposed fix the packets won't be sent to a different node outside of the VXLAN. I am wondering whether you are not thinking of egress router rather than egress IP. The issue here is with the later.

alexanderConstantinescu: Yes, reading more closely I realize that it's true that on clouds such as Azure, pods will send their traffic to an external VIP when targeting an OpenShift route, which thus means that we can't implement what I said in OVS.   

> Egress IP is implemented by S-NAT in iptables on all the nodes.

No, that is wrong. There is a notion of "nodes selected by the user to be used as egress node for egress IP", this is what I referred to as "egress nodes" - and this is done by the user by specifying `hostsubnet.spec.egressIPs` or `hostsubnet.spec.egressCIDRs` on the node they would like to act as "egress nodes". The iptables rules you've modified are only applied on those nodes, not on all cluster nodes. 

Ideally, traffic for OpenShift routes should not even be routed to these egress nodes, since the endpoint is an internal cluster pod. However, as we've just discussed: that is not possible since the cloud's virtual IP space is not identifiable by openshift-sdn (at least today) and we can't tell OVS to ignore that traffic.

alexanderConstantinescu: That being said I am however thinking the customer might have done something wrong when claiming the egress IP on Azure in their setup. If they've done that right I don't see why Azure would not be able to route the egress IP correctly? However, your patch seems fine for what concerns the added filter in the iptables rules used.

fgiloux: > That being said I am however thinking the customer might have done something wrong when claiming the egress IP on Azure in their setup. If they've done that right I don't see why Azure would not be able to route the egress IP correctly? 

There is more information in the bugzilla but the traffic does not leave the node with the egress IP as far as I have seen. I was also able to reproduce on a newly installed cluster. Here is what I understand from the flow:

PREROUTING

Chain KUBE-SEP-GMF6HD62GMXKGYYH (1 references)
target     prot opt source               destination
KUBE-MARK-MASQ  all  --  10.130.4.33          0.0.0.0/0            /* openshift-ingress/router-default:http */
DNAT       tcp  --  0.0.0.0/0            0.0.0.0/0            /* openshift-ingress/router-default:http */ tcp to:10.130.4.33:80

At this point the destination IP has been changed from the load balancer IP to the IP of a router pod.

POSTROUTING

Chain OPENSHIFT-MASQUERADE (1 references)
target     prot opt source               destination
SNAT       all  --  10.128.0.0/14        0.0.0.0/0            mark match 0x6e52f2 to:10.0.116.8

Masquerading is done on all pods with the egress IP for packets that have a mark matching 0x6e52f2
At this point we have the packets that have the egress IP as source IP and router pod IP as destination IP. This should go to OVS but OVS will not accept traffic from external IP (the egress IP is not the IP of the node) to get directly routed to the pod and the packets get dropped.


> However, your patch seems fine for what concerns the added filter in the iptables rules used.

Good to read. I am looking forward to getting this progressed.

dcbw: @danwinship might have some input here too

danwinship, line 230: This doesn't deal correctly with multiple cluster CIDRs.
eg, if your cluster network is using `10.128.0.0/14` and `10.192.0.0/14`, then this change will make it so that a packet from `10.128.0.0/14` to `10.128.0.0/14` wouldn't get NATted, and a packet from `10.192.0.0/14` to `10.192.0.0/14` wouldn't get NATted, but a packet from `10.128.0.0/14` to `10.192.0.0/14` or vice-versa would

danwinship:
/retitle Bug 1937194: Routes not accessible from projects with egress IP configred on cloud providers

pkg/network/node/iptables.go
danwinship: This also doesn't seem to address the bug; note that the destination in your example is `10.0.116.7`, which is NOT inside `10.128.0.0/14`. Thus, the packet would still get SNATted anyway. (AWS/GCP/Azure clusters by default use `10.[128-131].x.y` for pods and `10.0.x.y` for nodes/LBs.)

fgiloux: >>> This doesn't deal correctly with multiple cluster CIDRs.
True. I completely missed that and will get it fixed.

>>>This also doesn't seem to address the bug; note that the destination in your example is 10.0.116.7, which is NOT inside 10.128.0.0/14. Thus, the packet would still get SNATted anyway. (AWS/GCP/Azure clusters by default use 10.[128-131].x.y for pods and 10.0.x.y for nodes/LBs.)
I agree with 10.0.116.7, not being inside 10.128.0.0/14 but in PREROUTING this gets changed (D-NAT) to the IP of a router pod, one being 10.130.4.33:
Chain KUBE-SERVICES (2 references)
target     prot opt source               destination
KUBE-FW-HEVFQXAKPPGAL4BV  tcp  --  0.0.0.0/0            10.0.116.7           /* openshift-ingress/router-default:http loadbalancer IP */ tcp dpt:80
KUBE-FW-MBAZS3WDHL45BPIZ  tcp  --  0.0.0.0/0            10.0.116.7           /* openshift-ingress/router-default:https loadbalancer IP */ tcp dpt:443
[...]
Chain KUBE-SEP-GMF6HD62GMXKGYYH (1 references)
target     prot opt source               destination
KUBE-MARK-MASQ  all  --  10.130.4.33          0.0.0.0/0            /* openshift-ingress/router-default:http */
DNAT       tcp  --  0.0.0.0/0            0.0.0.0/0            /* openshift-ingress/router-default:http */ tcp to:10.130.4.33:80
Am I missing something?

fgiloux:
It seems that tests are broken. In a new directory I did:
- git clone of openshift/sdn
- make test
and it resulted in a failure

fgiloux:
@danwinship @alexanderConstantinescu now that the tests have been fixed by:
commit 5cdc9f8d7f2c35f8cb632c98803c003ba78aeb15 Author: Alexander Constantinescu <aconstan@redhat.com>
Fix flaky unit test `TestEgressCIDRAllocation`

I am keen on getting this PR progressed. The change is more significant than what I first anticipated as I needed to introduce a new chain to deal with multiple CIDR as "!" is not allowed with multiple source or destination IP addresses. 

I am looking forward to your review & feedback.

fgiloux:
@danwinship @dcbw anything I can do to get that progressed? On the other hand could you please let me know if that's not something you want to get?

PR #284, opened 2021-04-12
[WIP] Set MTU on SDN Overlay CIDR explicitly

mccv1r0:
WIP - 

When adding static route to the SDN Overlay CIDR, set the MTU value so the kernel doesn't rely on interface MTU when determining e.g. TCP MSS and when to IP Fragment UDP messages etc.
    
Signed-off-by: Michael Cambria <mcambria@redhat.com>


cmd/sdn-cni-plugin/openshift-sdn_linux.go
```diff
@@ -234,11 +234,30 @@ func (p *cniPlugin) CmdAdd(args *skel.CmdArgs) error {
...
+				return fmt.Errorf("failed to add mtu based route to dst: %v via SDN: %v", cn.ClusterCIDR, err)
+			}
+		}
+
 		dsts = append(dsts, serviceIPNet)
 		for _, dst := range dsts {
 			route := &netlink.Route{
 				Dst: dst,
 				Gw:  defaultGW,
+				MTU: routeMTU,
```
danwinship, line 260: Some service traffic will end up going to nodes, not pods... will that screw anything up if we get a spurious ICMP in response to that? (I guess probably not since the ICMP would be in response to traffic that had been SNATted to a node, so it presumably wouldn't make it back to the pod anyway...)

pkg/network/node/node.go
```diff
@@ -174,7 +174,7 @@ func New(c *OsdnNodeConfig) (*OsdnNode, error) {
 		recorder:         c.Recorder,
 		oc:               oc,
 		networkInfo:      networkInfo,
-		podManager:       newPodManager(c.KClient, policy, networkInfo.MTU, oc),
+		podManager:       newPodManager(c.KClient, policy, networkInfo.MTU+50, oc),
```
danwinship, line 177: Until now, everywhere the code talked about "MTU" it meant "the VXLAN tunnel MTU", but this changes it so that `podManager.mtu` and `cniserver.Config.MTU` are now the underlay network MTU. I think we need to rename all the "mtu" fields to make it clear which MTU they are talking about.

Also, it would be good to only do the math (+/-50) in one place and then just have separate "tunnel MTU" and "underlay MTU" values for everyone else to use. So like, replace the `MTU` field in `common.ParsedClusterNetwork` with `TunnelMTU` and `UnderlayMTU` (or other names if that would be better), where `pcn.TunnelMTU = *cn.MTU` and `pcn.UnderlayMTU = pcn.TunnelMTU + 50`. You'll have to pass both of them over `cniserver.Config`, but I think that makes things clearer.

cmd/sdn-cni-plugin/openshift-sdn_linux.go
mccv1r0: We should be ok.   This traffic flow is identical to the problem we're trying to solve.  What's different this time is that if the 'bug' occurs, when `ip_forward()` is done after DNAT from node to pod address, the pod `eth0` mtu (and that of `tun0`) will now be that of the underlay (e.g. 1500), so an ICMP will not be generated.

PR #287, opened 2021-04-13
Bug 1948999: Remove check enforcing single egress IP for automatic assignment

alexanderConstantinescu:
The load-balancing feature overlooked removing the check that exists
enforcing that only one egress IP is assigned in automatic assignment
mode. This patch removes that check. The concern with removing this
check is that we might have multiple egress IPs being assigned to the
same node (since any previous implementation didn't have to deal with
that). This is however covered by the verification in `egressIPActive`
which ensures that no more than one egress IP for each namespace is
assigned to every node.

FYI: I have tested this on a local setup with Kind and it works. Both the 
load-balancing feature functions with automatic assignment, and we never
assign more than one egress IP per node for each namespace.

/assign @JacobTanenbaum @danwinship

### DIFF
--- a/pkg/network/common/egressip_test.go
+++ b/pkg/network/common/egressip_test.go
@@ -1068,10 +1068,10 @@ func TestEgressCIDRAllocation(t *testing.T) {
 		t.Fatalf("%v", err)
 	}
 
-	// You can't mix multiple-egress-IP HA with auto-allocated-egress-IP HA
+	// The controller can auto-allocate multiple egress IPs to the same namespace
 	updateNetNamespaceEgress(eit, &networkv1.NetNamespace{
 		NetID:     45,
-		EgressIPs: []networkv1.NetNamespaceEgressIP{"172.17.0.102", "172.17.1.102"},
+		EgressIPs: []networkv1.NetNamespaceEgressIP{"172.17.0.102", "172.17.1.102"}, // 172.17.0.102 is already allocated above
 	})
 	updateNetNamespaceEgress(eit, &networkv1.NetNamespace{
 		NetID:     49,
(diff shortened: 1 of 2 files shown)


pkg/network/common/egressip_test.go
```diff
@@ -1068,10 +1068,10 @@ func TestEgressCIDRAllocation(t *testing.T) {
 		t.Fatalf("%v", err)
 	}
 
-	// You can't mix multiple-egress-IP HA with auto-allocated-egress-IP HA
+	// You can mix multiple-egress-IP HA with auto-allocated-egress-IP HA
```
danwinship, line 1071: You still can't mix the two different modes, it's just that now the auto-allocated mode supports multiple IPs. So the comment should be something like

    // The controller can auto-allocate multiple egress IPs to the same namespace

pkg/network/common/egressip_test.go
```diff
@@ -1068,10 +1068,10 @@ func TestEgressCIDRAllocation(t *testing.T) {
...
 	updateNetNamespaceEgress(eit, &networkv1.NetNamespace{
 		NetID:     45,
-		EgressIPs: []networkv1.NetNamespaceEgressIP{"172.17.0.102", "172.17.1.102"},
+		EgressIPs: []networkv1.NetNamespaceEgressIP{"172.17.0.102", "172.17.1.102"}, // 172.17.0.102 is already allocated above
```
danwinship, line 1074: comment might be more useful down below with the `assertChanges` output?

danwinship (commented):
yeah, I think the code is right, but one comment is wrong

pkg/network/common/egressip_test.go
alexanderConstantinescu: Done

pkg/network/common/egressip_test.go
alexanderConstantinescu: I am not sure it is, I did it like this to align it with all the other comments in the test, see: 
https://github.com/openshift/sdn/blob/b24f70e994471849788104c52a53f47dd8f511ba/pkg/network/common/egressip_test.go#L1034
or 
https://github.com/openshift/sdn/blob/b24f70e994471849788104c52a53f47dd8f511ba/pkg/network/common/egressip_test.go#L1001

PR #288, opened 2021-04-14
Bug 1949607: fix multiple egressIPs for a single namespace

JacobTanenbaum:
Leaving the check there causes openshift-sdn to treat multiple
egressIPs assigned to a single namespace as already allocated


JacobTanenbaum:
closed in favor of https://github.com/openshift/sdn/pull/287

PR #289, opened 2021-04-20
Bug 1879077: ValidateMTU: check only the interface that holds the node ip

fedepaol:
The only interface that needs to be checked is the one tied to the node ip. 
This prevents tainting the node when another interface is added with a lower MTU.

### DIFF
--- a/pkg/network/node/node.go
+++ b/pkg/network/node/node.go
@@ -268,6 +268,19 @@ func (node *OsdnNode) validateMTU() error {
 			return fmt.Errorf("could not retrieve link id %d while validating MTU", route.LinkIndex)
 		}
 
+		// we want to check the mtu only for the interface assigned to the node's primary ip
+		found := false
+		addresses, err := netlink.AddrList(link, netlink.FAMILY_V4)
+		for _, address := range addresses {
+			if node.localIP == address.IP.String() {
+				found = true
+				break
+			}
+		}
+		if !found {
+			continue
+		}
+
 		newmtu := link.Attrs().MTU
 		if newmtu > 0 && newmtu < mtu {
 			mtu = newmtu


pkg/network/node/node.go
```diff
@@ -268,6 +268,18 @@ func (node *OsdnNode) validateMTU() error {
 			return fmt.Errorf("could not retrieve link id %d while validating MTU", route.LinkIndex)
 		}
 
+		// we want to check the mtu only for the interface assigned to the node's primary ip
```
aojea, line 271: in a node with multiple interfaces, if one of them will forward traffic from pods  and it has a wrong mtu , it will cause networking issues, shouldn't all interfaces have the same mtu?

fedepaol: Not according to @dcbw 's comment here https://bugzilla.redhat.com/show_bug.cgi?id=1879077#c14
I think the tradeoff here is allowing extra interfaces not involved in ocp networking to carry a wrong mtu and giving responsibility to the user of making sure that's not involved.

In the customer scenario the added interface was a management one.

aojea: aha , I see 

> > Actually, the MTU-tainting code is buggy; it taints the node if there is _any_ interface with a default route and a too-small MTU, even if it's not the interface that will actually get used.
> 
> It probably ought to only check the interface that holds the primary node IP, since that's guaranteed to be the interface that at least inbound VXLAN traffic will use. If someone has a cluster with zany asymmetric routing such that outbound VXLAN uses a different interface, then they can just be responsible for sanity-checking the MTU on that interface themselves.

pkg/network/node/node.go
```diff
@@ -268,6 +268,18 @@ func (node *OsdnNode) validateMTU() error {
...
+		found := false
+		addresses, err := netlink.AddrList(link, netlink.FAMILY_V4)
+		for _, address := range addresses {
+			if node.localIP == address.IP.String() {
+				found = true
```
squeed, line 276: nit: `break` here too.

fedepaol: Done, thanks!

PR #291, opened 2021-04-23
generalize the polling of `monitorNodes` between node and master

alexanderConstantinescu:
This PR fixes two issues. The first commit is a problem introduced by the new load balancing feature. The second commit fixes an issue that has always been existing for manual assignment of egress IPs with opensift-sdn, but which we can/might want to fix now given the first commit.

I am not certain the second commit is needed, we could just argue that it's never really worked so no need to fix it now when this plugin is reaching a deprecation phase. I added it however as I'd like to get some input on if people think it could be useful.

/assign @danwinship @JacobTanenbaum

### DIFF
--- a/pkg/network/common/egressip.go
+++ b/pkg/network/common/egressip.go
@@ -21,9 +21,9 @@ import (
 	networkinformers "github.com/openshift/client-go/network/informers/externalversions/network/v1"
 )
 
-type nodeEgress struct {
-	nodeName string
-	nodeIP   string
+type NodeEgress struct {
+	NodeName string
+	NodeIP   string
 	sdnIP    string
 
 	requestedIPs   sets.String
@@ -50,7 +50,7 @@ type egressIPInfo struct {
 	ip     string
 	parsed net.IP
 
-	nodes      []*nodeEgress
+	nodes      []*NodeEgress
 	namespaces []*namespaceEgress
 
 	assignedNodeIP string
@@ -75,8 +75,8 @@ type EgressIPTracker struct {
 
 	watcher EgressIPWatcher
 
-	nodes            map[ktypes.UID]*nodeEgress
-	nodesByNodeIP    map[string]*nodeEgress
+	nodes            map[ktypes.UID]*NodeEgress
+	nodesByNodeIP    map[string]*NodeEgress
 	namespacesByVNID map[uint32]*namespaceEgress
 	egressIPs        map[string]*egressIPInfo
 	nodesWithCIDRs   int
@@ -90,8 +90,8 @@ func NewEgressIPTracker(watcher EgressIPWatcher) *EgressIPTracker {
 	return &EgressIPTracker{
 		watcher: watcher,
 
-		nodes:            make(map[ktypes.UID]*nodeEgress),
-		nodesByNodeIP:    make(map[string]*nodeEgress),
+		nodes:            make(map[ktypes.UID]*NodeEgress),
+		nodesByNodeIP:    make(map[string]*NodeEgress),
 		namespacesByVNID: make(map[uint32]*namespaceEgress),
 		egressIPs:        make(map[string]*egressIPInfo),
 
@@ -132,14 +132,14 @@ func (eit *EgressIPTracker) egressIPChanged(eg *egressIPInfo) {
 	}
 }
 
-func (eit *EgressIPTracker) addNodeEgressIP(node *nodeEgress, egressIP string) {
+func (eit *EgressIPTracker) addNodeEgressIP(node *NodeEgress, egressIP string) {
 	eg := eit.ensureEgressIPInfo(egressIP)
 	eg.nodes = append(eg.nodes, node)
 
 	eit.egressIPChanged(eg)
 }
 
-func (eit *EgressIPTracker) deleteNodeEgressIP(node *nodeEgress, egressIP string) {
+func (eit *EgressIPTracker) deleteNodeEgressIP(node *NodeEgress, egressIP string) {
 	eg := eit.egressIPs[egressIP]
 	if eg == nil {
 		return
@@ -221,9 +221,9 @@ func (eit *EgressIPTracker) UpdateHostSubnetEgress(hs *networkv1.HostSubnet) {
 		if len(hs.EgressIPs) == 0 && len(hs.EgressCIDRs) == 0 {
 			return
 		}
-		node = &nodeEgress{
-			nodeName:     hs.Host,
-			nodeIP:       hs.HostIP,
+		node = &NodeEgress{
+			NodeName:     hs.Host,
+			NodeIP:       hs.HostIP,
 			sdnIP:        sdnIP,
 			requestedIPs: sets.NewString(),
 		}
@@ -231,7 +231,7 @@ func (eit *EgressIPTracker) UpdateHostSubnetEgress(hs *networkv1.HostSubnet) {
 		eit.nodesByNodeIP[hs.HostIP] = node
 	} else if len(hs.EgressIPs) == 0 && len(hs.EgressCIDRs) == 0 {
 		delete(eit.nodes, hs.UID)
-		delete(eit.nodesByNodeIP, node.nodeIP)
+		delete(eit.nodesByNodeIP, node.NodeIP)
 	}
 
 	// Process EgressCIDRs
@@ -255,22 +255,22 @@ func (eit *EgressIPTracker) UpdateHostSubnetEgress(hs *networkv1.HostSubnet) {
 		eit.updateEgressCIDRs = true
 	}
 
-	if node.nodeIP != hs.HostIP {
+	if node.NodeIP != hs.HostIP {
 		// We have to clean up the old egress IP mappings and call syncEgressIPs
-		// before we can change node.nodeIP
+		// before we can change node.NodeIP
 		movedEgressIPs := make([]string, 0, node.requestedIPs.Len())
 		for _, ip := range node.requestedIPs.UnsortedList() {
 			eg := eit.egressIPs[ip]
-			if eg != nil && eg.assignedNodeIP == node.nodeIP {
+			if eg != nil && eg.assignedNodeIP == node.NodeIP {
 				movedEgressIPs = append(movedEgressIPs, ip)
 				eit.deleteNodeEgressIP(node, ip)
 			}
 		}
 		eit.syncEgressIPs()
 
-		delete(eit.nodesByNodeIP, node.nodeIP)
-		node.nodeIP = hs.HostIP
-		eit.nodesByNodeIP[node.nodeIP] = node
+		delete(eit.nodesByNodeIP, node.NodeIP)
+		node.NodeIP = hs.HostIP
+		eit.nodesByNodeIP[node.NodeIP] = node
 
 		for _, ip := range movedEgressIPs {
 			eit.addNodeEgressIP(node, ip)
@@ -361,7 +361,7 @@ func (eit *EgressIPTracker) egressIPActive(eg *egressIPInfo) (bool, error) {
 		return false, nil
 	}
 	if len(eg.nodes) > 1 {
-		return false, fmt.Errorf("Multiple nodes (%s, %s) claiming EgressIP %s", eg.nodes[0].nodeIP, eg.nodes[1].nodeIP, eg.ip)
+		return false, fmt.Errorf("Multiple nodes (%s, %s) claiming EgressIP %s", eg.nodes[0].NodeIP, eg.nodes[1].NodeIP, eg.ip)
 	}
 	if len(eg.namespaces) > 1 {
 		return false, fmt.Errorf("Multiple namespaces (%d, %d) claiming EgressIP %s", eg.namespaces[0].vnid, eg.namespaces[1].vnid, eg.ip)
@@ -369,7 +369,7 @@ func (eit *EgressIPTracker) egressIPActive(eg *egressIPInfo) (bool, error) {
 	for _, ip := range eg.namespaces[0].requestedIPs {
 		eg2 := eit.egressIPs[ip]
 		if eg2 != nil && eg2 != eg && len(eg2.nodes) == 1 && eg2.nodes[0] == eg.nodes[0] {
-			return false, fmt.Errorf("Multiple EgressIPs (%s, %s) for VNID %d on node %s", eg.ip, eg2.ip, eg.namespaces[0].vnid, eg.nodes[0].nodeIP)
+			return false, fmt.Errorf("Multiple EgressIPs (%s, %s) for VNID %d on node %s", eg.ip, eg2.ip, eg.namespaces[0].vnid, eg.nodes[0].NodeIP)
 		}
 	}
 	return true, nil
@@ -409,9 +409,9 @@ func (eit *EgressIPTracker) syncEgressIPs() {
 }
 
 func (eit *EgressIPTracker) syncEgressNodeState(eg *egressIPInfo, active bool) {
-	if active && eg.assignedNodeIP != eg.nodes[0].nodeIP {
-		klog.V(4).Infof("Assigning egress IP %s to node %s", eg.ip, eg.nodes[0].nodeIP)
-		eg.assignedNodeIP = eg.nodes[0].nodeIP
+	if active && eg.assignedNodeIP != eg.nodes[0].NodeIP {
+		klog.V(4).Infof("Assigning egress IP %s to node %s", eg.ip, eg.nodes[0].NodeIP)
+		eg.assignedNodeIP = eg.nodes[0].NodeIP
 		eit.watcher.ClaimEgressIP(eg.namespaces[0].vnid, eg.ip, eg.assignedNodeIP)
 	} else if !active && eg.assignedNodeIP != "" {
 		klog.V(4).Infof("Removing egress IP %s from node %s", eg.ip, eg.assignedNodeIP)
@@ -529,9 +529,9 @@ func (eit *EgressIPTracker) Ping(ip string, timeout time.Duration) bool {
 	return true
 }
 
-func (eit *EgressIPTracker) nodeHasEgressIPForNamespace(node *nodeEgress, eip *egressIPInfo, allocation map[string][]string) bool {
+func (eit *EgressIPTracker) nodeHasEgressIPForNamespace(node *NodeEgress, eip *egressIPInfo, allocation map[string][]string) bool {
 	if namespace, ok := eit.namespacesByVNID[eip.assignedVNID]; ok {
-		if sets.NewString(allocation[node.nodeName]...).HasAny(namespace.requestedIPs...) {
+		if sets.NewString(allocation[node.NodeName]...).HasAny(namespace.requestedIPs...) {
 			return true
 		}
 	}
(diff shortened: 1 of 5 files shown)


pkg/network/node/node.go
```diff
@@ -369,7 +369,9 @@ func (node *OsdnNode) Start() error {
 		if err := node.SetupEgressNetworkPolicy(); err != nil {
 			return err
 		}
-		if err := node.egressIP.Start(node.networkInformers, node.nodeIPTables); err != nil {
+		nodeinformer := node.kubeInformers.Core().V1().Nodes()
+		nodeinformer.Informer().GetController()
```
danwinship, line 373: what?
(Docs for "GetController": "`// GetController is deprecated, it does nothing useful`)

also, maybe better to just pass `node.kubeInformers` to `Start` and let it pick out the specific informers it needs, rather than having node.go know

pkg/network/node/vxlan_monitor_test.go
```diff
@@ -1,315 +0,0 @@
...
-func peekUpdate(updates chan struct{}, evm *egressVXLANMonitor) []egressVXLANNode {
-	select {
-	case <-updates:
-		return evm.GetUpdates()
-	default:
-		return nil
-	}
-}
-
-func TestEgressVXLANMonitor(t *testing.T) {
```
danwinship, line 42: seems like probably we need more testing in `egressip_test.go` to make up for the lost testing in `vxlan_monitor_test.go`

pkg/network/common/egressip.go
```diff
@@ -290,6 +303,86 @@ func (eit *EgressIPTracker) UpdateHostSubnetEgress(hs *networkv1.HostSubnet) {
...
+}
+
+const (
+	DefaultPollInterval = 5 * time.Second
+	repollInterval      = time.Second
+	maxRetries          = 2
+)
+
+func (eit *EgressIPTracker) Poll() (bool, error) {
+	klog.Infof("Starting poll")
```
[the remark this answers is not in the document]
alexanderConstantinescu: Yeah, indeed something I forgot to clean up before submitting the patch. I'll remove these.

pkg/network/node/node.go
alexanderConstantinescu: This is also something left over from my testing, I had some hick-ups while writing this and saw this comment: https://github.com/openshift/sdn/blob/46fd1cd3346980b0acc869e3dfdc8e6c816f7de2/pkg/network/master/master.go#L79-L84 thinking it might solve my problem. 

I'll remove that.

pkg/network/master/egressip.go
```diff
@@ -121,11 +108,11 @@ func (eim *egressIPManager) maybeDoUpdateEgressCIDRs() (bool, error) {
 		}
 	}
 
-	eim.monitorNodes = monitorNodes
+	eim.tracker.MonitorNodes = monitorNodes
```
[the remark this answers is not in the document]
alexanderConstantinescu: I agree here, but it's not that evident. Master and node have two different set of rules for how they wish to monitor nodes. Nodes only decide to monitor when the egress IP is not assigned to the specific node in question, and master only monitors when `egressCIDRs` are set (and subsequently stops monitoring when the assigned egress IPs, calculated from `egressCIDRs`, is empty). I am not sure we can have `common` manage that by itself, and even if it would be possible, it might require more refactor than we're willing to do. What I can do to avoid master and node tampering with common's data, is: define some public methods (`SetMonitorNodes` for master, `AddEgressIP`/`RemoveEgressIP` for node) which would do what master and node already do, but inside common instead.

pkg/network/master/egressip.go
```diff
@@ -121,11 +108,11 @@ func (eim *egressIPManager) maybeDoUpdateEgressCIDRs() (bool, error) {
...
 	if len(monitorNodes) > 0 {
 		if eim.stop == nil {
 			eim.stop = make(chan struct{})
-			go eim.poll(eim.stop)
+			go utilwait.PollUntil(common.DefaultPollInterval, eim.tracker.Poll, eim.stop)
```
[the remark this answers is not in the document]
alexanderConstantinescu: Here I am using the poll API that node (vxlan_monitor) defined. I thought it was cleaner utilizing the `utilwait` functionality as much as possible. I guess you're saying that `eim.poll` had a tighter retry loop than `evm.poll` did (0 sec vs. 1 sec), and that we should use that instead?

pkg/network/node/vxlan_monitor_test.go
alexanderConstantinescu: I'll try to write something up. Although, `vxlan_monitor_test.go` only tested the logic that founded itself on parsing OVS stats. There were no tests for what concerns polling nodes and mocking connectivity issues.

pkg/network/master/egressip.go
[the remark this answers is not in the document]
danwinship: no, I mean, why make the caller do the `utilwait.PollUtil` (and have to know the right `DefaultPollInterval` to use) rather than just having `eit.Poll` do that itself

pkg/network/node/vxlan_monitor_test.go
danwinship: ok, if we haven't lost any non-ovs-parsing testing then we don't need anything more

pkg/network/node/node.go
danwinship: lol. yay crufty code

pkg/network/master/egressip.go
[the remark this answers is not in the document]
alexanderConstantinescu: OK, that concern is handled by the public methods I described here: https://github.com/openshift/sdn/pull/291#discussion_r620981098 

Given the existence of a a public `AddEgressIP` function which `EgressIPTracker` would expose, we can mask `DefaultPollInterval` from the caller and keep it internal to `common`. I will re-push soon with a version of this.

alexanderConstantinescu:
> Sharing the code between node/ and master/ is definitely the right move. I just reviewed the two commits together. You should squash them.

Squashed. Please take another look when you get a chance.

/bugzilla cc-qa

@huiran0826 : could you verify both bugs listed in the commit and verify that this solves them? I have tested this on a local setup on my computer, but would like your input

alexanderConstantinescu:
> @huiran0826 : could you verify both bugs listed in the commit and verify that this solves them? I have tested this on a local setup on my computer, but would like your input

Thank you for the help! As per comment: https://bugzilla.redhat.com/show_bug.cgi?id=1948436#c1 and https://bugzilla.redhat.com/show_bug.cgi?id=1948969#c1 this patch fixes both bugs.

pkg/network/common/egressip.go
```diff
@@ -90,24 +97,27 @@ func NewEgressIPTracker(watcher EgressIPWatcher) *EgressIPTracker {
 	return &EgressIPTracker{
 		watcher: watcher,
 
-		nodes:            make(map[ktypes.UID]*nodeEgress),
-		nodesByNodeIP:    make(map[string]*nodeEgress),
+		nodes:            make(map[ktypes.UID]*NodeEgress),
+		nodesByNodeIP:    make(map[string]*NodeEgress),
```
danwinship, line 101: (It would be nice to have a separate commit for renaming/exporting NodeEgress, so that all the parts that are just capitalization are separate from the stuff that actually changes functionality)

pkg/network/common/egressip.go
```diff
@@ -290,6 +300,130 @@ func (eit *EgressIPTracker) UpdateHostSubnetEgress(hs *networkv1.HostSubnet) {
...
+func nodeIsReady(node *v1.Node) bool {
+	nodeReady := true
+	for _, cond := range node.Status.Conditions {
+		if cond.Type == v1.NodeReady {
+			if cond.Status == v1.ConditionFalse || cond.Status == v1.ConditionUnknown {
+				nodeReady = false
+			}
+		}
+	}
+	return nodeReady
```
danwinship, line 312: this function doesn't actually need `nodeReady`. You can just `return false` if you find a non-ready condition, and `return true` at the end if you don't

pkg/network/common/egressip.go
```diff
@@ -290,6 +300,130 @@ func (eit *EgressIPTracker) UpdateHostSubnetEgress(hs *networkv1.HostSubnet) {
...
+			eit.stop = make(chan struct{})
+			go utilwait.PollUntil(DefaultPollInterval, eit.poll, eit.stop)
+		}
+	} else {
+		if eit.stop != nil {
+			close(eit.stop)
+			eit.stop = nil
+		}
+	}
+}
```
danwinship, line 328: This really feels like it needs a mutex... or if not, there need to be comments explaining why it doesn't...

(applies to all of the newly-public API...)

pkg/network/common/egressip.go
```diff
@@ -290,6 +300,130 @@ func (eit *EgressIPTracker) UpdateHostSubnetEgress(hs *networkv1.HostSubnet) {
...
+		delete(eit.monitorNodes, nodeIP)
+		if len(eit.monitorNodes) == 0 && eit.stop != nil {
+			close(eit.stop)
+			eit.stop = nil
+		}
+	}
+}
+
+const (
+	DefaultPollInterval = 5 * time.Second
```
danwinship, line 364: no longer needs to be exported, so `defaultPollInterval`

pkg/network/common/egressip.go
alexanderConstantinescu: I agree, I will change that.

pkg/network/common/egressip.go
alexanderConstantinescu: It does not need a mutex, the lock on `EgressIPTracker` will already be held once this function is accessed since this function is called from `syncEgressNodeState`, which in turn is called from `syncEgressIPs` which is always called with the `EgressIPTracker` lock held. I will however add a comment mentioning this, and mention that we always expect the caller to hold the lock

alexanderConstantinescu: Oh, sorry, it wasn't clear from the Github review which function your comment was referencing. I now see your comment mainly referenced `SetMonitorNodes`. This function is only called from `maybeDoUpdateEgressCIDRs` (sdn-controller) which already has its own lock held: `egressIPManager` lock. We can thus assume that all sdn-controller go-routines will access this function safely. The only concern I had when going over this in my mind was w.r.t `check`, which does not hold the `EgressIPTracker` lock, it however ends up calling `SetNodeOffline` which does...but I think it might be safer if I move the locking from `SetNodeOffline` to `check`

alexanderConstantinescu: Circling back on my latest comment, moving the locking from `SetNodeOffline` to `check` will obviously mean that we lock much more often (every 5 seconds) with a possibility of locking for an extended period of time (depending on how many nodes we need to ping and the time it takes to process the answer). I'm not sure we want to do that? Maybe this is the reason we never did so to begin with?

danwinship: > Maybe this is the reason we never did so to begin with?

If the code doesn't hold a lock now then either (a) the current design of the code ensures that it doesn't need to hold a lock there, or (b) it's buggy and we just haven't gotten bitten yet. If the new design of the code doesn't ensure that it doesn't need to hold a lock there (eg, because now it depends on what the manager is doing, which is outside of the tracker's control), then we need to hold a lock, or else rearrange the code.

If `SetMonitorNodes` is being called in one thread by the manager and `check` is iterating `monitorNodes` in another thread, then we need a lock. One possibility would be to have `check` call a `getMonitorNodes` function that holds the lock and then gets the list of nodes then drops the lock. You'd need to make sure the handling of the individual EgressNodes was thread-safe after that.

Actually... it seems like `SetMonitorNodes` is currently wrong; it needs to merge the caller-provided node data in with the current data, to ensure that the private state for each node is kept

pkg/network/common/egressip.go
alexanderConstantinescu: Done

danwinship:
er, I meant have a commit renaming things *first*, then have a commit changing the functionality. So that everything still compiles after both commits.

Also, please squash the fixes. It's really easier to review that way

pkg/network/common/egressip.go
```diff
@@ -290,6 +300,153 @@ func (eit *EgressIPTracker) UpdateHostSubnetEgress(hs *networkv1.HostSubnet) {
...
+	return true
+}
+
+// SetMonitorNodes updates EgressIPTracker with the node data provided and
+// starts a go-routine (if one is not already running) which polls monitorNodes
+// to check if they are ready and reachable
+func (eit *EgressIPTracker) SetMonitorNodes(monitorNodes map[string]*NodeEgress) {
+	eit.Lock()
+	defer eit.Unlock()
+	eit.monitorNodes = monitorNodes
```
danwinship, line 320: This still has the problem I was talking about before with losing information. Every time a `NetNamespace` or `HostSubnet` gets updated and the master calls `SetMonitorNodes`, it will cause the tracker to forget the `sdnIP`, `requestedIPs`, `requestedCIDRs`, `parsedCIDRs`, `activeEgressIPs`, `offline`, and `retries` of all the nodes it had been tracking before.

You have to do a more complicated sort of "sync" operation here, adding elements to and removing elements from `eit.monitorNodes` so that at the end they have the same keys, but whenever there's overlap, it needs to keep the original copy from `eit.monitorNodes`, not replace it with the one from `monitorNodes`

pkg/network/common/egressip.go
```diff
@@ -290,6 +300,153 @@ func (eit *EgressIPTracker) UpdateHostSubnetEgress(hs *networkv1.HostSubnet) {
...
+// GetMonitorNodes gets the node data from EgressIPTracker
+func (eit *EgressIPTracker) GetMonitorNodes() map[string]*NodeEgress {
+	eit.Lock()
+	defer eit.Unlock()
+	return eit.monitorNodes
+}
+
+// AddEgressIP assumes the EgressIPTracker lock is already held. The access to
+// this function call is currently percolated down from `syncEgressIPs`, which
+// always has the lock held. It starts a go-routine (if one is not already
```
danwinship, line 343: Ugh. This is pretty sketchy; an exported API shouldn't be making assumptions about its caller like that...
(I agree that the current state of the code is _correct_; it's just not very good API design on our part.)
Maybe leave a FIXME comment here pointing out that this could be made better?

danwinship (commented):
Blah. This is more complicated than it had seemed originally. Maybe we should go back to your original simpler not-merging-the-master-and-node-together fix for now?

pkg/network/common/egressip.go
alexanderConstantinescu: > This still has the problem I was talking about before with losing information.

No, I don't think what you're describing is correct. In the most recent, and current, version of this patch I added: https://github.com/openshift/sdn/blob/0551a495b59c87b6d5da5144a76e6ba2da4a717f/pkg/network/master/egressip.go#L90-L92 which means that `maybeDoUpdateEgressCIDRs` only does `SetMonitorNodes` for nodes not currently monitored. Since the access here: https://github.com/openshift/sdn/blob/0551a495b59c87b6d5da5144a76e6ba2da4a717f/pkg/network/master/egressip.go#L82  is thread-safe, there shouldn't be any problem w.r.t overriding data that's currently existing in `monitorNodes` 

I need will test what I am saying right now and confirm that this works.

pkg/network/common/egressip.go
alexanderConstantinescu: I will add a FIXME

pkg/network/common/egressip.go
alexanderConstantinescu: Actually, I think I see what you mean now. Indeed if any item in `monitorNodes` happens to be modified in between `GetMonitorNodes` and `SetMonitorNodes` in `maybeDoUpdateEgressCIDRs`, we will override the modified field with the old one. If we however instead do 

```go
if _, exists := monitorNodes[hs.HostIP]; !exists {  	
    eim.tracker.AddToMonitorNodes(&amp;common.NodeEgress{NodeIP: hs.HostIP, NodeName: nodeName})
} 
```
and have `AddToMonitorNodes` be thread-safe while removing `SetMonitorNodes`, we should be fine, right?

alexanderConstantinescu: Doing all of this I realized that `maybeDoUpdateEgressCIDRs` was inherently wrong. There was never any situation in which we would stop monitoring nodes (even if nodes are not marked for egress IP anymore). I've fixed this now, by implementing something that resembles the current implementation.

alexanderConstantinescu:
> Blah. This is more complicated than it had seemed originally. Maybe we should go back to your original simpler not-merging-the-master-and-node-together fix for now?

Honestly, I can't even tell anymore what the old implementation was exactly...Essentially, the code now might look a bit off for what concerns the `EgressIPTracker` API, but it works. I would rather get this PR in as it is now, fix the problem at hand (which is blocking 4.8) and figure out a way to clean this up in the future.

alexanderConstantinescu:
Putting a "hold" on this until after code freeze, giving us time to think of a better way to do this.

/retitle  generalize the polling of `monitorNodes` between node and master

pkg/network/common/egressip.go
```diff
@@ -290,6 +300,174 @@ func (eit *EgressIPTracker) UpdateHostSubnetEgress(hs *networkv1.HostSubnet) {
...
+		activeEgressIPs: sets.NewString(egressIP),
+		NodeName:        nodeName,
+	})
+	if eit.getMonitorNodesLen() == 1 && eit.stop == nil {
+		eit.stop = make(chan struct{})
+		go utilwait.PollUntil(defaultPollInterval, eit.poll, eit.stop)
+	}
+}
+
+func (eit *EgressIPTracker) getMonitorNodesLen() int {
```
danwinship, line 369: The fact that you have to implement this by hand (in this way) suggests that a `sync.Map` isn't really the data structure we want here. Why did you change this from a mutex+map again? I don't see that in the comments.

pkg/network/common/egressip.go
```diff
@@ -290,6 +300,174 @@ func (eit *EgressIPTracker) UpdateHostSubnetEgress(hs *networkv1.HostSubnet) {
...
+			}
+		}
+	}
+	return true
+}
+
+// SetMonitorNodes updates EgressIPTracker with the node data provided and
+// starts a go-routine (if one is not already running) which polls monitorNodes
+// to check if they are ready and reachable.
+func (eit *EgressIPTracker) SetMonitorNodes(monitorNodes []*NodeEgress) {
```
danwinship, line 317: So this fixes the problem of "not overwriting the tracker's data with blank objects from the caller" now, but it makes the whole thing kind of weird... We mostly aren't actually using the objects passed in from the caller, so we didn't really need to export NodeEgress as a type; it would have made more sense to just use a different type here that's just `struct { Name string; IP string }`, or just use a `map[string]string`...

danwinship (commented):
This is no longer fixing any bug, right? It's just trying to consolidate code between the master and node implementations? Maybe we should just not bother doing that...

alexanderConstantinescu:
> This is no longer fixing any bug, right? It's just trying to consolidate code between the master and node implementations? Maybe we should just not bother doing that...

It does fix bug: https://bugzilla.redhat.com/show_bug.cgi?id=1948969 (excuse me, I must have un-linked the PR and Bugzilla at some point). We could however close the BZ as it's more of usability problem than anything else. Essentially: sdn health checks egress node by checking if they are reachable (can be curled) and sdn-controller also checks if the kubelet is reporting a `Ready` state for the egress nodes. QE was expecting this to be the case for sdn too, hence this PR aligning the behavior.

alexanderConstantinescu:
[Discussed off-band] Closing this PR, reasons are:

- Not a critical issue (if there is a customer case in the future, we could re-open it)
- Underestimated complexity of implementing it
- Not enough time to re-work the PR and fix the underestimated complexity

/close

PR #292, opened 2021-04-24
no masquerade intra cluster traffic

aojea:
Observed in https://bugzilla.redhat.com/show_bug.cgi?id=1953032

Pods traffic to nodes should not be masquerade, however, this creates an asymmetric routing problem ... should traffic directed to nodeIPs directed through the tunnel interface?

xref: https://kubernetes.io/docs/tasks/administer-cluster/ip-masq-agent/

Signed-off-by: Antonio Ojea <antonio.ojea.garcia@gmail.com>

### DIFF
--- a/pkg/network/common/common.go
+++ b/pkg/network/common/common.go
@@ -33,6 +33,7 @@ func ClusterNetworkListContains(clusterNetworks []ParsedClusterNetworkEntry, ipa
 }
 
 type ParsedClusterNetwork struct {
+	MachineNetwork  *net.IPNet
 	PluginName      string
 	ClusterNetworks []ParsedClusterNetworkEntry
 	ServiceNetwork  *net.IPNet
@@ -71,6 +72,15 @@ func ParseClusterNetwork(cn *networkv1.ClusterNetwork) (*ParsedClusterNetwork, e
 			return nil, fmt.Errorf("failed to parse ServiceNetwork CIDR %s: %v", cn.ServiceNetwork, err)
 		}
 		utilruntime.HandleError(fmt.Errorf("Configured serviceNetworkCIDR value %q is invalid; treating it as %q", cn.ServiceNetwork, pcn.ServiceNetwork.String()))
+	}
+
+	pcn.MachineNetwork, err = networkutils.ParseCIDRMask(cn.Network)
+	if err != nil {
+		_, pcn.MachineNetwork, err = net.ParseCIDR(cn.Network)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse Machine Network CIDR %s: %v", cn.Network, err)
+		}
+		utilruntime.HandleError(fmt.Errorf("Configured MachineNetworkCIDR value %q is invalid; treating it as %q", cn.Network, pcn.MachineNetwork.String()))
 	}
 
 	if cn.VXLANPort != nil {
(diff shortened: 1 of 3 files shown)


pkg/network/common/common.go
```diff
@@ -73,6 +74,15 @@ func ParseClusterNetwork(cn *networkv1.ClusterNetwork) (*ParsedClusterNetwork, e
 		utilruntime.HandleError(fmt.Errorf("Configured serviceNetworkCIDR value %q is invalid; treating it as %q", cn.ServiceNetwork, pcn.ServiceNetwork.String()))
 	}
 
+	pcn.MachineNetwork, err = networkutils.ParseCIDRMask(cn.Network)
```
danwinship, line 77: `cn.Network` isn't the machine network, it's the deprecated single-value form of `cn.ClusterNetworks`.
There is no CIDR in an OCP cluster that reliably includes the nodes and nothing else.

aojea: I didn't want to bother you on weekends but since you are here, I think that will be easier to send pod traffic through the tunnel and avoid the asymetric traffic 

Pods traffic to nodes should not be masquerade, however, this creates an asymmetric routing problem ... should traffic directed to nodeIPs directed through the tunnel interface?

danwinship: > Pods traffic to nodes should not be masquerade

Says who?

aojea: https://kubernetes.io/docs/concepts/cluster-administration/networking/

> Kubernetes imposes the following fundamental requirements on any networking implementation (barring any intentional network segmentation policies):
> 
> pods on a node can communicate with all pods on all nodes without NAT
> agents on a node (e.g. system daemons, kubelet) can communicate with all pods on that node
> Note: For those platforms that support Pods running in the host network (e.g. Linux):
> 
> pods in the host network of a node can communicate with all pods on all nodes without NAT

` pods in the host network of a node can communicate with all pods on all nodes without NAT`

danwinship: > ` pods in the host network of a node can communicate with all pods on all nodes without NAT`

That doesn't say that the reverse has to be true though.

OTOH

> pods on a node can communicate with all pods on all nodes without NAT

If you assume that the second "pods" includes host-network pods, then we'd be wrong.

But if you assume that the second "pods" includes host-network pods, then you should presumably also assume that the first "pods" includes host-network pods, in which case there is no need for the "pods in the host network ..." statement because it would be redundant. So that implies that "pods" alone is supposed to be understood to mean "pod-network pods".

Or maybe there's a difference between "pods" and "all pods"? ...

We should probably get this text clarified, and then make sure that there are conformance tests ensuring all of the actually-required behaviors.

danwinship: At any rate, if we want to cause pod-to-node traffic to be not-masqueraded, then that has to be done with per-node-ip rules, not with a CIDR match. The easiest way to do this would be in ovscontroller.go rather than iptables.go; have `AddHostSubnetRules` add VXLAN rules for `nw_dst=${subnet.HostIP}` too, not just `nw_dst=${subnet.Subnet}`

aojea:
I don't think we can make this call at this time, also, this doesn't seem to have any negative effect, sdn is not the only one doing this

PR #294, opened 2021-04-28
Bug 1928851: prevent manually creating netns with existing NetID

msherif1234:
Signed-off-by: Mohamed Mahmoud <mmahmoud@redhat.com>
check if newly added netns reuse and existing NetID for a different netns, print a warning message and skip creating this netns

### DIFF
--- a/pkg/network/node/vnids.go
+++ b/pkg/network/node/vnids.go
@@ -136,6 +136,16 @@ func (vmap *nodeVNIDMap) getVNID(name string) (uint32, error) {
 	return 0, fmt.Errorf("failed to find netid for namespace: %s in vnid map", name)
 }
 
+func (vmap *nodeVNIDMap) findDuplicateNetID(namespace string, netID uint32) (string, bool) {
+	names := vmap.GetNamespaces(netID)
+	for _, name := range names {
+		if name != namespace {
+			return name, true
+		}
+	}
+	return "", false
+}
+
 func (vmap *nodeVNIDMap) setVNID(name string, id uint32, mcEnabled bool) {
 	vmap.lock.Lock()
 	defer vmap.lock.Unlock()
@@ -204,6 +214,13 @@ func (vmap *nodeVNIDMap) handleAddOrUpdateNetNamespace(obj, _ interface{}, event
 	netns := obj.(*networkapi.NetNamespace)
 	klog.V(5).Infof("Watch %s event for NetNamespace %q", eventType, netns.Name)
 
+	// Skip this event if NetID already exists under different netns name
+	if name, found := vmap.findDuplicateNetID(netns.NetName, netns.NetID); found == true {
+		klog.Warningf("Netid %d for namespace %s already exists under different namespace %s",
+			netns.NetID, netns.NetName, name)
+		return
+	}
+
 	// Skip this event if nothing has changed
 	oldNetID, err := vmap.getVNID(netns.NetName)
 	oldMCEnabled := vmap.mcEnabled[netns.NetName]
@@ -225,6 +242,9 @@ func (vmap *nodeVNIDMap) handleDeleteNetNamespace(obj interface{}) {
 	klog.V(5).Infof("Watch %s event for NetNamespace %q", watch.Deleted, netns.Name)
 
 	// Unset VNID first so further operations don't see the deleted VNID
-	vmap.unsetVNID(netns.NetName)
+	if _, err := vmap.unsetVNID(netns.NetName); err != nil {
+		klog.Warningf("Failed to delete namespace %s with error %v", netns.NetName, err)
+		return
+	}
 	vmap.policy.DeleteNetNamespace(netns)
 }
--- a/pkg/network/node/vnids_test.go
+++ b/pkg/network/node/vnids_test.go
@@ -143,6 +143,27 @@ func TestNodeVNIDMap(t *testing.T) {
 	checkNamespaces(t, vmap, 5, []string{"foxtrot"})
 
 	checkAllocatedVNIDs(t, vmap, []uint32{1, 2, 5})
+
+	vmap.setVNID("dup_ns1", 10, false)
+	vmap.setVNID("dup_ns2", 20, false)
+
+	checkForDuplicateNetID(t, vmap, "ns1", 10) // duplicate with dup_ns1
+	checkForDuplicateNetID(t, vmap, "ns2", 20) // duplicate with dup_ns2
+
+	checkForNoDuplicateNetID(t, vmap, "ns1", 15)
+	checkForNoDuplicateNetID(t, vmap, "ns2", 25)
+}
+
+func checkForDuplicateNetID(t *testing.T, vmap *nodeVNIDMap, name string, id uint32) {
+	if _, found := vmap.findDuplicateNetID(name, id); found == false {
+		t.Fatalf("Unexpected failure: name %s id %d", name, id)
+	}
+}
+
+func checkForNoDuplicateNetID(t *testing.T, vmap *nodeVNIDMap, name string, id uint32) {
+	if _, found := vmap.findDuplicateNetID(name, id); found == true {
+		t.Fatalf("Unexpected failure: name %s id %d", name, id)
+	}
 }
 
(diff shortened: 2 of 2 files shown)


pkg/network/node/vnids.go
```diff
@@ -136,6 +136,21 @@ func (vmap *nodeVNIDMap) getVNID(name string) (uint32, error) {
 	return 0, fmt.Errorf("failed to find netid for namespace: %s in vnid map", name)
 }
 
+func (vmap *nodeVNIDMap) findDuplicateNetID(namespace string, netID uint32) (bool) {
+	vmap.lock.Lock()
+	defer vmap.lock.Unlock()
+
+	for name, id := range vmap.ids {
+		if name != namespace && id == netID {
+			klog.Warningf("Netid %d for namespace %s already exists under different namespsace %s",
```
alexanderConstantinescu, line 145: `s/namespsace/namespace`

pkg/network/node/vnids_test.go
```diff
@@ -143,6 +143,27 @@ func TestNodeVNIDMap(t *testing.T) {
...
+	checkForDuplicateNetID(t, vmap, "ns1", 10) // duplicate with dup_ns1
+	checkForDuplicateNetID(t, vmap, "ns2", 20) // duplicate with dup_ns2
+
+	checkForNoDuplicateNetID(t, vmap, "ns1", 15)
+	checkForNoDuplicateNetID(t, vmap, "ns2", 25)
+}
+
+func checkForDuplicateNetID( t *testing.T, vmap *nodeVNIDMap, name string, id uint32) {
+	if vmap.findDuplicateNetID(name, id) != true {
+		t.Fatalf("Unexpected success: name %s id %d", name, id)
```
alexanderConstantinescu, line 159: `s/Unexpected success/Unexpected failure`

alexanderConstantinescu: Moreover, I prefer to compare equality rather than inequality when possible, so please change to  `vmap.findDuplicateNetID(name, id) == false`

pkg/network/node/vnids.go
```diff
@@ -208,6 +223,9 @@ func (vmap *nodeVNIDMap) handleAddOrUpdateNetNamespace(obj, _ interface{}, event
 	oldNetID, err := vmap.getVNID(netns.NetName)
 	oldMCEnabled := vmap.mcEnabled[netns.NetName]
 	mcEnabled := netnsIsMulticastEnabled(netns)
+	if vmap.findDuplicateNetID(netns.NetName, netns.NetID) {
```
alexanderConstantinescu, line 226: This has no relation to this block of code, which is semantically tied to the comment `// Skip this event if nothing has changed`. Could you please move this before the comment or after?

alexanderConstantinescu (changes requested):
I don't think this is the right change. Reading through the BZ three options were proposed, this was not one of them. What we'll be doing now is silently fail - I doubt a user doing something like this will read through the SDN logs. We should either update the invalid VNID or at the very least trigger an evident message / notification to the end-user (for example by: triggering an event) telling them to update it. 

That being said: I am actually for solution 1) in the BZ, which is: get docs to update: https://docs.openshift.com/container-platform/4.7/networking/openshift_sdn/assigning-egress-ips.html and tell them to specifically mention that we don't support the user creating netnamespaces, they should be updating them (that would also mean that we can back-port the change across the OCP 4.X more easily).

msherif1234:
> I don't think this is the right change. Reading through the BZ three options were proposed, this was not one of them. What we'll be doing now is silently fail - I doubt a user doing something like this will read through the SDN logs. We should either update the invalid VNID or at the very least trigger an evident message / notification to the end-user (for example by: triggering an event) telling them to update it.
> 
> That being said: I am actually for solution 1) in the BZ, which is: get docs to update: https://docs.openshift.com/container-platform/4.7/networking/openshift_sdn/assigning-egress-ips.html and tell them to specifically mention that we don't support the user creating netnamespaces, they should be updating them (that would also mean that we can back-port the change across the OCP 4.X more easily).

I did check with Dan Winship yesterday about options in the BZ and he said if we see netns that has NetID which duplicate with existing netns we need to log warning and ignore it.

pkg/network/node/vnids.go
```diff
@@ -136,6 +136,21 @@ func (vmap *nodeVNIDMap) getVNID(name string) (uint32, error) {
 	return 0, fmt.Errorf("failed to find netid for namespace: %s in vnid map", name)
 }
 
+func (vmap *nodeVNIDMap) findDuplicateNetID(namespace string, netID uint32) (bool) {
+	vmap.lock.Lock()
+	defer vmap.lock.Unlock()
+
+	for name, id := range vmap.ids {
```
danwinship, line 143: `vmap` has indexes both by name and by id; you can call `vmap.GetNamespaces(vnid)` to get the list of namespaces with a given vnid (which will hopefully be either empty, or just a one-element list containing the namespace you were expecting)

pkg/network/node/vnids.go
```diff
@@ -204,6 +219,11 @@ func (vmap *nodeVNIDMap) handleAddOrUpdateNetNamespace(obj, _ interface{}, event
 	netns := obj.(*networkapi.NetNamespace)
 	klog.V(5).Infof("Watch %s event for NetNamespace %q", eventType, netns.Name)
 
+	// Skip this event if NetID already exists under different netns name
+	if vmap.findDuplicateNetID(netns.NetName, netns.NetID) {
```
danwinship, line 223: you should update `handleDeleteNetNamespace` too, to not call `vmap.policy.DeleteNetNamespace` if `vmap.unsetVNID` failed (since `policy.DeleteNetNamespace` might do the wrong thing if you pass it a NetNamespace with a duplicate vnid)

pkg/network/node/vnids.go
```diff
@@ -136,6 +136,18 @@ func (vmap *nodeVNIDMap) getVNID(name string) (uint32, error) {
 	return 0, fmt.Errorf("failed to find netid for namespace: %s in vnid map", name)
 }
 
+func (vmap *nodeVNIDMap) findDuplicateNetID(namespace string, netID uint32) (bool) {
+	names := vmap.GetNamespaces(netID)
+	for _, name := range names {
+		if name != namespace {
+			klog.Warningf("Netid %d for namespace %s already exists under different namespace %s",
+						  netID, namespace, name)
```
danwinship, line 144: indentation is wrong here, which is making the `verify` test fail. Run `gofmt` to fix it

pkg/network/node/vnids.go
```diff
@@ -136,6 +136,18 @@ func (vmap *nodeVNIDMap) getVNID(name string) (uint32, error) {
 	return 0, fmt.Errorf("failed to find netid for namespace: %s in vnid map", name)
 }
 
+func (vmap *nodeVNIDMap) findDuplicateNetID(namespace string, netID uint32) bool {
+	names := vmap.GetNamespaces(netID)
+	for _, name := range names {
+		if name != namespace {
```
alexanderConstantinescu, line 142: Shouldn't this be `if name == namspace`?

alexanderConstantinescu (commented):
Minor comment, please correct me if I am wrong on it. But that looks strange to me.

pkg/network/node/vnids.go
alexanderConstantinescu: Or no, this seems right actually given that we're retrieving based on `netID`

msherif1234: no the logic I have is right if we found netID but in different namespace we indicate the duplicate

pkg/network/node/vnids.go
```diff
@@ -136,6 +136,18 @@ func (vmap *nodeVNIDMap) getVNID(name string) (uint32, error) {
...
+			klog.Warningf("Netid %d for namespace %s already exists under different namespace %s",
+				netID, namespace, name)
```
danwinship, line 144: sorry, the warning should be in the caller, not here. Helper functions should (ideally) not log messages

msherif1234: the reason I had it here because I had access to all useful info netid, incorrect ns and the expected ns, if I moved to the caller I won't be able to have this guranularity ?

PR #295, opened 2021-04-29
Bug 1953257: Fix invalid egress IP assignments in sdn-controller

alexanderConstantinescu:
sdn-controller wasn't checking if any of the egress IPs requested for a
namespace had already been assigned to a given node. This led to
confusing results in `hostsubnet.egressIPs` where multiple egress IPs
could be assigned to the same node (though in reality no setup was made
because the sdn pod on the node invalidated the assignment). This patch
now fixes that and only has sdn-controller assign the egress IPs that
can, leaving the rest of them out of the result.

/assign @danwinship

### DIFF
--- a/pkg/network/common/egressip.go
+++ b/pkg/network/common/egressip.go
@@ -445,6 +445,7 @@ func (eit *EgressIPTracker) syncEgressNamespaceState(ns *namespaceEgress) {
 			activeEgressIPs = nil
 			break
 		}
+		eg.assignedVNID = ns.vnid
 		if eg.assignedNodeIP == "" {
 			klog.V(4).Infof("VNID %d cannot use unassigned egress IP %s", ns.vnid, eg.ip)
 		} else if len(ns.requestedIPs) > 1 && eg.nodes[0].offline {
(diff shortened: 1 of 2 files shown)


pkg/network/common/egressip.go
```diff
@@ -530,17 +531,22 @@ func (eit *EgressIPTracker) Ping(ip string, timeout time.Duration) bool {
...
+func (eit *EgressIPTracker) findEgressIPAllocation(eip *egressIPInfo, allocation map[string][]string) (string, bool) {
 	bestNode := ""
 	otherNodes := false
 
 	for _, node := range eit.nodes {
 		if node.offline {
 			continue
 		}
+		if namespace, ok := eit.namespacesByVNID[eip.assignedVNID]; ok {
+			if sets.NewString(allocation[node.nodeName]...).HasAny(namespace.requestedIPs...) {
```
danwinship, line 543: This could maybe be a little less ad-hoc... like, you could keep track of the namespaces/VNIDs currently represented on a node in its `nodeEgress`?

or at least just split this out into a function with an obvious name like `nodeHasEgressIPForNamespace`

and anyway, it shouldn't be based on `namespace.requestedIPs` it should be based on `namespace.activeEgressIPs`. I think?

danwinship (changes requested):
you should add something to the unit test

PR #296, opened 2021-04-30
proxy: filter/delegate EndpointSlices as well as Endpoints

danwinship:
Updates `OsdnProxy`/`HybridProxier` to handle both Endpoints and EndpointSlices, also adding unit tests and doing a bunch of refactoring along the way.

The sdn-level code is compatible with `EndpointSliceProxying` being either enabled or disabled, and also compatible with EndpointSlice support being added to the userspace proxy in the future. (In fact, this PR alone doesn't actually make openshift-sdn use EndpointSlices, because CNO is currently always passing `--feature-gates=EndpointSliceProxying=false`. But once this merges that can be reverted.

(for 4.9)

/cc @aojea

### DIFF
--- a/pkg/network/common/common.go
+++ b/pkg/network/common/common.go
@@ -88,6 +88,26 @@ func ParseClusterNetwork(cn *networkv1.ClusterNetwork) (*ParsedClusterNetwork, e
 	return pcn, nil
 }
 
+// PodNetworkContains determines whether pcn's pod network contains ip
+func (pcn *ParsedClusterNetwork) PodNetworkContains(ip net.IP) bool {
+	for _, cn := range pcn.ClusterNetworks {
+		if cn.ClusterCIDR.Contains(ip) {
+			return true
+		}
+	}
+	return false
+}
+
+// ServiceNetworkContains determines whether pcn's service network contains ip
+func (pcn *ParsedClusterNetwork) ServiceNetworkContains(ip net.IP) bool {
+	if pcn.ServiceNetwork != nil {
+		if pcn.ServiceNetwork.Contains(ip) {
+			return true
+		}
+	}
+	return false
+}
+
 func (pcn *ParsedClusterNetwork) ValidateNodeIP(nodeIP string) error {
 	if nodeIP == "" || nodeIP == "127.0.0.1" {
 		return fmt.Errorf("invalid node IP %q", nodeIP)
@@ -132,7 +152,7 @@ func (pcn *ParsedClusterNetwork) CheckClusterObjects(subnets []networkv1.HostSub
 		subnetIP, _, _ := net.ParseCIDR(subnet.Subnet)
 		if subnetIP == nil {
 			errList = append(errList, fmt.Errorf("failed to parse network address: %s", subnet.Subnet))
-		} else if _, contains := ClusterNetworkListContains(pcn.ClusterNetworks, subnetIP); !contains {
+		} else if !pcn.PodNetworkContains(subnetIP) {
 			errList = append(errList, fmt.Errorf("existing node subnet: %s is not part of any cluster network CIDR", subnet.Subnet))
 		}
 		if len(errList) >= 10 {
@@ -143,7 +163,11 @@ func (pcn *ParsedClusterNetwork) CheckClusterObjects(subnets []networkv1.HostSub
 		if pod.Spec.HostNetwork {
 			continue
 		}
-		if _, contains := ClusterNetworkListContains(pcn.ClusterNetworks, net.ParseIP(pod.Status.PodIP)); !contains && pod.Status.PodIP != "" {
+		podIP := net.ParseIP(pod.Status.PodIP)
+		if podIP == nil {
+			continue
+		}
+		if !pcn.PodNetworkContains(podIP) {
 			errList = append(errList, fmt.Errorf("existing pod %s:%s with IP %s is not part of cluster network", pod.Namespace, pod.Name, pod.Status.PodIP))
 			if len(errList) >= 10 {
 				break
@@ -152,7 +176,10 @@ func (pcn *ParsedClusterNetwork) CheckClusterObjects(subnets []networkv1.HostSub
 	}
 	for _, svc := range services {
 		svcIP := net.ParseIP(svc.Spec.ClusterIP)
-		if svcIP != nil && !pcn.ServiceNetwork.Contains(svcIP) {
+		if svcIP == nil {
+			continue
+		}
+		if !pcn.ServiceNetworkContains(svcIP) {
 			errList = append(errList, fmt.Errorf("existing service %s:%s with IP %s is not part of service network %s", svc.Namespace, svc.Name, svc.Spec.ClusterIP, pcn.ServiceNetwork.String()))
 			if len(errList) >= 10 {
 				break
--- a/pkg/network/proxy/hybridproxier.go
+++ b/pkg/network/proxy/hybridproxier.go
@@ -1,4 +1,4 @@
-package hybrid
+package proxy
 
 import (
 	"fmt"
--- a/pkg/network/proxy/hybridproxier.go
+++ b/pkg/network/proxy/hybridproxier.go
@@ -34,8 +34,9 @@ type HybridProxier struct {
 
 	mainProxy     HybridizableProxy
 	unidlingProxy HybridizableProxy
-	minSyncPeriod time.Duration
+
 	serviceLister corev1listers.ServiceLister
+	syncRunner    *async.BoundedFrequencyRunner
 
 	// TODO(directxman12): figure out a good way to avoid duplicating this information
 	// (it's saved in the individual proxies as well)
@@ -52,9 +53,6 @@ type HybridProxier struct {
 	switchedToUserspace     map[types.NamespacedName]bool
 	switchedToUserspaceLock sync.Mutex
 
-	// A gate that prevents iptables from being exhausted by proxy calls.
-	syncRunner *async.BoundedFrequencyRunner
-
 	// This map is used in unidling proxy mode to ensure the service is correctly deleted
 	// if it's deletion occurs after the deletion of the endpoint.
 	pendingDeletion map[types.NamespacedName]bool
@@ -69,7 +67,7 @@ func NewHybridProxier(
 	p := &HybridProxier{
 		mainProxy:     mainProxy,
 		unidlingProxy: unidlingProxy,
-		minSyncPeriod: minSyncPeriod,
+
 		serviceLister: serviceLister,
 
 		usingUserspace:      make(map[types.NamespacedName]bool),
--- a/pkg/network/proxy/hybridproxier.go
+++ b/pkg/network/proxy/hybridproxier.go
@@ -235,21 +235,15 @@ func (p *HybridProxier) switchService(svcName types.NamespacedName) {
 func (p *HybridProxier) OnEndpointsAdd(endpoints *corev1.Endpoints) {
 	svcName := types.NamespacedName{Namespace: endpoints.Namespace, Name: endpoints.Name}
 
-	// we track all endpoints in the unidling endpoints handler so that we can succesfully
-	// detect when a service become unidling
-	klog.V(6).Infof("(always) add ep %s in unidling proxy", svcName)
+	klog.V(6).Infof("add ep %s", svcName)
 	p.unidlingProxy.OnEndpointsAdd(endpoints)
+	p.mainProxy.OnEndpointsAdd(endpoints)
 
 	p.usingUserspaceLock.Lock()
 	defer p.usingUserspaceLock.Unlock()
 
 	wasUsingUserspace, knownEndpoints := p.usingUserspace[svcName]
 	p.usingUserspace[svcName] = p.shouldEndpointsUseUserspace(endpoints)
-
-	if !p.usingUserspace[svcName] {
-		klog.V(6).Infof("add ep %s in main proxy", svcName)
-		p.mainProxy.OnEndpointsAdd(endpoints)
-	}
 
 	// a service could appear before endpoints, so we have to treat this as a potential
 	// state modification for services, and not just an addition (since we could flip proxies).
@@ -261,10 +255,9 @@ func (p *HybridProxier) OnEndpointsAdd(endpoints *corev1.Endpoints) {
 func (p *HybridProxier) OnEndpointsUpdate(oldEndpoints, endpoints *corev1.Endpoints) {
 	svcName := types.NamespacedName{Namespace: endpoints.Namespace, Name: endpoints.Name}
 
-	// we track all endpoints in the unidling endpoints handler so that we can succesfully
-	// detect when a service become unidling
-	klog.V(6).Infof("(always) update ep %s in unidling proxy", svcName)
+	klog.V(6).Infof("update ep %s", svcName)
 	p.unidlingProxy.OnEndpointsUpdate(oldEndpoints, endpoints)
+	p.mainProxy.OnEndpointsUpdate(oldEndpoints, endpoints)
 
 	p.usingUserspaceLock.Lock()
 	defer p.usingUserspaceLock.Unlock()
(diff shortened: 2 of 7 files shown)


pkg/network/proxy/proxy.go
```diff
@@ -337,7 +337,7 @@ func (proxy *OsdnProxy) OnEndpointsUpdate(old, ep *corev1.Endpoints) {
 	case !wasBlocked && !pep.blocked:
 		proxy.baseProxy.OnEndpointsUpdate(old, ep)
 	case !wasBlocked && pep.blocked:
-		proxy.baseProxy.OnEndpointsDelete(ep)
+		proxy.baseProxy.OnEndpointsDelete(old)
```
aojea, line 478: I see, wasBlocked tell us that it was not added previously

pkg/network/common/common.go
```diff
@@ -88,6 +79,21 @@ func ParseClusterNetwork(cn *networkv1.ClusterNetwork) (*ParsedClusterNetwork, e
 	return pcn, nil
 }
 
+// Contains determines whether pcn contains ip
```
aojea, line 82: nit, I see more readable to have 2 methods instead , i.e
`PodNetworkContains()` and `ServiceNetworkContains()`
 discrimination on the output variables makes the code to use a lot underscores for unused variables

pkg/network/proxy/proxy.go
```diff
@@ -161,35 +177,34 @@ func (proxy *OsdnProxy) handleAddOrUpdateNetNamespace(obj, _ interface{}, eventT
...
 
-	if proxy.ids[ns] == network.GlobalVNID {
-		return true
+	ns := proxy.namespaces[netns.Name]
+	if ns == nil {
+		return
 	}
-	return false
+
+	ns.global = false
```
aojea, line 253: if we want to delete a global namespace do we have to force it to be no global?

pkg/network/proxy/proxy.go
```diff
@@ -161,35 +177,34 @@ func (proxy *OsdnProxy) handleAddOrUpdateNetNamespace(obj, _ interface{}, eventT
...
+	proxy.maybeDeleteNamespace(netns.Name, ns)
 }
 
 func (proxy *OsdnProxy) updateEgressNetworkPolicy(policy networkv1.EgressNetworkPolicy) {
-	ns := policy.Namespace
-	if proxy.isNamespaceGlobal(ns) {
+	ns := proxy.getNamespace(policy.Namespace)
```
aojea, line 258: this call is not under the `proxy.Lock()`

pkg/network/proxy/proxy.go
```diff
@@ -221,73 +236,65 @@ func (proxy *OsdnProxy) updateEgressNetworkPolicy(policy networkv1.EgressNetwork
...
-	if ref, ok := proxy.firewall[ns]; ok {
-		if len(ref.namespaceFirewalls) == 1 {
-			for uid := range ref.namespaceFirewalls {
-				ref.activePolicy = &uid
-				klog.Infof("Applied firewall egress network policy: %q to namespace: %q", uid, ns)
-			}
-		} else {
-			ref.activePolicy = nil
+	if len(ns.firewalls) == 1 {
+		for uid := range ns.firewalls {
```
aojea, line 300: if len == 1 it must be ns.firewalls[0], no?

pkg/network/proxy/proxy.go
```diff
@@ -221,73 +236,65 @@ func (proxy *OsdnProxy) updateEgressNetworkPolicy(policy networkv1.EgressNetwork
...
+	}
 }
 
-func (proxy *OsdnProxy) firewallBlocksIP(namespace string, ip net.IP) bool {
-	if ref, ok := proxy.firewall[namespace]; ok {
-		if ref.activePolicy == nil {
-			// Block all connections if active policy is not set
-			return true
-		}
+func (ns *proxyNamespace) firewallBlocksIP(ip net.IP) bool {
```
aojea: this is not thread safe

aojea: nevermind, it is called from EndpointsAdd and EndpointsUpdated that hold the lock

pkg/network/proxy/proxy.go
```diff
@@ -318,36 +329,53 @@ func (proxy *OsdnProxy) OnEndpointsAdd(ep *corev1.Endpoints) {
...
 	defer proxy.Unlock()
 
 	ns := proxy.getNamespace(ep.Namespace)
-	pep := ns.endpoints[ep.UID]
+	isBlockable := proxy.endpointsBlockable(ns, ep)
+	isBlocked := isBlockable && proxy.endpointsBlocked(ns, ep)
+
+	pep := ns.blockableEndpoints[ep.UID]
 	if pep == nil {
-		klog.Warningf("Got OnEndpointsUpdate for unknown Endpoints %#v", ep)
```
aojea, line 334: we should keep warning that this endpoint wasn't "blockable" before

pkg/network/proxy/hybridproxier.go
```diff
@@ -8,11 +8,11 @@ import (
 	"k8s.io/klog/v2"
 
 	corev1 "k8s.io/api/core/v1"
+	discoveryv1beta1 "k8s.io/api/discovery/v1beta1"
```
aojea, line 11: you should have the v1 , it went GA in 1.21 https://github.com/kubernetes/kubernetes/pull/99870

pkg/network/proxy/hybridproxier.go
```diff
@@ -22,6 +22,9 @@ import (
 type HybridizableProxy interface {
 	proxy.Provider
 
+	UsesEndpoints() bool
+	UsesEndpointSlices() bool
```
aojea, line 26: should we make this exlusive, if UsesEndpointSlices false means UseEndpoints true?

aojea: nvm, it is based on the featuregate 
```
	return !utilfeature.DefaultFeatureGate.Enabled(features.EndpointSliceProxying)
```

pkg/network/proxy/proxy.go
danwinship: If we want `proxy.maybeDeleteNamespace` to garbage-collect the namespace, it has to be marked as non-global (since global namespaces are always interesting and so can't be GC'ed). But a namespace with no associated NetNamespace can't be global, so when a global namespace's NetNamespace is deleted, we mark it as being no-longer global, so that it can potentially be cleaned up.

pkg/network/proxy/proxy.go
danwinship: `ns.firewalls` is a map; if there is only 1 value, we want the key of that value.

Yeah, this code is really weird, but it was already like that before; this code is just being moved around. But yeah, it should probably be rewritten to be less weird.

pkg/network/proxy/proxy.go
danwinship: I should be clearer about what functions assume the lock is already held...

pkg/network/proxy/proxy.go
```diff
@@ -273,14 +273,21 @@ func (proxy *OsdnProxy) updateEgressNetworkPolicy(policy networkv1.EgressNetwork
...
-		return false
-	} else if ns.activePolicy == nil {
-		// Block all connections if active policy is not set
-		return true
+// Returns true if ep contains at least one blockable (ie, non-local) endpoint
+func (proxy *OsdnProxy) endpointsBlockable(ns *proxyNamespace, ep *corev1.Endpoints) bool {
+	for _, ss := range ep.Subsets {
+		for _, addr := range ss.Addresses {
+			ip := net.ParseIP(addr.IP)
+			if _, _, subnet := proxy.networkInfo.Contains(ip); subnet == nil {
```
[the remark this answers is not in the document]
danwinship: can't be changed at runtime

pkg/network/proxy/hybridproxier.go
danwinship: ah, copied this from Clayton's earlier PR...

pkg/network/proxy/hybridproxier.go
danwinship: It's not exclusive. In particular, when the feature gate is enabled, `HybridProxier` returns true for both `UsesEndpoints` and `UsesEndpointSlices`, because the unidling proxy needs Endpoints but the iptables proxy needs EndpointSlices.

pkg/network/proxy/proxy.go
danwinship: It is. `updateEgressNetworkPolicy` requires its caller to have gotten the lock. I added some comments clarify this.

pkg/network/proxy/proxy.go
danwinship: That was a "can't happen" situation, and with the new code organization, we can't tell the difference between "we don't know about this namespace because it wasn't worth paying attention to before" and "we don't know about this namespace because the informer is buggy". So we just assume it's the former.

pkg/network/proxy/hybridproxier.go
danwinship: actually, `proxy.Provider` still uses the `v1beta1` API

pkg/network/common/common.go
```diff
@@ -88,6 +88,26 @@ func ParseClusterNetwork(cn *networkv1.ClusterNetwork) (*ParsedClusterNetwork, e
...
+		}
+	}
+	return false
+}
+
+// ServiceNetworkContains determines whether pcn's service network contains ip
+func (pcn *ParsedClusterNetwork) ServiceNetworkContains(ip net.IP) bool {
+	if pcn.ServiceNetwork != nil {
+		if pcn.ServiceNetwork.Contains(ip) {
+			return false
```
aojea, line 105: return true

pkg/network/proxy/hybridproxier_test.go
```diff
@@ -0,0 +1,291 @@
...
+	proxy.OnEndpointsAdd(ep1)
+
+	mainProxy.assertEvents("after creating first endpoints",
+		"add endpoints testns/one 1.2.3.4",
+	)
+	unidlingProxy.assertEvents("after creating first endpoints",
+		"add endpoints testns/one 1.2.3.4",
+	)
+
+	// Idle the service; both the service and the endpoints will be removed from the
```
aojea, line 94: didn't we move to label services instead of endpoints to support endpointslices too?

pkg/network/proxy/hybridproxier.go
```diff
@@ -274,10 +268,9 @@ func (p *HybridProxier) OnEndpointsAdd(endpoints *corev1.Endpoints) {
 }
 
 func (p *HybridProxier) OnEndpointsUpdate(oldEndpoints, endpoints *corev1.Endpoints) {
-	// we track all endpoints in the unidling endpoints handler so that we can succesfully
-	// detect when a service become unidling
-	klog.V(6).Infof("hybrid proxy: (always) update ep %s/%s in unidling proxy", endpoints.Namespace, endpoints.Name)
+	klog.V(6).Infof("hybrid proxy: update ep %s/%s", endpoints.Namespace, endpoints.Name)
 	p.unidlingProxy.OnEndpointsUpdate(oldEndpoints, endpoints)
+	p.mainProxy.OnEndpointsUpdate(oldEndpoints, endpoints)
```
aojea: any possible performance implications related to the duplication?

pkg/network/proxy/hybridproxier_test.go
danwinship: Did we? We didn't _stop_ labeling Endpoints though, and we still have to track those anyway...
It would make the code cleaner if we could track idleness via Services, but that can happen later.

pkg/network/proxy/hybridproxier.go
danwinship: The change that you are commenting on shouldn't really have noticeable performance implications. Unless you have a cluster with LOTS of idled services, the main proxy would be receiving almost all of the events anyway. It might even end up helping performance by getting rid of pointless allocations and frees...

OTOH, the "watch both Endpoints and EndpointSlices" commit does have performance implications... we'll now have both Endpoints and EndpointSlice informers, and so will be caching both sets of objects.

Perhaps if we can use Services rather than Endpoints to track idling, then we could cheat and just synthesize fake Endpoints from the EndpointSlice events to pass to the unidling proxy, rather than actually listening for real Endpoints events?

aojea: we do have annotations on services 
https://github.com/openshift/oc/pull/720

aojea: https://github.com/openshift/openshift-controller-manager/commit/dca955d79996d6c9e8de574a79b699d2d3437d8d

smarterclayton: I am strongly in favor of at least attempting to convert endpoints to synthetic EndpointSlices.  Double watching doubles memory usage and CPU and network traffic to control plane on all nodes on big dense clusters.  It might (for instance) blow up a very large cluster.

danwinship:
/assign @squeed 
This is not really related to bz 1956535 at all but
1. if you're going to be looking at the proxy code anyway, can you review?
2. there are two small drive-by bug fixes in here which probably aren't relevant but maybe? (in `OsdnProxy: minor fix to "endpoints become blocked" case` and `OsdnProxy: minor fix to "endpoints become blocked" case`)
3. the logging changes in `HybridProxier: logging cleanup/consistency` might help proxy problem debugging in the future and I'm thinking everything up to there (unit tests, code reorg, drive-by fixes, logging) can be backported?
4. (I was thinking that `HybridProxier: don't pass irrelevant Endpoints to unidling proxy` might help with scale-related problems, but that's just internal memory usage, not iptables interactions, so never mind...)

pkg/network/proxy/hybridproxier.go
```diff
@@ -0,0 +1,451 @@
...
+func (proxier *HybridProxier) OnNodeDelete(node *corev1.Node) {
+	// TODO implement https://github.com/kubernetes/enhancements/pull/640
+}
+
+func (proxier *HybridProxier) OnNodeSynced() {
+	// TODO implement https://github.com/kubernetes/enhancements/pull/640
+}
+
+func (p *HybridProxier) getService(svcName types.NamespacedName) *hybridProxierService {
+	p.serviceLock.Lock()
```
squeed, line 133: Want to add a comment why you're not releasing this lock?

pkg/network/proxy/hybridproxier.go
```diff
@@ -0,0 +1,451 @@
...
+func (p *HybridProxier) Sync() {
+	p.syncRunner.Run()
+}
+
+// syncProxyRules actually applies the proxy rules to the node.
+// It is called by our SyncRunner.
+// We do this so that we can guarantee that changes are applied to both
+// proxies, especially when unidling a newly-awoken service.
+func (p *HybridProxier) syncProxyRules() {
+	klog.V(4).Infof("syncProxyRules start")
```
squeed, line 432: minor nit: can this be V(3)?

pkg/network/proxy/proxy.go
```diff
@@ -229,83 +265,135 @@ func (proxy *OsdnProxy) updateEgressNetworkPolicy(policy networkv1.EgressNetwork
...
-		}
-		proxy.firewall[ns].namespaceFirewalls[policy.UID] = firewall
-	} else if _, ok := proxy.firewall[ns]; ok {
-		delete(proxy.firewall[ns].namespaceFirewalls, policy.UID)
-		if len(proxy.firewall[ns].namespaceFirewalls) == 0 {
-			delete(proxy.firewall, ns)
-		}
+		ns.firewalls[policy.UID] = firewall
+	} else {
+		delete(ns.firewalls, policy.UID)
```
squeed, line 270: While you're at it, do you want to move the key from UID to NamespacedName? If an object is deleted-and-recreated with the same name, we could leak objects.

squeed: I agree this code is weird - if you don't want to touch it, that's OK.

pkg/network/proxy/hybridproxier.go
```diff
@@ -0,0 +1,451 @@
...
+//   - Idled (known to the unidlingProxy but not the mainProxy). An Idled Service becomes
+//     Unidling when either its Service annotation is removed or its Endpoints become
+//     non-empty.
+//
+//   - Unidling (the Service and Endpoints are known to the mainProxy, and the Endpoints
+//     are known to the unidling proxy). While a Service is Unidling, Endpoints events are
+//     sent to both proxies (so the unidling proxy socket can redirect its connection to
+//     the correct place). An Unidling Service becomes Not Idled if its Endpoints are
+//     deleted, or else the next time it receives an Endpoints event more than 1 minute
+//     after becoming Unidling. (Alternatively it could also become Idled again.)
```
squeed, line 47: This timeline is different from the previous behavior - why did you do this? Wouldn't a service transition from Unidling to Not Idled iff it has non-empty endpoints? I'm assuming it's because a service can't possible unidle for more than a minute, but does that actually help?

Also, why would deleting Endpoints make a service Not Idled? Just because it means the service is going away?

danwinship: The old behavior was that we always sent all Endpoints events to the unidling proxy. The new behavior is that we're trying to not do that, because we have to synthesize fake Endpoints objects from EndpointSlices, and we don't want to have to do that for every EndpointSlice and double our memory usage (especially since almost all of the undling Endpoints updates are pointless; the unidling proxy has no need to ever know the Endpoints of Services that never get idled).

So the goal is "figure out if the unidling proxy actually cares about the endpoints for this service or not, and if not, then don't bother synthesizing an Endpoints from the EndpointSlice if it doesn't actually need to know".

When a Service has just un-idled and the UnidlerSocket still has pending connections to proxy, we need to send it Endpoints updates so it can know where to proxy the connections to. Once it has connected all of its pending connections, then it doesn't need to know about any further Endpoints events. It's likely this will happen within a few seconds of it receiving the first non-empty Endpoints update.

> Wouldn't a service transition from Unidling to Not Idled iff it has non-empty endpoints?

If the Service is backed by a Deployment/ReplicaSet/etc, then when it un-idles, we will first get an Endpoints update saying it has 1 endpoint, quickly followed by another Endpoints update saying it has 2 endpoints, etc. We _could_ decide we were only going to send the unidling proxy the first update (so that all connections captured by the unidling proxy would get proxied to the first endpoint to come up) but it seemed better to send it the others too?

danwinship: > Also, why would deleting Endpoints make a service Not Idled? Just because it means the service is going away?

yes

PR #299, opened 2021-05-14
Bug 1959711: Reverse table order for egress IP and egress network policy set up 

alexanderConstantinescu:
A bug was introduced during the implementation of the egress IP with the
traffic load balancing feature, namely: the egress IP flows were evaluated
before the egress network policy ones - just like it was before, however: where
previously the egress IP related flows terminated in a `goto_table:101`
action (allowing the egress network policy flows to determine the final
action) the new feature had it end in `output:tun0`(thus short circuiting
any egress network policy related flows).

Since OVS group actions does not allow for a `goto_table` action, we
need to reverse the table ordering of egress IP and egress network
policy. This patch implements that.

Bug: https://bugzilla.redhat.com/show_bug.cgi?id=1959711 required that
we short circuit the egress IP flows for reply traffic (in case the
egress IP matching pod is the server for whatever connection) and send
directly to egress netwok policy. This has been re-worked a bit to keep
the functionality the same. We now add the flow to the egress IP related
one, knowing that if we hit that flow the egress network policy ones
have already been evaluated. We thus need to short circuit the egress IP
flows, which we do due to the higher priority.

/assign @danwinship @JacobTanenbaum 
/cc @huiran0826 

@huiran0826 - could you re-test the suite of egress network policy & egress IP test cases? I want to verify that this works without introducing a regression, sorry for the added work.

### DIFF
--- a/pkg/network/node/egressip_test.go
+++ b/pkg/network/node/egressip_test.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"reflect"
 	"sort"
-	//"strconv"
 	"strings"
 	"testing"
 
@@ -72,7 +71,7 @@ type egressOVSChange struct {
 // the new/current set of flows in flows.
 func assertOVSChanges(eip *egressIPWatcher, flows *[]string, changes ...egressOVSChange) error {
 	oldFlows := *flows
-	newFlows, err := eip.oc.ovs.DumpFlows("table=100")
+	newFlows, err := eip.oc.ovs.DumpFlows("table=101")
 	if err != nil {
 		return fmt.Errorf("unexpected error dumping OVS flows: %v", err)
 	}
@@ -176,7 +175,7 @@ func setupEgressIPWatcher(t *testing.T) (*egressIPWatcher, []string) {
 	eip := newEgressIPWatcher(oc, "172.17.0.4", &masqBit)
 	eip.testModeChan = make(chan string, 10)
 
-	flows, err := eip.oc.ovs.DumpFlows("table=100")
+	flows, err := eip.oc.ovs.DumpFlows("table=101")
 	if err != nil {
 		t.Fatalf("unexpected error dumping OVS flows: %v", err)
 	}
--- a/pkg/network/node/ovscontroller.go
+++ b/pkg/network/node/ovscontroller.go
@@ -36,7 +36,7 @@ const (
 	Vxlan0 = "vxlan0"
 
 	// rule versioning; increment each time flow rules change
-	ruleVersion = 10
+	ruleVersion = 11
 
 	ruleVersionTable = 253
 )
@@ -133,6 +133,7 @@ func (oc *ovsController) SetupOVS(clusterNetworkCIDR []string, serviceNetworkCID
 	// eg, "table=20, priority=100, in_port=${ovs_port}, arp, nw_src=${ipaddr}, arp_sha=${macaddr}, actions=load:${tenant_id}->NXM_NX_REG0[], goto_table:21"
 	//     "table=20, priority=100, in_port=${ovs_port}, ip, nw_src=${ipaddr}, actions=load:${tenant_id}->NXM_NX_REG0[], goto_table:21"
 	// (${tenant_id} is always 0 for single-tenant)
+	otx.AddFlow("table=20, priority=300, udp, udp_dst=%d, actions=drop", vxlanPort)
 	otx.AddFlow("table=20, priority=0, actions=drop")
 
 	// Table 21: from OpenShift container; NetworkPolicy plugin uses this for connection tracking
@@ -165,7 +166,7 @@ func (oc *ovsController) SetupOVS(clusterNetworkCIDR []string, serviceNetworkCID
 	// Multicast coming from local pods
 	otx.AddFlow("table=30, priority=25, ip, nw_dst=224.0.0.0/4, actions=goto_table:110")
 
-	otx.AddFlow("table=30, priority=0, ip, actions=goto_table:100")
+	otx.AddFlow("table=30, priority=0, ip, actions=goto_table:99")
 	otx.AddFlow("table=30, priority=0, arp, actions=drop")
 
 	// Table 40: ARP to local container, filled in by setupPodFlows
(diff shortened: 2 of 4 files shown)


pkg/network/node/egressip_test.go
```diff
@@ -72,7 +73,7 @@ type egressOVSChange struct {
 // the new/current set of flows in flows.
 func assertOVSChanges(eip *egressIPWatcher, flows *[]string, changes ...egressOVSChange) error {
 	oldFlows := *flows
-	newFlows, err := eip.oc.ovs.DumpFlows("table=100")
+	newFlows, err := eip.oc.ovs.DumpFlows("table=101")
```
JacobTanenbaum, line 76: instead of hard coding the table numbers could we make a file that stores the table numbers and priorities or whatnot as constants? perhaps pkg/types/constants.go?

alexanderConstantinescu: We could, I would however not do that for the entire openshift-sdn setup, since that would make this PR a lot bigger than what it already is. I could apply that to only the egress IP and egress network policy tables though.

pkg/network/node/egressip_test.go
```diff
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"reflect"
 	"sort"
+
 	//"strconv"
```
danwinship, line 10: not sure how that ended up there but can you remove both of these lines?

pkg/network/node/ovscontroller.go
```diff
@@ -36,8 +36,8 @@ const (
 	Vxlan0 = "vxlan0"
 
 	// rule versioning; increment each time flow rules change
-	ruleVersion = 10
-
+	ruleVersion      = 11
+	egressIPPriority = 100
```
danwinship, line 40: put that somewhere not between `ruleVersion` and `ruleVersionTable`, which really ought to be together

pkg/network/node/ovscontroller.go
```diff
@@ -199,16 +199,15 @@ func (oc *ovsController) SetupOVS(clusterNetworkCIDR []string, serviceNetworkCID
...
-	// Table 100: egress routing; edited by SetNamespaceEgress*()
-	otx.AddFlow("table=100, priority=300,udp,udp_dst=%d,actions=drop", vxlanPort)
-	otx.AddFlow("table=100, priority=200,tcp,tcp_dst=53,nw_dst=%s,actions=output:2", oc.localIP)
-	otx.AddFlow("table=100, priority=200,udp,udp_dst=53,nw_dst=%s,actions=output:2", oc.localIP)
-	otx.AddFlow("table=100, priority=150,ct_state=+rpl,actions=goto_table:101")
-	// eg, "table=100, priority=100, reg0=${tenant_id}, ip, actions=set_field:${tun0_mac}->eth_dst,set_field:${egress_mark}->pkt_mark,goto_table:101"
-	otx.AddFlow("table=100, priority=0, actions=goto_table:101")
+	// Table 100: egress network policy dispatch; edited by UpdateEgressNetworkPolicy()
+	// eg, "table=100, reg0=${tenant_id}, priority=2, ip, nw_dst=${external_cidr}, actions=drop
+	otx.AddFlow("table=100, priority=0, actions=output:2")
```
danwinship, line 204: needs to be `actions=goto_table:101`

pkg/network/node/ovscontroller.go
```diff
@@ -494,31 +493,34 @@ func (oc *ovsController) UpdateEgressNetworkPolicyRules(policies []networkapi.Eg
...
 	} else /* vnid != 0 && len(policies) == 1 */ {
-		otx.DeleteFlows("table=101, reg0=%d", vnid)
+		otx.DeleteFlows("table=100, reg0=%d", vnid)
 
 		for i, rule := range policies[0].Spec.Egress {
-			priority := len(policies[0].Spec.Egress) - i
+			// The priority here needs to be higher than the egressIPPriority
+			// for the rules to be evaluated in the order: egress network policy
+			// -> egress IP
+			priority := len(policies[0].Spec.Egress) - i + egressIPPriority
```
danwinship, line 519: no... ENPs and egress IPs are in different tables. You shouldn't have to change anything with priority

pkg/network/node/egressip_test.go
danwinship: I had a branch a very long time ago that let you use "variable names" in OVS flows so you could say stuff like like `"table=${EGRESS-IPS}"` but it ended up not working well, IIRC

danwinship: (https://github.com/openshift/origin/commit/479264578ca5e96c31749e65fbaaeee3a6320721 in https://github.com/openshift/origin/pull/10190)

pkg/network/node/ovscontroller.go
alexanderConstantinescu: Removing it altogether. As you mentioned here: https://github.com/openshift/sdn/pull/299/files#r633656457 I shouldn't even need it. The only reason I thought I did was because of what you mentioned here: https://github.com/openshift/sdn/pull/299/files#r633654820

pkg/network/node/egressip_test.go
alexanderConstantinescu: OK, so I don't want to make this PR too intricate this close to code freeze. I am thus going to skip doing this for now. If we feel this really deserves attention going forward, then let's address it during the 4.9 cycle.

alexanderConstantinescu:
@JacobTanenbaum / @danwinship : I've addressed/responded to all comments, PTAL. Thanks in advance!

pkg/network/node/ovscontroller.go
```diff
@@ -199,16 +199,15 @@ func (oc *ovsController) SetupOVS(clusterNetworkCIDR []string, serviceNetworkCID
 	// eg, "table=90, priority=100, ip, nw_dst=${remote_subnet_cidr}, actions=move:NXM_NX_REG0[]->NXM_NX_TUN_ID[0..31], set_field:${remote_node_ip}->tun_dst,output:1"
 	otx.AddFlow("table=90, priority=0, actions=drop")
 
-	// Table 100: egress routing; edited by SetNamespaceEgress*()
-	otx.AddFlow("table=100, priority=300,udp,udp_dst=%d,actions=drop", vxlanPort)
-	otx.AddFlow("table=100, priority=200,tcp,tcp_dst=53,nw_dst=%s,actions=output:2", oc.localIP)
-	otx.AddFlow("table=100, priority=200,udp,udp_dst=53,nw_dst=%s,actions=output:2", oc.localIP)
```
danwinship, line 205: sorry, should have caught this before; these rules should stay in table 100, because we want them to bypass both egress IPs and EgressNetworkPolicy.

well... probably the DNS rules no-ops now anyway since we don't do DNS to the node IP any more in 4.x... but if someone is relying on this, we don't want to break it. (That is, if someone has configured pod DNS to talk to the node IP, and also has an EgressNetworkPolicy saying "Deny 0.0.0.0/0", then DNS to the node IP is still expected to work.)

alexanderConstantinescu: That will need a dedicated table 99 in that case. The reason I say that is because table 100 will now have priorities between 0 and number of rules specific in a egress network policy object. We won't be able to use a deterministic priority which overrides all rules, since we don't know what the max priority for all rules could be.

pkg/network/node/ovscontroller.go
```diff
@@ -199,16 +199,21 @@ func (oc *ovsController) SetupOVS(clusterNetworkCIDR []string, serviceNetworkCID
...
-	otx.AddFlow("table=100, priority=150,ct_state=+rpl,actions=goto_table:101")
-	// eg, "table=100, priority=100, reg0=${tenant_id}, ip, actions=set_field:${tun0_mac}->eth_dst,set_field:${egress_mark}->pkt_mark,goto_table:101"
+	// Table 99: legacy DNS rules needed to override egress IP and egress
+	// network policy, from 3.11. These are used in case someone has configured
+	// pod DNS to talk to the node IP, and also has an EgressNetworkPolicy
+	// saying "Deny 0.0.0.0/0", then DNS to the node IP is still expected to
+	// work
+	otx.AddFlow("table=99, priority=300,udp,udp_dst=%d,actions=drop", vxlanPort)
+	otx.AddFlow("table=99, priority=200,tcp,tcp_dst=53,nw_dst=%s,actions=output:2", oc.localIP)
+	otx.AddFlow("table=99, priority=200,udp,udp_dst=53,nw_dst=%s,actions=output:2", oc.localIP)
```
danwinship, line 209: You need to change the `actions=goto_table:100` rule in table 30 to `goto_table:99` now. Also, you need

    otx.AddFlow("table=99, priority=0, actions=goto_table:100")

at the end. Or, alternatively, keep these rules as table 100 (and do `goto_table:101` at the end), have ENP at 101, and have egress routing at 102. That would be slightly more consistent with how we number things elsewhere but I don't have a strong preference.

danwinship: If you're touching these lines can you add spaces after all the commas in the flow to make them consistent with the rest of the file.

Also, the comment about DNS applies to the two `priority=200` rules, but the `vxlanPort` rule is an unrelated security fix. Actually, you could move that to the top of table 20 if you wanted, since it only applies to traffic coming out of pods.

alexanderConstantinescu: OK, I've done:

- moved the `vxlanPort` rule to the top of table 20
- changed the `actions=goto_table:100` rule in table 30 to `goto_table:99` (and hence kept table 99) 
- Added the missing flow: `otx.AddFlow("table=99, priority=0, actions=goto_table:100")`
- Added spaces after all the commas 

Thanks for spotting these!

pkg/network/node/egressip_test.go
```diff
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"reflect"
 	"sort"
-	//"strconv"
```
danwinship, line 9: fix

pkg/network/node/ovscontroller.go
```diff
@@ -129,6 +129,7 @@ func (oc *ovsController) SetupOVS(clusterNetworkCIDR []string, serviceNetworkCID
 	// eg, "table=10, priority=100, tun_src=${remote_node_ip}, actions=goto_table:30"
 	otx.AddFlow("table=10, priority=0, actions=drop")
 
+	otx.AddFlow("table=20, priority=300, udp, udp_dst=%d, actions=drop", vxlanPort)
 	// Table 20: from OpenShift container; validate IP/MAC, assign tenant-id; filled in by setupPodFlows
 	// eg, "table=20, priority=100, in_port=${ovs_port}, arp, nw_src=${ipaddr}, arp_sha=${macaddr}, actions=load:${tenant_id}->NXM_NX_REG0[], goto_table:21"
 	//     "table=20, priority=100, in_port=${ovs_port}, ip, nw_src=${ipaddr}, actions=load:${tenant_id}->NXM_NX_REG0[], goto_table:21"
```
danwinship, line 135: put the new AddFlow after the comment introducing table 20

danwinship (commented):
just two nitpicks

danwinship:
/approve
(you can get someone else to `/lgtm` after you fix the above if I'm not around)

pkg/network/node/ovscontroller.go
alexanderConstantinescu: done

knobunc:
/override ci/prow/e2e-aws-upgrade

This job was broken by the rebase and will not work until https://github.com/openshift/cluster-network-operator/pull/1118 ships in stable

PR #305, opened 2021-05-22
Bug 1964059: rebase to latest sdn-4.8-kubernetes-1.21.1

danwinship:
rebase to latest sdn-4.8-kubernetes-1.21.1, including https://github.com/openshift/kubernetes/pull/762 (windows kube-proxy fix) and https://github.com/openshift/kubernetes/pull/764 (fixing kube-proxy hack for local dns)

/cc @aravindhp @Miciah @aojea


vendor/k8s.io/kubernetes/cmd/kube-proxy/app/server_others.go
```diff
@@ -364,6 +364,12 @@ func newProxyServer(
 		}
 	}
 
+	useEndpointSlices := utilfeature.DefaultFeatureGate.Enabled(features.EndpointSliceProxying)
+	if proxyMode == proxyModeUserspace {
+		// userspace mode doesn't support endpointslice.
+		useEndpointSlices = false
```
aojea, line 370: is this consistent with your last PR?

danwinship: 1. Note that we have not _merged_ any PRs that affect endpointslice handling.
2. If you mean "will it cause any conflicts with #296", then no, because `cmd/kube-proxy/` is only used by the actual kube-proxy binary, not openshift-sdn's embedded copy of kube-proxy.
3. If you mean "will it cause inconsistency between standalone kube-proxy's behavior and embedded kube-proxy's behavior", then that's not a relevant question. Currently standalone kube-proxy in 4.8 in userspace mode doesn't work, and this fixes it. Embedded kube-proxy in 4.8 in userspace mode already worked (and still works after #295) and this change doesn't affect it.

danwinship:
/retitle Bug 1964059: rebase to latest sdn-4.8-kubernetes-1.21.1

(un-stealing WMCO bug since apparently they have their own distinct vendored copy of kube-proxy)

vendor/k8s.io/kubernetes/cmd/kube-proxy/app/server_others.go
aojea: it was about 2. :)

PR #311, opened 2021-06-07
Bug 1948436: remove vxlan_monitor and OVS packet stat parsing

alexanderConstantinescu:
Going back to the first draft of #291 where we remove the OVS packet parsing and `vxlan_monitor.go`, but keep the polling as it was. This is not optimal since it results in code duplication between master and node for what concerns monitoring egress nodes, it however introduces less code change and hence reduces the risk of introducing other regressions this close to code freeze. 

#291 is kept for the next release cycle as to gain time to think of a better solution, assess the risks introduced by generalizing the code and track the discussion and concerns this far. 

/assign @danwinship


pkg/network/node/egressip.go
```diff
@@ -136,11 +149,88 @@ func (eip *egressIPWatcher) ReleaseEgressIP(egressIP, nodeIP string) {
...
 		}
-	} else if eip.vxlanMonitor != nil {
-		eip.vxlanMonitor.RemoveEgressIP(nodeIP, egressIP)
+	} else {
+		eip.removeEgressIP(nodeIP, egressIP)
 	}
 }
 
+func (eip *egressIPWatcher) addEgressIP(nodeIP, egressIP string) {
+	if eip.monitorNodes[nodeIP] != nil {
```
danwinship, line 158: `monitorNodes` still needs a mutex like in `egressVXLANMonitor`, because `addEgressIP` / `removeEgressIP` modify it from one thread, and `poll` iterates it from another thread.

pkg/network/node/egressip.go
```diff
@@ -136,11 +149,88 @@ func (eip *egressIPWatcher) ReleaseEgressIP(egressIP, nodeIP string) {
...
+		klog.V(4).Infof("Unmonitoring node %s", nodeIP)
+		delete(eip.monitorNodes, nodeIP)
+		if len(eip.monitorNodes) == 0 && eip.stop != nil {
+			close(eip.stop)
+			eip.stop = nil
+		}
+	}
+}
+
+func (eip *egressIPWatcher) poll() (bool, error) {
```
danwinship, line 189: in `egressVXLANMonitor` we hold the evm mutex around all of `poll`, which is bad because of the sleep-and-retry... it would be better to only hold it around `check`. (Although that's also not ideal now, since doing all the pings could take a while... but I guess `addNodeIP` and `removeNodeIP` don't get called that often, so there won't be that many situations where they have to wait for a `check` to finish and release the mutex, so we can survive with this for now.)

pkg/network/node/egressip.go
alexanderConstantinescu: I am not convinced it does. `addEgressIP` / `removeEgressIP` only modifies the `.egressIPs` field. `check` modifies `.retries` and `.offline`, I thus fail to see how they can "step on each others toes". The only granted thread-un-safe part as I see it is: `addEgressIP` / `removeEgressIP`  adding or removing a node to `.monitorNodes` while `check` is checking...but `check` should pick up the new data in its next polling cycle, so I don't think it's of a great concern.

In any case, I added a mutex around all of this. However as you mentioned in your other comment we will lock for longer duration now. I also need to wrap specific parts of `check` in the mutex since `SetNodeOffline` percolates back to `addEgressIP` / `removeEgressIP` and would have dead-locked otherwise.

danwinship: > I thus fail to see how they can "step on each others toes"

golang's built-in data structures are not thread-safe. If one thread modifies the map while another thread is reading it, then anything is possible, including "the program crashes because the reading thread ends up dereferencing a no-longer-valid pointer". (This is not made terribly obvious in the golang docs and tends to trip up everyone at least once.)

> I also need to wrap specific parts of `check` in the mutex since `SetNodeOffline` percolates back to `addEgressIP` / `removeEgressIP` and would have dead-locked otherwise.

The other possibility would be to make it more like how it was in the vxlanmonitor version, where instead of doing each update from within the loop, you just keep an array of nodes that will need to be updated after the loop...

Or maybe use a `sync.Map` instead of a `map`

alexanderConstantinescu: > The other possibility would be to make it more like how it was in the vxlanmonitor version, where instead of doing each update from within the loop, you just keep an array of nodes that will need to be updated after the loop...

Yeah, indeed one of the options I had as a resolution to this problem was "just keep vxlan_monitor and turn it into a ping_monitor" instead. 

> Or maybe use a sync.Map instead of a map

I studied that option, but `sync.Map`'s `.Range` function is only atomically thread-safe on the item it's currently looping over - not over the entire map. And since `check` needs to loop all items in the map, it won't really solve our problem here more than the current solution does.

alexanderConstantinescu: > The other possibility would be to make it more like how it was in the vxlanmonitor version, where instead of doing each update from within the loop, you just keep an array of nodes that will need to be updated after the loop...

By the way, this is almost what the current version of the code does (in case you didn't have a look yet - I pushed a couple of hours ago)...but I am using a map instead of a slice, but that shouldn't change anything here.

knobunc:
/override ci/prow/e2e-aws-upgrade

This job was broken by the rebase and will not work until openshift/cluster-network-operator#1118 ships in stable

PR #316, opened 2021-06-14
Bug 1942122: when assigning and releasing egressIP try more than once before failing

JacobTanenbaum:
There may be something else on the system running iptables and it
can circumvent "Resource temporarily not avalible" errors to retry
running iptables commands

### DIFF
--- a/pkg/network/node/iptables.go
+++ b/pkg/network/node/iptables.go
@@ -12,7 +12,9 @@ import (
 
 	"k8s.io/klog/v2"
 
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/kubernetes/pkg/util/iptables"
+	utilexec "k8s.io/utils/exec"
 )
 
 type NodeIPTables struct {
@@ -25,6 +27,39 @@ type NodeIPTables struct {
 	mu sync.Mutex // Protects concurrent access to syncIPTableRules()
 
 	egressIPs map[string]string
+}
+
+// this will retry 10 times over a period of 13 seconds
+var iptablesBackoff wait.Backoff = wait.Backoff{
+	Duration: 500 * time.Millisecond,
+	Factor:   1.25,
+	Steps:    10,
+}
+
+// execIPTablesWithRetry allows a simple way to retry iptables commands if they fail the first time
+func execIPTablesWithRetry(f func() error) error {
+	return wait.ExponentialBackoff(iptablesBackoff, func() (bool, error) {
+		if err := f(); err != nil {
+			if isResourceError(err) {
+				klog.V(5).Infof("Call to iptables failed with transient failure: %v", err)
+				return false, nil
+			}
+			return false, err
+		}
+		return true, nil
+	})
+}
+
+const iptablesStatusResourceProblem = 4
+
+// isResourceError returns true if the error indicates that iptables ran into a "resource
+// problem" and was unable to attempt the request. In particular, this will be true if it
+// times out trying to get the iptables lock.
+func isResourceError(err error) bool {
+	if ee, isExitError := err.(utilexec.ExitError); isExitError {
+		return ee.ExitStatus() == iptablesStatusResourceProblem
+	}
+	return false
 }
 
 func newNodeIPTables(ipt iptables.Interface, clusterNetworkCIDR []string, masqueradeServices bool, vxlanPort uint32, masqueradeBit uint32) *NodeIPTables {
@@ -57,7 +92,12 @@ type Chain struct {
 func (n *NodeIPTables) addChainRules(chain Chain) (bool, error) {
 	allExisted := true
 	for _, rule := range chain.rules {
-		existed, err := n.ipt.EnsureRule(iptables.Append, iptables.Table(chain.table), iptables.Chain(chain.name), rule...)
+		var existed bool
+		err := execIPTablesWithRetry(func() error {
+			var err error
+			existed, err = n.ipt.EnsureRule(iptables.Append, iptables.Table(chain.table), iptables.Chain(chain.name), rule...)
+			return err
+		})
 		if err != nil {
 			return false, fmt.Errorf("failed to ensure rule %v exists: %v", rule, err)
 		}
@@ -84,7 +124,12 @@ func (n *NodeIPTables) syncIPTableRules() error {
 	for i := len(chains) - 1; i >= 0; i-- {
 		chain := chains[i]
 		// Create chain if it does not already exist
-		chainExisted, err := n.ipt.EnsureChain(iptables.Table(chain.table), iptables.Chain(chain.name))
+		var chainExisted bool
+		err := execIPTablesWithRetry(func() error {
+			var err error
+			chainExisted, err = n.ipt.EnsureChain(iptables.Table(chain.table), iptables.Chain(chain.name))
+			return err
+		})
 		if err != nil {
 			return fmt.Errorf("failed to ensure chain %s exists: %v", chain.name, err)
 		}
@@ -94,7 +139,10 @@ func (n *NodeIPTables) syncIPTableRules() error {
 			// chains with the same table and srcChain (ie, OPENSHIFT-FIREWALL-FORWARD
 			// and OPENSHIFT-ADMIN-OUTPUT-RULES) will run in the same order as they
 			// appear in getNodeIPTablesChains().
-			_, err = n.ipt.EnsureRule(iptables.Prepend, iptables.Table(chain.table), iptables.Chain(chain.srcChain), append(chain.srcRule, "-j", chain.name)...)
+			err = execIPTablesWithRetry(func() error {
+				_, err = n.ipt.EnsureRule(iptables.Prepend, iptables.Table(chain.table), iptables.Chain(chain.srcChain), append(chain.srcRule, "-j", chain.name)...)
+				return err
+			})
 			if err != nil {
 				return fmt.Errorf("failed to ensure rule from %s to %s exists: %v", chain.srcChain, chain.name, err)
 			}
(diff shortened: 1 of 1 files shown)


pkg/network/node/egressip.go
```diff
@@ -136,7 +136,12 @@ func (eip *egressIPWatcher) ClaimEgressIP(vnid uint32, egressIP, nodeIP string)
 	if nodeIP == eip.localIP {
 		mark := getMarkForVNID(vnid, eip.masqueradeBit)
 		eip.iptablesMark[egressIP] = mark
-		if err := eip.assignEgressIP(egressIP, mark); err != nil {
+		if err := utilwait.PollImmediate(10*time.Millisecond, 2*time.Second, func() (bool, error) {
+			if err := eip.assignEgressIP(egressIP, mark); err != nil {
+				return false, err
```
alexanderConstantinescu, line 141: This will not do what you want it to. You need to `return false, nil`. From the doc of `PollImmediate`:

```
PollImmediate tries a condition func until it returns true, an error, or the timeout is reached.
```

Hence if we need to retry we can't return an error when we experience one. 

Moreover, maybe do the same for `releaseEgressIP`? If this is a problem for assign, then it might also be for remove. 

And third: maybe log the error inside `PollImmediate` so that we know what fails every time something is tried, instead of just the final timeout error.

alexanderConstantinescu (changes requested):

pkg/network/node/egressip.go
JacobTanenbaum: done

alexanderConstantinescu (commented):
Actually come to think of it, if failing iptables executions is really a problem: we could just wrap the `ipt iptables.Interface` - which we use in our code - with our own interface which uses a retry for all iptables calls? That would make sure all iptables calls in the code are covered by a retry.

JacobTanenbaum:
failed to require lease 

/retest

pkg/network/node/iptables.go
```diff
@@ -109,7 +140,11 @@ func (n *NodeIPTables) syncIPTableRules() error {
 			// Chain existed but not with the expected rules; this probably means
 			// it contained rules referring to a *different* subnet; flush them
 			// and try again.
-			if err = n.ipt.FlushChain(iptables.Table(chain.table), iptables.Chain(chain.name)); err != nil {
+			err = execIPTablesWithRetry(func() error {
+				err := n.ipt.FlushChain(iptables.Table(chain.table), iptables.Chain(chain.name))
```
alexanderConstantinescu, line 144: nit; can be `return n.ipt.FlushChain(iptables.Table(chain.table), iptables.Chain(chain.name))`

pkg/network/node/iptables.go
```diff
@@ -254,12 +295,19 @@ func (n *NodeIPTables) DeleteEgressIPRules(egressIP, mark string) error {
 	delete(n.egressIPs, egressIP)
 
 	for _, cidr := range n.clusterNetworkCIDR {
-		err := n.ipt.DeleteRule(iptables.TableNAT, iptables.Chain("OPENSHIFT-MASQUERADE"), "-s", cidr, "-m", "mark", "--mark", mark, "-j", "SNAT", "--to-source", egressIP)
+		err := execIPTablesWithRetry(func() error {
+			err := n.ipt.DeleteRule(iptables.TableNAT, iptables.Chain("OPENSHIFT-MASQUERADE"), "-s", cidr, "-m", "mark", "--mark", mark, "-j", "SNAT", "--to-source", egressIP)
```
alexanderConstantinescu, line 299: nit: can be `return n.ipt.DeleteRule(iptables.TableNAT, iptables.Chain("OPENSHIFT-MASQUERADE"), "-s", cidr, "-m", "mark", "--mark", mark, "-j", "SNAT", "--to-source", egressIP)`

pkg/network/node/iptables.go
```diff
@@ -254,12 +295,19 @@ func (n *NodeIPTables) DeleteEgressIPRules(egressIP, mark string) error {
...
+			return err
+		})
 		if err != nil {
 			return err
 		}
 	}
-	return n.ipt.DeleteRule(iptables.TableFilter, iptables.Chain("OPENSHIFT-FIREWALL-ALLOW"), "-d", egressIP, "-m", "conntrack", "--ctstate", "NEW", "-j", "REJECT")
+	err := execIPTablesWithRetry(func() error {
+		err := n.ipt.DeleteRule(iptables.TableFilter, iptables.Chain("OPENSHIFT-FIREWALL-ALLOW"), "-d", egressIP, "-m", "conntrack", "--ctstate", "NEW", "-j", "REJECT")
```
alexanderConstantinescu, line 307: nit: can be `return n.ipt.DeleteRule(iptables.TableFilter, iptables.Chain("OPENSHIFT-FIREWALL-ALLOW"), "-d", egressIP, "-m", "conntrack", "--ctstate", "NEW", "-j", "REJECT")`

pkg/network/node/iptables.go
```diff
@@ -305,7 +353,10 @@ func (n *NodeIPTables) SyncEgressIPRules() {
 			continue
 		}
 		args = args[2:]
-		err := n.ipt.DeleteRule(iptables.TableNAT, iptables.Chain("OPENSHIFT-MASQUERADE"), args...)
+		err := execIPTablesWithRetry(func() error {
+			err := n.ipt.DeleteRule(iptables.TableNAT, iptables.Chain("OPENSHIFT-MASQUERADE"), args...)
```
alexanderConstantinescu, line 357: same nit as above

pkg/network/node/iptables.go
```diff
@@ -319,7 +370,10 @@ func (n *NodeIPTables) SyncEgressIPRules() {
 			continue
 		}
 		args = args[2:]
-		err := n.ipt.DeleteRule(iptables.TableFilter, iptables.Chain("OPENSHIFT-FIREWALL-ALLOW"), args...)
+		err := execIPTablesWithRetry(func() error {
+			err := n.ipt.DeleteRule(iptables.TableFilter, iptables.Chain("OPENSHIFT-FIREWALL-ALLOW"), args...)
```
alexanderConstantinescu, line 374: same nit as above

alexanderConstantinescu (approved):
Looks good now, just had a few nit comments, but I am not that rigid about it either

/lgtm

danwinship:
/retitle Bug 1942122: when assigning and releasing egressIP try more than once before failing

pkg/network/node/iptables.go
```diff
@@ -27,6 +28,23 @@ type NodeIPTables struct {
 	egressIPs map[string]string
 }
 
+var iptablesBackoff wait.Backoff = wait.Backoff{
+	Duration: 500 * time.Millisecond,
+	Factor:   1.25,
+	Steps:    10,
+}
```
danwinship, line 35: assuming I did the math right, this will wait a total of about 13 seconds. Worth a comment

pkg/network/node/iptables.go
```diff
@@ -27,6 +28,23 @@ type NodeIPTables struct {
...
+
+// execIPTablesWithRetry allows a simple way to retry iptables commands if they fail the first time
+func execIPTablesWithRetry(f func() error) error {
+	return wait.ExponentialBackoff(iptablesBackoff, func() (bool, error) {
+		if err := f(); err != nil {
+			klog.V(5).Infof("Call to iptables failed: %v", err)
+			return false, nil
```
danwinship, line 42: This will retry on _any_ iptables failure, but we really only want to retry on "temporary"/"transient" failures; make a copy of `isResourceError` from `vendor/k8s.io/kubernetes/pkg/util/iptables/iptables.go`

pkg/network/node/iptables.go
JacobTanenbaum: done

pkg/network/node/iptables.go
JacobTanenbaum: done

pkg/network/node/iptables.go
JacobTanenbaum: done

pkg/network/node/iptables.go
JacobTanenbaum: done

pkg/network/node/iptables.go
JacobTanenbaum: done

pkg/network/node/iptables.go
JacobTanenbaum: done

pkg/network/node/iptables.go
JacobTanenbaum: done

JacobTanenbaum:
ci/prow/e2e-gcp - not enough resources available to acquire a lease 

/retest

pkg/network/node/iptables.go
```diff
@@ -27,6 +29,39 @@ type NodeIPTables struct {
 	egressIPs map[string]string
 }
 
+// this could potentially add 13 to an iptables call, this should happen rarely
```
danwinship, line 32: missing "seconds"
and it doesn't have to be phrased like a warning, I was just saying, often when you have a retry like this, you want to be able to figure out how long it's going to be retrying for, and that's hard to do with exponential backoff, so it's helpful to have a comment like "this will retry 10 times over a period of about 13 seconds" to clarify

danwinship (approved):
looks good, just a comment nitpick

pkg/network/node/iptables.go
JacobTanenbaum: done

palonsoro:
//cherry-pick release-4.8

PR #320, opened 2021-06-16
Bug 1971808: fix local-with-fallback

danwinship:
The new `local-with-fallback` hack was interfering with cluster-internal traffic to local-with-fallback services (which already would have worked fine even when there were no local pods), breaking cluster-internal source IP preservation (and maybe other things).

This reverts the previous hack and replaces it with a hack that preserves almost all of the current handling of `externalTrafficPolicy: Local` services, and only changes the bit at the end. In particular, this preserves the short-circuit of pod-to-loadbalancer-IP for Local services, which the previous hack did not.

/cc @Miciah @smarterclayton @sgreene570 

rebases to https://github.com/openshift/kubernetes/pull/813


danwinship:
ok, this seems to work AFAICT:
- The router responds via its NodePort on all node IPs, even the nodes it doesn't have an endpoint on
- If I create an `externalTrafficPolicy: Local` NodePort service without the annotation, it only responds on the host that holds the endpoint. But if I then add the `local-with-fallback` annotation, it then responds on all nodes.
- Following the reproducer from bz 1971808, the service gives the correct source IP from any node

danwinship:
Whatever. GCP failures are infrastructural, and I was testing on GCP when I wrote https://github.com/openshift/sdn/pull/320#issuecomment-864123385 so I know it would work if the cluster actually came up.

PR #324, opened 2021-07-26
Bug 1985336: Disable conntrack for vxlan traffic.

palonsoro:
It is not needed because vxlan always uses the same destination port (4789 by default) for every packet.
Keeping it can be a source of performance issues on large clusters.

Fixes [BZ#1985336](https://bugzilla.redhat.com/show_bug.cgi?id=1985336)

### DIFF
--- a/pkg/network/node/iptables.go
+++ b/pkg/network/node/iptables.go
@@ -273,6 +273,28 @@ func (n *NodeIPTables) getNodeIPTablesChains() []Chain {
 		},
 	)
 
+	// Don't track vxlan traffic with conntrack, as it is not needed and decreases performance
+	chainArray = append(chainArray,
+		Chain{
+			table:    "raw",
+			name:     "OPENSHIFT-VXLAN-NOTRACK",
+			srcChain: "OUTPUT",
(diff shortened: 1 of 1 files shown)


pkg/network/node/iptables.go
```diff
@@ -273,6 +273,28 @@ func (n *NodeIPTables) getNodeIPTablesChains() []Chain {
 		},
 	)
 
+	// Don't track vxlan traffic with conntrack, as it is not needed and decreases performance
+	chainArray = append(chainArray,
+		Chain{
+			table:    "raw",
+			name:     "OPENSHIFT-VXLAN-NOTRACK",
```
danwinship, line 280: assuming we do this, I'd make the table be `OPENSHIFT-NOTRACK` or even `OPENSHIFT-RAW-PROCESSING` or something like that (I don't really like "processing"). Something slightly more generic, in case we want to use it for more things later.

danwinship (commented):
I kind of feel like if this is correct then OVS or the kernel ought to do this itself? When _would_ you want conntracking of VXLAN?

PR #330, opened 2021-08-06
Bug 1980127: recopy openshift-sdn binary and cni config if they are deleted

JacobTanenbaum:
we cannot depend on the daemonset controller to seriealize pod
creation and deletion, because of this we need to verify if the CNI
config and the openshift-sdn binaries exist or have been deleted by the
previous terminating pod

Signed-off-by: Jacob Tanenbaum <jtanenba@redhat.com>


danwinship:
This bug was only seen on 3.11 wasn't it?
I would lean toward assuming that it is only broken there.

PR #331, opened 2021-08-09
Bug 1991565: misc cleanup

danwinship:
openshift-sdn has been building up low-level tech debt because we're like "well it's eventually going away", but we're still going to be maintaining it for a while so let's make it prettier.

some of these commits are large but all of them are boring

/cc @squeed


squeed:
nice stuff.
/lgtm

PR #332, opened 2021-08-09
Bug 1991551: update usage of Events for 1.22 rebase

danwinship:
The new `k8s.io/client-go/tools/events` version of `Eventf()` sneakily added a new argument before the trailing message+args, so we're ending up with events with

    Action:"The service-port %s:%s needs pods."
    Note:"test-service%!(EXTRA string=http)"

This adds an "action" to each event. The documentation is not very useful, but this seems consistent with how it is used in some places in k/k...

----

Additionally, it turns out that openshift-controller-manager is not compatible with openshift-sdn using the new Event API for unidling, because it assumes the `LastTimestamp` field of the event will be set, but this field is deprecated in `eventsv1.Event`, and not set by the convenience APIs. Thus, for now, revert back to using the legacy API for the unidling proxy.

### DIFF
--- a/pkg/cmd/openshift-sdn-node/proxy.go
+++ b/pkg/cmd/openshift-sdn-node/proxy.go
@@ -17,7 +17,9 @@ import (
 	"k8s.io/apiserver/pkg/server/routes"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	"k8s.io/client-go/kubernetes/scheme"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/events"
+	"k8s.io/client-go/tools/record"
 	"k8s.io/component-base/metrics/legacyregistry"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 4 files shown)


pkg/cmd/openshift-sdn-node/proxy.go
```diff
@@ -180,7 +182,15 @@ func (sdn *openShiftSDN) runProxy(waitChan chan<- bool) {
 	}
 
 	if enableUnidling {
-		signaler := unidler.NewEventSignaler(recorder)
+		// FIXME: openshift-controller-manager assumes the LastTimestamp field in
```
astoycos, line 185: Is there some way to track the fix we need to make in `openshift-controller-manager` for this?

astoycos:
/lgtm

I just wan't to make sure we actually remember to go back and fix the `openshift-controller-manager`

pkg/cmd/openshift-sdn-node/proxy.go
danwinship: https://bugzilla.redhat.com/show_bug.cgi?id=1996140

PR #342, opened 2021-09-03
Bug 1966521: proxy: don't re-check every userspace proxy rule on every change

danwinship:
The userspace proxy re-checks every one of its rules every time
anything changes, to ensure that rules accidentally deleted by
iptables or firewalld restart get recreated promptly. But this sucks
up a ton of CPU when there are lots of idled services and lots of
changes to other (unrelated) services.

Fix it to only do the full re-check in response to openshift-sdn's
iptables.Monitor being triggered, rather than on every sync.


aojea:
does it work?

danwinship:
> does it work?

yes, it drastically cuts down on calls to `iptables` when lots of services are idled
/hold cancel

PR #345, opened 2021-09-07
remove a bunch of dead/useless code

danwinship:
Stuff noticed while reviewing the codebase...

` 568 files changed, 37 insertions(+), 195479 deletions(-)`

(ok, fine, it's `11 files changed, 36 insertions(+), 136 deletions(-)` if you ignore `vendor/`, but still!)


pkg/network/common/validation.go
```diff
@@ -140,8 +140,8 @@ func ValidateHostSubnet(hs *osdnv1.HostSubnet) error {
 			allErrs = append(allErrs, field.Invalid(field.NewPath("subnet"), hs.Subnet, err.Error()))
 		}
 	}
-	// In theory this has to be IPv4, but it's possible some clusters might be limping along with IPv6 values?
-	if net.ParseIP(hs.HostIP) == nil {
+	_, err := validateIPv4(hs.HostIP)
```
alexanderConstantinescu, line 143: > But in OCP 4.x, the HostSubnet CRD requires the field to be IPv4.

Nit: do we need to validate it then? There shouldn't really be any case where that is not already upheld by the CRD validator (well, unless someone removes it or writes an incorrect regular expression when trying to match IPv4....yeah, maybe let's keep it)

alexanderConstantinescu (approved):
/lgtm

PR #347, opened 2021-09-13
Bug 2002290: [4.8] proxy: don't re-check every userspace proxy rule on every change

danwinship:
backport of #342

/hold
because it depends on https://github.com/openshift/kubernetes/pull/952


hardys:
[patch-manager] Approving for cherry-pick with a score of 2.00

abhat:
@zhaozhanqi we need to get this PR verified so the 4.7 bug can move forward.

PR #348, opened 2021-09-13
Bug 2002291: [4.7] proxy: don't re-check every userspace proxy rule on every change

danwinship:
backport of #342

/hold
because it depends on openshift/kubernetes#953


dhellmann:
[patch-manager] :rocket: Approved for z-stream by score: 1.00

picked

PR #349, opened 2021-09-14
Bug 2003859: Fix up event recorder usage (again)

danwinship:
pkg/cmd/openshift-sdn-node creates separate recorders for the node and
the proxy, and only the proxy one got updated to the new API, but the
callers of the node one also got (partially) updated to use the new
API even though they were still using an old recorder, resulting in
events like:

    "message": "Starting%!(EXTRA string=openshift-sdn done initializing node networking.)",


danwinship:
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_sdn/349/pull-ci-openshift-sdn-master-e2e-aws/1437804170076753920/artifacts/e2e-aws/gather-extra/artifacts/events.json:

            "message": "openshift-sdn done initializing node networking.",

PR #352, opened 2021-09-15
resync proxy startup code with upstream

danwinship:
Our proxy startup code is a partial fork of the upstream kube-proxy startup code, but there have been a bunch of changes upstream that we haven't totally kept in sync with. This reorganizes the code to look more like the upstream code and then syncs us with the latest version (which, as it turns out, doesn't change all that much, but at least now we know, and it will be easier to update the next time something _does_ change).

This also updates our LocalTrafficDetector to just check the source interface of the traffic rather than the cluster CIDR, thus fixing the problem with multiple cluster CIDRs.

/assign @tssurya 
since you looked at the LocalTrafficDetector stuff before and this contains a (very simple, openshift-sdn-specific) version of [SDN-1303](https://issues.redhat.com/browse/SDN-1303)


tssurya:
Thanks Dan, will review this, meanwhile I've been meaning to get to this https://github.com/kubernetes/kubernetes/pull/95400 as well, keeps jumping down on my prio list.

I guess what you have done is an sdn specifc detector for allowing traffic via match on source interface for tun0. Nice! I'll work on my upstream PR as well this week and try and get that in.

Will leave a review tomorrow.

pkg/cmd/openshift-sdn-node/proxy.go
```diff
@@ -193,91 +86,55 @@ func (sdn *openShiftSDN) runProxy(waitChan chan<- bool) {
...
+func (d *sdnLocalDetector) IsImplemented() bool {
+	return true
 }
 
-func (sdn *openShiftSDN) startMetricsServer() {
-	if sdn.proxyConfig.MetricsBindAddress == "" {
-		return
-	}
+func (d *sdnLocalDetector) JumpIfLocal(args []string, toChain string) []string {
+	return append(args, "-i", sdnnode.Tun0, "-j", toChain)
```
tssurya, line 131: ++ looks good. It's easier since in SDN the interface would always be `tun0`

pkg/cmd/openshift-sdn-node/kube_proxy.go
```diff
@@ -119,11 +118,7 @@ func newProxyServer(config *kubeproxyconfig.KubeProxyConfiguration, client clien
 			return nil, fmt.Errorf("unable to read IPTables MasqueradeBit from config")
 		}
 
-		var localDetector proxyutiliptables.LocalTrafficDetector
-		localDetector, err = getLocalDetector(config, iptInterface)
-		if err != nil {
-			return nil, fmt.Errorf("unable to create proxier: %v", err)
-		}
+		localDetector := getLocalDetector()
```
tssurya, line 126: ok so we are going to by default just use the interface detector for SDN and that way we don't have to even check if CIDRs are multiple or single.

q: This would be something that's done on startup right? how would it look during upgrades? The iptables rules we add for the previous version of cidr detector gets removed and re-added as these new rules? I don't think we have code to remove anything, so I am assuming upon restart of sdn pods on upgrades, the rules are all recreated freshly?

tssurya: or maybe for existing services it would still be the old rules and newer services it would be the new rules?

tssurya:
The upgrade job hasn't passed, waiting for that.
All other commits look good, mostly involve refactoring SDN code to reflect upstream code.
Note: I didn't fully understand the last commit where we are using a second informer since I am not very familiar with multi-tenant mode so lgtm-ing that commit with half-knowledge.

Checked the new sdn local detector rules in a cluster-bot. Looking good:
```
-A KUBE-SVC-DZZGCZT3USY56SM6 -d 172.30.137.149/32 ! -i tun0 -p tcp -m comment --comment "openshift-ingress-operator/metrics:metrics cluster IP" -m tcp --dport 9393 -j KUBE-MARK-MASQ
-A KUBE-SVC-CIUYVLZDADCHPTYT -d 172.30.152.178/32 ! -i tun0 -p tcp -m comment --comment "openshift-apiserver-operator/metrics:https cluster IP" -m tcp --dport 443 -j KUBE-MARK-MASQ
-A KUBE-SVC-LR44LCGLBA5H46DK -d 172.30.78.132/32 ! -i tun0 -p tcp -m comment --comment "openshift-cluster-version/cluster-version-operator:metrics cluster IP" -m tcp --dport 9099 -j KUBE-MARK-MASQ
-A KUBE-SVC-PFY2VR2AT5VQM74G -d 172.30.4.83/32 ! -i tun0 -p tcp -m comment --comment "openshift-machine-config-operator/machine-config-daemon:metrics cluster IP" -m tcp --dport 9001 -j KUBE-MARK-MASQ
-A KUBE-SVC-GGV3SPGNRULALRSD -d 172.30.199.204/32 ! -i tun0 -p tcp -m comment --comment "openshift-machine-api/cluster-autoscaler-operator:metrics cluster IP" -m tcp --dport 9192 -j KUBE-MARK-MASQ
-A KUBE-SVC-NM6OF7LZYCSWPYSN -d 172.30.183.179/32 ! -i tun0 -p tcp -m comment --comment "openshift-apiserver/api:https cluster IP" -m tcp --dport 443 -j KUBE-MARK-MASQ
-A KUBE-SVC-HWYXEEIGDEK65VFZ -d 172.30.227.107/32 ! -i tun0 -p tcp -m comment --comment "openshift-multus/multus-admission-controller:metrics cluster IP" -m tcp --dport 8443 -j KUBE-MARK-MASQ
-A KUBE-SVC-3ZP2SBSMEIDDD7TA -d 172.30.23.74/32 ! -i tun0 -p tcp -m comment --comment "openshift-apiserver/check-endpoints:check-endpoints cluster IP" -m tcp --dport 17698 -j KUBE-MARK-MASQ

```

Will update the interface detector PR upstream as well. Since we don't have any other detectors except CIDR which is useless in case of multiple CIDRs, we should be good to use the interface detector by default.

pkg/cmd/openshift-sdn-node/kube_proxy.go
danwinship: the kube-proxy code overwrites the previous service rules every time it updates

danwinship:
> The upgrade job hasn't passed on this PR even once, waiting for that.

looks like it currently has about a 50% pass rate across PRs in this repo: https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-openshift-sdn-master-e2e-aws-upgrade

>  Note: I didn't fully understand the last commit where we are using a second informer since I am not very familiar with multi-tenant mode so lgtm-ing that commit with half-knowledge.

The issue is that the kube-proxy code listens for Service events, and the multitenant code also listens for Service events. And previously they both used the same Informer, but that was wrong because the kube-proxy code only wants a filtered set of Services, but the multitenant code wants all Services. So now the multitenant code will get an unmodified Informer, and the kube-proxy code gets its own special informer.

PR #359, opened 2021-10-12
Remove locking from EgressIPTracker.Ping

Kyrtapz:
This change addresses the following deadlock scenario:
1. `getOfflineResult` tries to ping the nodes, it locks `monitorNodesLock` at the start.
2. In the meantime there is an event on `HostSubnet`, `eit` is locked and there is a change to `monitorNodes` and the code waits for `monitorNodesLock` that is blocked by `getOfflineResult`.
3. `getOfflineResult` calls Ping which calls `lookupNodeIP` which tries to lock `eit` resulting in a deadlock.

Instead of looking for the `sdnIP` each time Ping is called it is added directly in `ClaimEgressIP`.

```
goroutine 452 [semacquire, 1472 minutes]:
sync.runtime_SemacquireMutex(0xc0005da544, 0x4e2a00, 0x1)
	runtime/sema.go:71 +0x47
sync.(*Mutex).lockSlow(0xc0005da540)
	sync/mutex.go:138 +0x105
sync.(*Mutex).Lock(...)
	sync/mutex.go:81
github.com/openshift/sdn/pkg/network/common.(*EgressIPTracker).UpdateNetNamespaceEgress(0xc0005da540, 0xc001322000)
	github.com/openshift/sdn/pkg/network/common/egressip.go:313 +0x766
github.com/openshift/sdn/pkg/network/common.(*EgressIPTracker).handleAddOrUpdateNetNamespace(0xc0005da540, 0x1f8aca0, 0xc001322000, 0x1f8aca0, 0xc000c039a8, 0x1fba85d, 0x8)
	github.com/openshift/sdn/pkg/network/common/egressip.go:302 +0x175
github.com/openshift/sdn/pkg/network/common.InformerFuncs.func2(0x1f8aca0, 0xc000c039a8, 0x1f8aca0, 0xc001322000)
	github.com/openshift/sdn/pkg/network/common/informers.go:23 +0x62
k8s.io/client-go/tools/cache.ResourceEventHandlerFuncs.OnUpdate(...)
	k8s.io/client-go@v1.21.1/tools/cache/controller.go:238
k8s.io/client-go/tools/cache.(*processorListener).run.func1()
	k8s.io/client-go@v1.21.1/tools/cache/shared_informer.go:775 +0x1c5
k8s.io/apimachinery/pkg/util/wait.BackoffUntil.func1(0xc000a5ef60)
	k8s.io/apimachinery@v1.21.1/pkg/util/wait/wait.go:155 +0x5f
k8s.io/apimachinery/pkg/util/wait.BackoffUntil(0xc0006dbf60, 0x22641c0, 0xc0006ad080, 0x1cb9c01, 0xc00044c120)
	k8s.io/apimachinery@v1.21.1/pkg/util/wait/wait.go:156 +0x9b
k8s.io/apimachinery/pkg/util/wait.JitterUntil(0xc000a5ef60, 0x3b9aca00, 0x0, 0xc000a5ef01, 0xc00044c120)
	k8s.io/apimachinery@v1.21.1/pkg/util/wait/wait.go:133 +0x98
k8s.io/apimachinery/pkg/util/wait.Until(...)
	k8s.io/apimachinery@v1.21.1/pkg/util/wait/wait.go:90
k8s.io/client-go/tools/cache.(*processorListener).run(0xc000571380)
	k8s.io/client-go@v1.21.1/tools/cache/shared_informer.go:771 +0x95
k8s.io/apimachinery/pkg/util/wait.(*Group).Start.func1(0xc0002a4b50, 0xc0005c9df0)
	k8s.io/apimachinery@v1.21.1/pkg/util/wait/wait.go:73 +0x51
created by k8s.io/apimachinery/pkg/util/wait.(*Group).Start
	k8s.io/apimachinery@v1.21.1/pkg/util/wait/wait.go:71 +0x65
```


alexanderConstantinescu:
/retitle Bug 2008987: Remove locking from EgressIPTracker.Ping

pkg/network/common/egressip.go
```diff
@@ -211,8 +211,9 @@ func (eit *EgressIPTracker) UpdateHostSubnetEgress(hs *osdnv1.HostSubnet) {
 		_, cidr, err := net.ParseCIDR(hs.Subnet)
 		if err != nil {
 			klog.Errorf("Could not parse HostSubnet %q CIDR: %v", hs.Name, err)
+		} else {
+			sdnIP = GenerateDefaultGateway(cidr).String()
 		}
-		sdnIP = GenerateDefaultGateway(cidr).String()
```
danwinship, line 215: We should ignore the HostSubnet completely if it is invalid. Rather, we should never even get to this point if the HostSubnet has any invalid data.

We call `ValidateHostSubnetEgress` in `handleAddOrUpdateHostSubnet`/`handleDeleteHostSubnet`, but that function doesn't check `hs.Subnet`, which is a bug. `ValidateHostSubnetEgress` should just call `ValidateHostSubnet` as well, since we require both `hs.Subnet` and `hs.HostIP` to be valid in the egress-handling code.

Then we can safely assume that `net.ParseCIDR(hs.Subnet)` doesn't return an error (and thus we can assume that we'll always have an `sdnIP` for every node).

pkg/network/common/egressip.go
```diff
@@ -493,26 +494,13 @@ func (eit *EgressIPTracker) SetNodeOffline(nodeIP string, offline bool) {
...
-
 // Ping a node and return whether or not we think it is online. We do this by trying to
 // open a TCP connection to the "discard" service (port 9); if the node is offline, the
 // attempt will either time out with no response, or else return "no route to host" (and
 // we will return false). If the node is online then we presumably will get a "connection
 // refused" error; but the code below assumes that anything other than timeout or "no
 // route" indicates that the node is online.
 func (eit *EgressIPTracker) Ping(ip string, timeout time.Duration) bool {
-	// If the caller used a public node IP, replace it with the SDN IP
-	ip = eit.lookupNodeIP(ip)
```
danwinship, line 514: should clarify in the doc comment that you must pass an IP on the SDN, not an IP on the host network. Maybe also change the parameter name from `ip` to `sdnIP`?

pkg/network/master/egressip.go
```diff
@@ -36,6 +37,7 @@ type egressIPManager struct {
 
 type egressNode struct {
 	ip      string
+	sdnIP   string
```
danwinship, line 40: These changes to `pkg/network/master/egressip.go` don't seem to do anything? Because nothing ever uses `sdnIP`.

Oh... ugh... the master code is pinging the hostnetwork IPs rather than the SDN IPs... That shouldn't work... (The firewalls ought to be blocking access to the port it uses.) Hm... I guess if people only use egress IPs on bare metal, and they aren't bothering to set up fully-restrictive firewalls, then this would work... But it's still wrong.

The code in `check()` that calls `Ping()` should be fixed to use the sdn IP rather than the node IP... but that's a big change and not related to this bug, so let's do that as a second commit, and then _not_ backport that commit right away. (So, the first commit would have no changes to `pkg/network/master/egressip.go` except for updating the signature of `ClaimEgressIP`, and then the second commit would add these changes back in [except without the extra `ParseCIDR` error checks because those should be unneeded here too] and also fix the `Ping()` call.)

pkg/network/node/egressip.go
```diff
@@ -153,18 +154,21 @@ func (eip *egressIPWatcher) ReleaseEgressIP(egressIP, nodeIP string) {
...
 }
 
-func (eip *egressIPWatcher) addEgressIP(nodeIP, egressIP string) {
+func (eip *egressIPWatcher) addEgressIP(nodeIP, egressIP, sdnIP string) {
 	eip.monitorNodesLock.Lock()
 	defer eip.monitorNodesLock.Unlock()
 
 	if eip.monitorNodes[nodeIP] != nil {
 		eip.monitorNodes[nodeIP].egressIPs.Insert(egressIP)
+		eip.monitorNodes[nodeIP].sdnIP = sdnIP
```
danwinship, line 163: This should not be needed; the sdn IP can't change while we have an entry for the node IP

pkg/network/node/egressip.go
```diff
@@ -226,7 +230,11 @@ func (eip *egressIPWatcher) getOfflineResult(retrying bool) (map[string]bool, bo
 			continue
 		}
 
-		online := eip.tracker.Ping(node.nodeIP, timeout)
+		dest := node.nodeIP
+		if node.sdnIP != "" {
+			dest = node.sdnIP
```
danwinship, line 235: `egressNode`s are only created in `addEgressIP`, and `sdnIP` is always set there, so this should be unnecessary; you can just pass `node.sdnIP` to `Ping`

danwinship (changes requested):

pkg/network/common/egressip.go
Kyrtapz: After `ValidateHostSubnet` is added to `ValidateHostSubnetEgress` what should we do here? return from function or ignore the error? I would prefer the former

pkg/network/master/egressip.go
danwinship: No, you didn't miss anything; the code was wrong before your PR

pkg/network/common/egressip.go
```diff
@@ -493,27 +501,15 @@ func (eit *EgressIPTracker) SetNodeOffline(nodeIP string, offline bool) {
...
 // refused" error; but the code below assumes that anything other than timeout or "no
 // route" indicates that the node is online.
-func (eit *EgressIPTracker) Ping(ip string, timeout time.Duration) bool {
-	// If the caller used a public node IP, replace it with the SDN IP
-	ip = eit.lookupNodeIP(ip)
-
-	conn, err := net.DialTimeout("tcp", ip+":9", timeout)
+// It is required that the IP provided is from SDN, nodes primary IP might drop traffic destined to port 9
+func (eit *EgressIPTracker) Ping(sdnIP string, timeout time.Duration) bool {
+	conn, err := net.DialTimeout("tcp", sdnIP+":9", timeout)
```
danwinship, line 512: hm, can you fix that to `net.JoinHostPort(sdnIP, "9")` while we're here?

pkg/network/common/egressip_test.go
```diff
@@ -120,9 +120,11 @@ func TestEgressIP(t *testing.T) {
 	eit, w := setupEgressIPTracker(t)
 
 	updateHostSubnetEgress(eit, &osdnv1.HostSubnet{
+		Subnet: "172.17.0.0/24",
 		HostIP: "172.17.0.3",
```
danwinship, line 124: `Subnet` is the SDN subnet assigned to the node, so it doesn't make sense for it to be overlapping `HostIP`... and yeah, it doesn't actually matter for the tests _currently_, but let's just get it right so no one has to change it again in the future.
`"10.128.0.0/23"` is an acceptable value to use instead.

(I guess I should have said that on the `pkg/network/common` tests too. And likewise in the `pkg/network/node` tests below.)

PR #366, opened 2021-11-06
Bug 1996140: Switch from core.Event to events.Event API for unidling

martinkennelly:
client-go new APIs to manage k8s Event API.
This new k8 API is a departure from core's Event API.
We need to upgrade to the new API before
client-go tools remove support for core's Event
API.

In order for unidling to work, we need to upgrade openshift
controller manager to use the new Events APIs.

This patch only focuses on upgrading the Events API for
unidling and subsequent work is needed to update other
parts of this codebase to use the new Event API.

WIP until https://github.com/openshift/openshift-controller-manager/pull/208 merges.

### DIFF
--- a/pkg/cmd/openshift-sdn-node/sdn.go
+++ b/pkg/cmd/openshift-sdn-node/sdn.go
@@ -1,12 +1,12 @@
 package openshift_sdn_node
 
 import (
+	"fmt"
 	"io/ioutil"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/kubernetes/scheme"
-	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
-	"k8s.io/client-go/tools/record"
+	eventsv1 "k8s.io/client-go/tools/events"
 
 	sdnnode "github.com/openshift/sdn/pkg/network/node"
 )
@@ -14,10 +14,10 @@ import (
 const openshiftCNIFile string = "/etc/cni/net.d/80-openshift-network.conf"
 
 // initSDN sets up the sdn process.
-func (sdn *openShiftSDN) initSDN() error {
-	eventBroadcaster := record.NewBroadcaster()
-	eventBroadcaster.StartRecordingToSink(&corev1client.EventSinkImpl{Interface: sdn.informers.kubeClient.CoreV1().Events("")})
-	sdn.sdnRecorder = eventBroadcaster.NewRecorder(scheme.Scheme, corev1.EventSource{Component: "openshift-sdn", Host: sdn.nodeName})
+func (sdn *openShiftSDN) initSDN(stopCh <-chan struct{}) error {
+	eventBroadcaster := eventsv1.NewBroadcaster(&eventsv1.EventSinkImpl{Interface: sdn.informers.kubeClient.EventsV1()})
+	eventBroadcaster.StartRecordingToSink(stopCh)
+	sdn.sdnRecorder = eventBroadcaster.NewRecorder(scheme.Scheme, "openshift-sdn")
 
 	var err error
 	sdn.osdnNode, err = sdnnode.New(&sdnnode.OsdnNodeConfig{
@@ -45,7 +45,9 @@ func (sdn *openShiftSDN) runSDN() error {
 
 func (sdn *openShiftSDN) writeConfigFile() error {
 	// Make an event that openshift-sdn started
-	sdn.sdnRecorder.Eventf(&corev1.ObjectReference{Kind: "Node", Name: sdn.nodeName}, corev1.EventTypeNormal, "Starting", "openshift-sdn done initializing node networking.")
+	nodeRef := &corev1.ObjectReference{Kind: "Node", Name: sdn.nodeName}
+	note := fmt.Sprintf("openshift-sdn completed initializing networking for node %s", sdn.nodeName)
+	sdn.sdnRecorder.Eventf(nodeRef, nil, corev1.EventTypeNormal, "", "NodeNetworkReady", note)
 
 	// Write our CNI config file out to disk to signal to kubelet that
 	// our network plugin is ready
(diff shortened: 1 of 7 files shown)


martinkennelly:
Forced push to change `NeedsPods` event `action` to `None` instead of `unidling` because according to [1]:

> 'action' explains what happened with regarding/what action did the ReportingController
> (ReportingController is a type of a Controller reporting an Event, e.g. k8s.io/node-controller, k8s.io/kubelet.)
> take in regarding's name; it should be in UpperCamelCase format (starting with a capital letter).

I tried to find _if_ and _what_ string should denote no action taken and could not find any examples in k8 nor in any docs - so I choose `None`.


[1] https://github.com/kubernetes/client-go/blob/9b0b23a8ade2b5323d6624146cea2ad7b8928f25/tools/events/interfaces.go#L42

pkg/network/proxy/unidler/unidlerproxy.go
```diff
@@ -35,14 +35,14 @@ func (sig *eventSignaler) NeedPods(serviceName types.NamespacedName, port string
 	}
 
 	// HACK: make the message different to prevent event aggregation
-	sig.recorder.Eventf(&serviceRef, v1.EventTypeNormal, unidlingapi.NeedPodsReason, "The service-port %s:%s needs pods.", serviceRef.Name, port)
+	sig.recorder.Eventf(&serviceRef, nil, v1.EventTypeNormal, unidlingapi.NeedPodsReason, "None", "The service-port %s:%s needs pods.", serviceRef.Name, port)
```
danwinship, line 38:
```suggestion
	sig.recorder.Eventf(&serviceRef, nil, v1.EventTypeNormal, unidlingapi.NeedPodsReason, "Unidling", "The service-port %s:%s needs pods.", serviceRef.Name, port)
```

martinkennelly: I disagree with this `action`. SDN is not performing this action - another software component is performing this action and SDN is requesting it via this event. Let me know what you think.

From client-go event recorder interface [1]:
> 	// 'action' explains what happened with regarding/what action did the **ReportingController**
	// (ReportingController is a type of a Controller reporting an Event, e.g. k8s.io/node-controller, k8s.io/kubelet.)
	// take in regarding's name; it should be in UpperCamelCase format (starting with a capital letter).
	// 'note' is intended to be human readable.

[1] https://github.com/kubernetes/client-go/blob/9b0b23a8ade2b5323d6624146cea2ad7b8928f25/tools/events/interfaces.go#L42

martinkennelly: I have not seen any precedent in upstream k8 code using this `None` action label.

danwinship: Yeah, I read that as, there are two ways you can use `action`; it either "explains what happened with regarding" or "explains what action did the ReportingController take in regarding's name". `"Unidling"` fits the former case.

martinkennelly:
Rebased PR, updated node events also, re-tested it against OCM patch.
Unable to update LE events to v1 due to LE lib not yet supporting new Eventf interface.

pkg/cmd/openshift-sdn-node/sdn.go
```diff
@@ -1,23 +1,23 @@
...
 
 // initSDN sets up the sdn process.
-func (sdn *openShiftSDN) initSDN() error {
-	eventBroadcaster := record.NewBroadcaster()
-	eventBroadcaster.StartRecordingToSink(&corev1client.EventSinkImpl{Interface: sdn.informers.kubeClient.CoreV1().Events("")})
-	sdn.sdnRecorder = eventBroadcaster.NewRecorder(scheme.Scheme, corev1.EventSource{Component: "openshift-sdn", Host: sdn.nodeName})
+func (sdn *openShiftSDN) initSDN(stopCh <-chan struct{}) error {
+	eventBroadcaster := eventsv1.NewBroadcaster(&eventsv1.EventSinkImpl{Interface: sdn.informers.kubeClient.EventsV1()})
+	eventBroadcaster.StartRecordingToSink(stopCh)
+	sdn.sdnRecorder = eventBroadcaster.NewRecorder(scheme.Scheme, fmt.Sprintf("openshift-sdn-node-%s", sdn.nodeName))
```
danwinship, line 20: It appears that the new API has both `ReportingController` and `ReportingInstance`, where the string passed here becomes `ReportingController`, and the client-go code does:

```
	hostname, _ := os.Hostname()
	reportingInstance := reportingController + "-" + hostname
```

so we shouldn't manually add the nodename. Just keep passing "openshift-sdn"

martinkennelly:
@danwinship I think we should wait until 4.12 to get this in. WDYT?
It works now for OVN-K and SDN. I fixed the version skew issue I did not consider by simply checking `depreciatedLastTimestamp` is not zero:

https://github.com/openshift/openshift-controller-manager/pull/208/files#diff-0a4e6f20fb6c679b979f881f9d072be1335939d753019afa0f43c5055a7ca75eR66

We can remove this check on `depreciatedLastTimestamp` from ocm in 4.13 if we update OVN-K and SDN in 4.12. This will allow for network plugin rollback.

If you wish to test the patch series for SDN:
`test e2e-all 4.11,openshift/sdn#366,openshift/openshift-apiserver#256,openshift/openshift-controller-manager#208`

When we proceed, we review and merge in the following order:
* openshift/openshift-apiserver#256
* openshift/openshift-controller-manager#208
* openshift/sdn#366

I found no migration guide upstream - I simply read: https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/383-new-event-api-ga-graduation and especially the comparison between the two APIs: https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/383-new-event-api-ga-graduation#comparison-between-old-and-new-apis

PR #372, opened 2021-11-18
adding network policy egress to openshift-sdn

JacobTanenbaum:
This series of commits adds support for NetworkPolicy egress to openshift-sdn


danwinship:
Oops, I was thinking you hadn't filed this yet. So did you figure out the failing e2e test or is it still failing?

>     Split up some of the NetworkPolicy parsing code 9766dbd
>     Remove superfluous reg0 checks from NetworkPolicy flows e1778c8
>     ovscontroller: document tables at the top 14ded4f

These already got added as part of #350, and the stuff you have for those commits here is just merge errors. So drop those three commits.

pkg/network/node/networkpolicy.go
```diff
@@ -125,13 +127,28 @@ func (np *networkPolicyPlugin) Start(node *OsdnNode) error {
...
+	// The dummy nat(src=0.0.0.0) ensures we handle tuple collisions; see
+	// https://bugzilla.redhat.com/show_bug.cgi?id=1910378
+	for _, scn := range np.node.networkInfo.ClusterNetworks {
+		for _, dcn := range np.node.networkInfo.ClusterNetworks {
+			otx.AddFlow("table=30, priority=100, ip, nw_src=%s, nw_dst=%s, ct_state=-rpl, actions=ct(commit,table=31)", scn.ClusterCIDR.String(), dcn.ClusterCIDR.String())
+		}
 	}
+
+	// skip policy enforcement for egress when the destination address is a ServiceIP
+	otx.AddFlow("table=27, priority=300, ip, nw_dst=%s , actions=goto_table:30", node.networkInfo.ServiceNetwork.String())
```
danwinship, line 146: We're not _skipping_ it, we're just postponing it. Maybe:

    // Egress enforcement for pod-to-Service IP will happen after the packets are rewritten by
    // iptables and come back via table 25.

Also, I generally like to add the rules in the "right" order, so that the order you read them in in the code is the same as the order they get processed in... oh, hah, but it was already wrong. OK, can you fix it so this goes:

- `otx := node.oc.NewTransaction()`
- "service IP" comment plus the priority 300 rule for service IPs
- "replies" comment plus the priority 200 rules
- the big "conntrack" comment plus the loop with the priority 100 rules

danwinship (commented):
Well, this looks good to me, but then, I wrote most of it so it would be good if someone else could lgtm it too... :slightly_smiling_face:

pkg/network/node/networkpolicy.go
JacobTanenbaum: done

JacobTanenbaum:
I am still trying to work out why that specific e2e test fails

pkg/network/node/networkpolicy.go
```diff
@@ -113,13 +127,29 @@ func (np *networkPolicyPlugin) Start(node *OsdnNode) error {
...
 	otx := node.oc.NewTransaction()
-	for _, cn := range np.node.networkInfo.ClusterNetworks {
-		// Must pass packets through CT NAT to ensure NAT state is handled
-		// correctly by OVS when NAT-ed packets have tuple collisions.
-		// https://bugzilla.redhat.com/show_bug.cgi?id=1910378
-		otx.AddFlow("table=21, priority=200, ip, nw_dst=%s, ct_state=-rpl, actions=ct(commit,nat(src=0.0.0.0),table=30)", cn.ClusterCIDR.String())
-	}
+
+	// Egress enforcement tfor pod-to-Service IP will happen after the packets are rewritten
+	// iptables and come back via table 25.
```
danwinship, line 132:
```suggestion
	// Egress enforcement for pod-to-Service IP will happen after the packets are rewritten by
	// iptables and come back via table 25.
```

danwinship:
also, doesn't compile...

pkg/network/node/networkpolicy.go
JacobTanenbaum: done

danwinship:
e2e-aws-upgrade has failed multiple times now but [the test grid](https://testgrid.k8s.io/redhat-openshift-ocp-release-4.10-informing#periodic-ci-openshift-release-master-ci-4.10-e2e-aws-upgrade) shows it pretty reliably flaky-but-passing. So maybe something in this PR breaks upgrades somehow?

danwinship:
Also, it seems that this isn't passing most of the egress tests?
https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-aws-modern/1465463763506302976
(That's from "`test e2e nightly,openshift/origin#26647,openshift/sdn#372 aws`" (which includes the ipblock except tests too because I was going to test against #374 too)

pkg/network/node/networkpolicy.go
```diff
@@ -849,9 +849,17 @@ func (np *networkPolicyPlugin) handleAddOrUpdatePod(obj, old interface{}, eventT
...
+		otx := np.node.oc.NewTransaction()
+		otx.DeleteFlows("table=27, cookie=1/-1, ip, nw_src=%s", pod.Status.PodIP)
+		otx.DeleteFlows("table=80, cookie=1/-1, ip, nw_src=%s", pod.Status.PodIP)
+		otx.Commit()
+
+		np.lock.Unlock()
+	}()
 
 	np.refreshPodNetworkPolicies(pod)
+
```
danwinship, line 862: extra blank line

pkg/network/node/ovscontroller.go
```diff
@@ -482,6 +482,12 @@ func (oc *ovsController) setupPodFlows(ofport int, podIP net.IP, vnid uint32) er
 	// IP traffic to container
 	otx.AddFlow("table=70, priority=100, ip, nw_dst=%s, actions=load:%d->NXM_NX_REG1[], load:%d->NXM_NX_REG2[], goto_table:80", ipstr, vnid, ofport)
 
+	// Isolate this pod from traffic until existing networkPolicy rules are applied
+	if oc.pluginId == 2 {
```
danwinship, line 486: ew, I don't like having ovscontroller encode this knowledge.

Can you add a `SetUpPod` API to the `osdnPolicy` interface? It can be a no-op for subnet and multitenant, and then for networkpolicy it can add these blocking rules. Then you can call `m.policy.SetUpPod` from `setup` in `pod.go` after calling `m.ovs.SetUpPod`. (Or maybe before?)

pkg/network/node/networkpolicy.go
JacobTanenbaum: removed

pkg/network/node/ovscontroller.go
JacobTanenbaum: DONE: I think that I setup as you intended with this comment

danwinship:
/lgtm
:tada:

PR #373, opened 2021-11-23
Bug 2024880: [EgressIP] move `ct(commit)` action from OVS group to flow

alexanderConstantinescu:
During the implementation of traffic load balancing for egress IP we
simply copied all actions specified on the flow (which did traffic
re-routing to the egress node) in previous versions, to the
group (which did traffic re-routing to multiple egress nodes). However
OVS flows use "OVS action lists", while groups use "OVS action sets", and the
semantics between both differ quite substantially, see:
https://docs.openvswitch.org/en/latest/ref/ovs-actions.7

The impact of this is: whenever a pod initiates an egress
connection and we route that connection to the egress node, we also do
not send the connection to conntrack. This means that the reply packets
will be treated as an incoming new connection which are susceptible to
network policy evaluation. If a network policy is specified which
happens to block ingress from the endpoint the pod is trying to connect
to, the connection will be blocked on the reply path - though it should
be allowed because the connection is actually an egress-ing connection.

This patch also fixes some egress IP tests which were logically flawed

/assign @danwinship

### DIFF
--- a/pkg/network/node/ovscontroller_test.go
+++ b/pkg/network/node/ovscontroller_test.go
@@ -1185,14 +1185,14 @@ var expectedFlows = []string{
 	" cookie=0, table=100, priority=1, reg0=42, ip, nw_dst=192.168.1.1/32, actions=goto_table:101",
 	" cookie=0, table=100, priority=0, actions=goto_table:101",
 	" cookie=0, table=101, priority=150, ct_state=+rpl, actions=output:2",
-	" cookie=0, table=101, priority=100, ip, reg0=37, actions=group:37",
+	" cookie=0, table=101, priority=100, ip, reg0=37, actions=ct(commit),group:37",
 	" cookie=0, table=101, priority=0, actions=output:2",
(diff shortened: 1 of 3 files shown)


pkg/network/node/ovscontroller_test.go
```diff
@@ -1185,14 +1185,14 @@ var expectedFlows = []string{
 	" cookie=0, table=100, priority=1, reg0=42, ip, nw_dst=192.168.1.1/32, actions=goto_table:101",
 	" cookie=0, table=100, priority=0, actions=goto_table:101",
 	" cookie=0, table=101, priority=150, ct_state=+rpl, actions=output:2",
-	" cookie=0, table=101, priority=100, ip, reg0=37, actions=group:37",
+	" cookie=0, table=101, priority=100, ip, reg0=37, actions=ct(commit),group:37",
```
danwinship, line 1188: hm... `TestRuleVersion` isn't really doing a good job of testing egress IPs... we should fix that (have actual proper rules for both a local egress IP and a remote one)

PR #378, opened 2021-12-06
block metadata service for Alibaba Cloud

Lyt99:
block metadata service '100.100.100.200' on Alibaba Cloud

wait for #365 to be merged to provide ability to detect platform type

Signed-off-by: xiayu.lyt <xiayu.lyt@alibaba-inc.com>


Lyt99:
Duplicated with #380, closed.

PR #381, opened 2021-12-13
Handle MTU migration

danwinship:
Implements https://issues.redhat.com/browse/SDN-2553
Depends on https://github.com/openshift/cluster-network-operator/pull/1241

/cc @Kyrtapz @jcaamano

### DIFF
--- a/pkg/cmd/openshift-sdn-cni/openshift-sdn.go
+++ b/pkg/cmd/openshift-sdn-cni/openshift-sdn.go
@@ -165,7 +165,8 @@ func (p *cniPlugin) CmdAdd(args *skel.CmdArgs) error {
 
 	// ipam.ConfigureIface handles the Routes "incorrectly" (if there is no gateway
 	// specified it uses the interface's default gateway as the next hop rather than
-	// passing nil as the next hop. So pull out the routes to handle on our own.
+	// passing nil as the next hop. Additionally, we may need to set the MTU on the
+	// routes. So pull out the routes to handle on our own.
 	defaultGW := result.IPs[0].Gateway
 	routes := result.Routes
 	result.Routes = nil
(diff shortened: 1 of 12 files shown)


Kyrtapz:
There is an issue with setting up the routes:
`error adding container to network "openshift-sdn": failed to add route to 10.128.0.0/14 via SDN: no such device`

I think that it is related to the fact that there is no link specified [here](https://github.com/openshift/sdn/pull/381/files#diff-8b4e24ba0764957c1addd72fceac5346db44409597863cd4884aa95ea885decdR240-R247).
`ConfigureIface` always specifies a link when adding routes:
```
func ConfigureIface(ifName string, res *current.Result) error {
...
		if err = ip.AddRoute(&r.Dst, gw, link); err != nil {
			// we skip over duplicate routes as we assume the first one wins
			if !os.IsExist(err) {
				return fmt.Errorf("failed to add route '%v via %v dev %v': %v", r.Dst, gw, ifName, err)
			}
		}
	}
```

jcaamano:
I am missing the routable-mtu being set on the service network route on the host itself.

pkg/cmd/openshift-sdn-cni/openshift-sdn.go
```diff
@@ -233,18 +229,22 @@ func (p *cniPlugin) CmdAdd(args *skel.CmdArgs) error {
...
 			}
 		}
 
-		dsts = append(dsts, serviceIPNet)
-		for _, dst := range dsts {
+		for _, cniroute := range routes {
 			route := &netlink.Route{
-				Dst: dst,
-				Gw:  defaultGW,
+				LinkIndex: link.Attrs().Index,
```
Kyrtapz, line 241: I think this should be a link of `args.IfName`, not `macvlan0`

jcaamano:
@danwinship We need to use the migration mtu here as well during migration, otherwise observing br0 and tun0 with an incorrect mtu.
https://github.com/openshift/sdn/blob/99d96f8a65b6ae9c7a33ebf3d7d2391101b213ad/pkg/network/node/sdn_controller.go#L157-L159

alexanderConstantinescu (approved):
/lgtm

PR #383, opened 2021-12-15
Fix flaky check for capacity test for egress IP, and better doc blocs

alexanderConstantinescu:
During the work on #365 we ended up removing the part of the assignment algorithm in `common.go`, which accepted the `fullReallocation` in some cases where more assignments could be made at the expense of changes. We however forgot to remove one check in the unit tests which was based upon this fact. This in turn lead to flakes, since any re-allocation now depends on the state of the previous round of assignments and we can't assume a full assignment anymore.

This has been verified by running: `go test -mod=vendor -race ./... -count 10` with success. I didn't try with a higher count than that though

/assign @danwinship


danwinship:
/lgtm
thanks for the quick turnaround

PR #384, opened 2021-12-20
Bug 2033332: It is possible to have Networkpolicy types with no rules

JacobTanenbaum:
If there is a NetworkPolicy type specified but no rules included the
namespaces vnid will have no flow for what to do for pods in that
namespace which will effectivly isolate them

change the behavior so that if there is no rule present but the policy
type is still specified that the default non-isolating rule is generated.

https://bugzilla.redhat.com/show_bug.cgi?id=2033332


JacobTanenbaum:
this is not actually an issue

JacobTanenbaum:
not actually an issues

PR #389, opened 2022-01-11
Bug 2038793: Use the kubeClient instead of the informer cache to fetch nodes for egress IP validation.

alexanderConstantinescu:
When validating the .egressIPs defined on a hostSubnet we were using
the informer cache to retrieve the node object and validate that the
egress IP is correctly defined on the cloud subnet. However, the node
informer cache is not guaranteed to be warm when the hostSubnet
informers start, which leads to race events and in case the hostSubnet
informer does not find the node object: a removal of a valid egress IP.

I investigated if there's a way to use WaitForCacheSync as to ensure
that the node informer cache is warm before the hostSubnet informer
starts, however due to the "legacy nature" which openshift-sdn employs
to start its controllers I found the task rather convoluted. Hence, the
solution used in this PR is just to use the kubeClient to fetch the
nodes.

### DIFF
--- a/pkg/network/common/egressip.go
+++ b/pkg/network/common/egressip.go
@@ -1,6 +1,7 @@
 package common
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"math"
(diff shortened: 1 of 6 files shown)


alexanderConstantinescu:
/retitle Bug 2038793: Use the kubeClient instead of the informer cache to fetch nodes for egress IP validation.
/assign @danwinship

squeed:
One idea I had was to track whether or not the node informer is synced, and use the direct client until that is the case. You already have a waitForCacheSynced(), could throw something in an atomic and just branch off of that.

This would only really be useful if you expect this to be a high-traffic call. I don't have enough context from the code to know how often you expect to be talking to the apiserver. If its only a few times a minute, then whatever.

danwinship:
/lgtm
:man_shrugging:

PR #392, opened 2022-01-20
clean up OWNERS

danwinship:
remove people who are no longer on the team


martinkennelly:
You may as well remove `alexanderConstantinescu` also while you're at it.

PR #393, opened 2022-01-20
move to-be-deleted userspace proxy code into sdn tree

danwinship:
The plot to kill off the userspace proxy upstream (https://github.com/kubernetes/kubernetes/issues/103860) is progressing, so let's absorb it into the sdn tree before the next rebase deletes it.

The first commit just moves the code (and updates the already-in-tree unidling code for the new import path). The remaining commits just get rid of some code that's no longer used (including abstractions that now only have one implementation).

There is probably more dead code in there, but it's not quite as obviously dead...

/assign @squeed

/hold
until master switches to 4.11


squeed:
Wow, that's not a lot of code.

squeed:
Seems basically good to me - hopefully this is an easy merge once we branch 4.11.

squeed:
(do we have good unidling e2e coverage? I see that there are a decent number of tests)

/lgtm

PR #395, opened 2022-01-24
Bug 2043802: Fix CloudPrivateIPConfig enqueueing during egressip reassignment

kyrtapz:
Compare node name against the existing CloudPrivateIPConfig instead of the new one.

### DIFF
--- a/pkg/network/master/egressip.go
+++ b/pkg/network/master/egressip.go
@@ -312,12 +312,10 @@ func (eim *egressIPManager) ClaimEgressIP(vnid uint32, egressIP, nodeIP, sdnIP s
 				Node: nodeName,
 			},
 		}
-		if _, err := eim.cloudNetworkClient.CloudV1().CloudPrivateIPConfigs().Create(context.TODO(), &cloudPrivateIPConfig, metav1.CreateOptions{}); err != nil {
-			if kerrors.IsAlreadyExists(err) {
-				if cloudPrivateIPConfig.Spec.Node != nodeName {
-					klog.Infof("CloudPrivateIPConfig: %s is being moved and still exists, enqueuing its creation", egressIP)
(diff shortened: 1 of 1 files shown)


pkg/network/master/egressip.go
```diff
@@ -312,12 +312,10 @@ func (eim *egressIPManager) ClaimEgressIP(vnid uint32, egressIP, nodeIP, sdnIP s
...
 			},
 		}
-		if _, err := eim.cloudNetworkClient.CloudV1().CloudPrivateIPConfigs().Create(context.TODO(), &cloudPrivateIPConfig, metav1.CreateOptions{}); err != nil {
-			if kerrors.IsAlreadyExists(err) {
-				if cloudPrivateIPConfig.Spec.Node != nodeName {
-					klog.Infof("CloudPrivateIPConfig: %s is being moved and still exists, enqueuing its creation", egressIP)
-					eim.cloudPrivateIPConfigCreationQueue[egressIP] = cloudPrivateIPConfig
-				}
+		if existingCloudPrivateIPConfig, err := eim.cloudNetworkClient.CloudV1().CloudPrivateIPConfigs().Create(context.TODO(), &cloudPrivateIPConfig, metav1.CreateOptions{}); err != nil {
+			if kerrors.IsAlreadyExists(err) && existingCloudPrivateIPConfig.Spec.Node != nodeName {
```
alexanderConstantinescu, line 316: Could you put these conditions in separate blocks, as it was before? The reason I am asking is to not log spurious error messages on L320, which would happen if the object is existing and assigned to the node we want it to.

alexanderConstantinescu (changes requested):

pkg/network/master/egressip.go
alexanderConstantinescu: Agreed off-band to keep things as they are

PR #400, opened 2022-02-24
Bug 2050230: Use informer cache when listing resources

jcaamano:
Using the resource List API without pagination overloads the api server
for a high number of resources.

The informer cache is populated using reflector which in turn does use
pagination.


pkg/network/common/list_pager.go
```diff
@@ -0,0 +1,110 @@
...
+
+	osdnv1 "github.com/openshift/api/network/v1"
+	osdnclient "github.com/openshift/client-go/network/clientset/versioned"
+)
+
+func ListAllEgressNetworkPoliciesWithPager(client osdnclient.Interface) ([]*osdnv1.EgressNetworkPolicy, error) {
+	list := []*osdnv1.EgressNetworkPolicy{}
+	err := pager.New(func(ctx context.Context, opts metav1.ListOptions) (runtime.Object, error) {
+		return client.NetworkV1().EgressNetworkPolicies(metav1.NamespaceAll).List(ctx, opts)
+	}).EachListItem(context.TODO(), metav1.ListOptions{}, func(obj runtime.Object) error {
```
squeed, line 23: This (and all similar methods) should take a Context as their first argument, rather than doing TODO here.

squeed:
If you're going down this route, can we just list from the informer cache instead?

jcaamano:
> If you're going down this route, can we just list from the informer cache instead?

I preferred & tried that first. I got into trouble. I ended up guessing this requires some control on waiting for the informers to be synced (to use them to list & initialize stuff on startup) before the informer handlers start doing their thing, which is something the current state of the code is not setup to easily achieve without some level of refactor I could not be bothered with in the context of the BZ (from the I-don't-want-to-break-anything-in-the process-of-fixing-something-else perspective).

But do let me know if you think we should give that refactor a go or you otherwise think there is not much of a refactor needed to achieve it.

squeed:
@jcaamano oh yeah, now that you mention that, I remember going down the same route. Indeed, such a refactor doesn't seem wise.

We'll, someday, have to bite the bullet and add a proper workqueue. Argh.

jcaamano:
@squeed PTAL :pray:

danwinship:
Alexander ran into the same problem in #389... Here's some of the discussion from Slack:

> The first version of the code had this: https://github.com/openshift/sdn/commit/519eaa195d7fb68ca416c77179b7d5adc5249ecc But that blocks indefinitely because the `controller` (https://github.com/kubernetes/client-go/blob/master/tools/cache/shared_informer.go#L427) is uninitialized at that point. The initialization happens when you `Run` the Informer,see: https://github.com/kubernetes/client-go/blob/master/tools/cache/shared_informer.go#L396 . We `Run` our informer afterwards, here: https://github.com/openshift/sdn/blob/master/pkg/cmd/openshift-sdn-node/cmd.go#L183 which is after `runSDN` which initializes `EgressIPWatcher`.  I also tried changing the order and have `sdn.informers.start(stopCh)` happen before `runSDN`, but that doesn't work because when we start the informer we first need to have it added to the list of informers in the factory, see: https://github.com/kubernetes/client-go/blob/master/informers/factory.go#L132 , this is done when `Informer()` is called, here: https://github.com/openshift/sdn/blob/master/pkg/network/common/egressip.go#L155
It seems to me that the right sequence of events as to not have `WaitForCacheSync` block is:
> 
> 1. Add the informer to the factory by calling `Informer()`
> 2. Start the informer
> 3. Call `WaitForCacheSync`
>
> we instead do 1. then 3. then 2. and refactoring that seemed complicated to me.

So, anyway, maybe we're going to keep bumping into this problem and should fix it?

jcaamano:
> Alexander ran into the same problem in #389... Here's some of the discussion from Slack:
> 
> > The first version of the code had this: [519eaa1](https://github.com/openshift/sdn/commit/519eaa195d7fb68ca416c77179b7d5adc5249ecc) But that blocks indefinitely because the `controller` (https://github.com/kubernetes/client-go/blob/master/tools/cache/shared_informer.go#L427) is uninitialized at that point. The initialization happens when you `Run` the Informer,see: https://github.com/kubernetes/client-go/blob/master/tools/cache/shared_informer.go#L396 . We `Run` our informer afterwards, here: https://github.com/openshift/sdn/blob/master/pkg/cmd/openshift-sdn-node/cmd.go#L183 which is after `runSDN` which initializes `EgressIPWatcher`.  I also tried changing the order and have `sdn.informers.start(stopCh)` happen before `runSDN`, but that doesn't work because when we start the informer we first need to have it added to the list of informers in the factory, see: https://github.com/kubernetes/client-go/blob/master/informers/factory.go#L132 , this is done when `Informer()` is called, here: https://github.com/openshift/sdn/blob/master/pkg/network/common/egressip.go#L155
> > It seems to me that the right sequence of events as to not have `WaitForCacheSync` block is:
> > 
> > 1. Add the informer to the factory by calling `Informer()`
> > 2. Start the informer
> > 3. Call `WaitForCacheSync`
> > 
> > we instead do 1. then 3. then 2. and refactoring that seemed complicated to me.
> 
> So, anyway, maybe we're going to keep bumping into this problem and should fix it?

@danwinship I did have that into account when I tried with informers cache. But there is a follow up problem: it requires pretty clear semantics and structure code differentiating initialization and start/run:

1. Add the informer to the factory by calling `Informer()`
2. Start the informer
3. Call `WaitForCacheSync`
4. Add handlers to informers (or start work queues, or whatever)

Now separating 2, required by 3, from 4 is the problem that requires a non-trivial refactor.

squeed:
The other alternative is to insert a workqueue between the informers and the workers. Then the step is 

1. Set up informers
2. Start up informers, fill up queues
3. WaitForCacheSync
4. Start up queue workers.

jcaamano:
@danwinship @squeed But do we want to do it in the context of this BZ?

squeed:
> @danwinship @squeed But do we want to do it in the context of this BZ?

Depends: does this need to be backported forever? If it does, then, no, we should not do a wholesale refactor in a bz :).

jcaamano:
> > @danwinship @squeed But do we want to do it in the context of this BZ?
> 
> Depends: does this need to be backported forever? If it does, then, no, we should not do a wholesale refactor in a bz :).

Probably to 4.8 which the BZ is opened against. And if we follow our usual backport-all-the-way policy that's gonna be 4.6. And then you never know when a backport request is going to come your way after the fact.

squeed:
Alas, c'est la vie.
/lgtm

afcollins (commented):
Thank you for the fix! @jcaamano 🙌

PR #401, opened 2022-03-01
Add unit tests for vxlan no track ip table rule

pperiyasamy:
This is the follow up PR adding unit tests for completeness of CI for the fix https://github.com/openshift/sdn/pull/324 .


palonsoro (commented):
If I am understanding this well, it just checks the existence of the chain.

It would be good if it could also check that the rules are correct.

Concretely, the `OPENSHIFT-NOTRACK` chain should have the `NOTRACK` rule like this:
```
$ iptables -t raw -S OPENSHIFT-NOTRACK
-N OPENSHIFT-NOTRACK
-A OPENSHIFT-NOTRACK -p udp -m udp --dport 4789 -j NOTRACK
```

And be invoked from `OUTPUT` and `PREROUTING` tables like this:
```
$ iptables -t raw -S OUTPUT
-P OUTPUT ACCEPT
-A OUTPUT -m comment --comment "disable conntrack for vxlan" -j OPENSHIFT-NOTRACK

$ iptables -t raw -S PREROUTING
-P PREROUTING ACCEPT
-A PREROUTING -m comment --comment "disable conntrack for vxlan" -j OPENSHIFT-NOTRACK
```

pperiyasamy:
> If I am understanding this well, it just checks the existence of the chain.
> 
> It would be good if it could also check that the rules are correct.
> 
> Concretely, the `OPENSHIFT-NOTRACK` chain should have the `NOTRACK` rule like this:
> 
> ```
> $ iptables -t raw -S OPENSHIFT-NOTRACK
> -N OPENSHIFT-NOTRACK
> -A OPENSHIFT-NOTRACK -p udp -m udp --dport 4789 -j NOTRACK
> ```
> 
> And be invoked from `OUTPUT` and `PREROUTING` tables like this:
> 
> ```
> $ iptables -t raw -S OUTPUT
> -P OUTPUT ACCEPT
> -A OUTPUT -m comment --comment "disable conntrack for vxlan" -j OPENSHIFT-NOTRACK
> 
> $ iptables -t raw -S PREROUTING
> -P PREROUTING ACCEPT
> -A PREROUTING -m comment --comment "disable conntrack for vxlan" -j OPENSHIFT-NOTRACK
> ```

Thanks @palonsoro, I've now added the check for validating these rules. hope that looks ok.

palonsoro:
It looks good to me

palonsoro:
I don't have much experience with the fake iptables testing, but the code itself seems to test what needs to be tested

knobunc:
/ok-to-test
/assign @astoycos 
/lgtm

@astoycos can you take a look?  It seems plausible to me, but you wrote the PR that implemented the change.  Are there nuances we aren't testing?

astoycos:
Hey @knobunc All I actually did in https://github.com/openshift/sdn/pull/335 was get some commits from @palonsoro merged.  I do just have a few comments/ Questions I'll post here.

pkg/network/node/iptables_test.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+func validateIPTableRuleForVxlanPort(t *testing.T, dstPort uint32) {
+	ipt := iptablestest.NewFake()
+	nodeIpt := newNodeIPTables(ipt, nil, true, dstPort, uint32(0))
+	err := nodeIpt.syncIPTableRules()
+	if err != nil {
+		t.Fatalf("unexpected error while syncing ip table rules: %v", err)
+	}
+	exists, err := ipt.EnsureRule(iptables.Append, iptables.Table("raw"), iptables.Chain("PREROUTING"), "-m", "comment", "--comment", "disable conntrack for vxlan", "-j", "OPENSHIFT-NOTRACK")
```
astoycos, line 26: So i'm not sure `EnsureRule` Is the right function here.... At least for an instance of iptablestest it always seems to return `true` see -> https://github.com/kubernetes/kubernetes/blob/0fc6edeb1da8e3975be0cf5fc1b3f6467188de63/pkg/util/iptables/testing/fake.go#L102

What we want here is pretty much what the kubeproxy does, i.e https://github.com/kubernetes/kubernetes/blob/f3bf7e1ced956d5c23c5870699eae22eb55654c1/pkg/proxy/iptables/proxier_test.go#L4223

Where we let the `nodeIpt` sync and then ensure that the iptables rules are correctly programmed, I _think_ we can get all the programed iptables rules with [`FakeIPtables.save()`](https://github.com/kubernetes/kubernetes/blob/0fc6edeb1da8e3975be0cf5fc1b3f6467188de63/pkg/util/iptables/testing/fake.go#L122)

Please feel free to correct me if I'm missing something here.

pkg/network/node/iptables_test.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	exists, err = ipt.EnsureRule(iptables.Append, iptables.Table("raw"), iptables.Chain("OPENSHIFT-NOTRACK"), "-p", "udp", "--dport", fmt.Sprint(dstPort), "-j", "NOTRACK")
+	if err != nil {
+		t.Fatalf("error while checking vxlan ip table notrack rule: %v", err)
+	}
+	if !exists {
+		t.Fatalf("vxlan ip table notrack rule must exist")
+	}
+	ipt.DeleteRule(iptables.Table("raw"), iptables.Chain("PREROUTING"), "-m", "comment", "--comment", "disable conntrack for vxlan", "-j", "OPENSHIFT-NOTRACK")
```
astoycos, line 49: You shouldn't have to manually clean anything up here, the fake rules are just programmed in the fakeiptables instance which is in memory and should be pruned once the unit test ends

astoycos:
@danwinship May also have an opinion here :)

pkg/network/node/iptables_test.go
pperiyasamy: The k8s [FakeIPTables#Lines](https://github.com/kubernetes/kubernetes/blob/0fc6edeb1da8e3975be0cf5fc1b3f6467188de63/pkg/util/iptables/testing/fake.go#L61) is never used for populating rules and it does almost nothing for iptable methods that are needed by `nodeIpt.syncIPTableRules`, so save method can't be used. Looks like kube proxy has its own [in memory](https://github.com/kubernetes/kubernetes/blob/f3bf7e1ced956d5c23c5870699eae22eb55654c1/pkg/proxy/iptables/proxier.go#L225) which is then used for its tests validation and guess this fake iptables is for completion purpose. 

The new commit adds `NodeFakeIPTables` with necessary methods implemented for `nodeIpt.syncIPTableRules`, is that ok for now ?

pkg/network/node/iptables_test.go
pperiyasamy: sure, done.

pkg/network/node/iptables_test.go
astoycos: A little clunky... but that's alright since we don't maintain an internal Cache 

This is now functionally correct but my main NIT is that the test shouldn't be calling `ipt.ensureRule`(which is mutating ipt.Rules by creating the chain if not there) since that is being done by the node controller when  `err := nodeIpt.syncIPTableRules()` is called.  All the test should be doing is verifying that  `ipt.Rules` contains the correct rules, so maybe define `ipt.Present()` to just check for existence? 

Sorry to be a pain :)

PR #403, opened 2022-03-01
WIP minimize iptables changes

danwinship:
/cc


danwinship:
Super-unscientific results:
Average syncProxyRules time on a randomly-selected node in the last e2e-aws run on #414: **219.677ms**
Average syncProxyRules time on a randomly-selected node in the last e2e-aws run on this PR: **104.894ms**

Hard to interpret given that the e2e environment is not really like the large customer environments this is supposed to help, but the answer is somewhere between "it makes all syncProxyRules runs take only 100ms", "it cuts syncProxyRules time in half" and "it only saves 100ms on each syncProxyRules call" :slightly_smiling_face:

PR #406, opened 2022-03-07
Bug 2060553: another fix for mixed ingress and egress policies

danwinship:
The previous fix was incomplete; it covered the case of "some policy isolates all pods for ingress and another policy selects random pods for egress and as a result all pods end up mistakenly isolated for egress", but it didn't cover "some policy isolates _some_ pods for ingress and another policy selects random pods for egress and as a result _some_ pods end up mistakenly isolated for egress".

Confirmed that the new unit tests fails without this fix.

### DIFF
--- a/pkg/network/node/networkpolicy.go
+++ b/pkg/network/node/networkpolicy.go
@@ -357,22 +357,21 @@ func (np *networkPolicyPlugin) generateNamespaceFlows(otx ovs.Transaction, npns 
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


wking:
/retitle Bug 2060553: WIP another fix for mixed ingress and egress policies

wking:
/retitle WIP: Bug 2060553: another fix for mixed ingress and egress policies

PR #408, opened 2022-03-08
Bug 2061916: mixed ingress and egress policies can result in half-isolated pods

danwinship:
If a namespace contained no policies that applied to the entire namespace, then pods which were selected by at least one ingress-only policy and at least one egress-only policy could end up only half-isolated, such that we would either allow ingress to the pod that should have been blocked, or allow egress from it that should have been blocked.

The basic problem with the last two bugs and this one is that we were trying to combine the ingress-rule-writing and egress-rule-writing together, but it's really hard to do that right. (Specifically, here, the problem is that we needed separate ingress and egress `selectedIPs` sets.) So this splits them apart, which should hopefully work better.

### DIFF
--- a/pkg/network/node/networkpolicy.go
+++ b/pkg/network/node/networkpolicy.go
@@ -375,10 +375,6 @@ func (np *networkPolicyPlugin) generateNamespaceFlows(otx ovs.Transaction, npns 
 			}
 		}
 
-		if np.skipIfTooManyFlows(&npp.policy, len(npp.ingressFlows)+len(npp.egressFlows)) {
-			continue
-		}
-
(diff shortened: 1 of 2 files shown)


danwinship:
> The only question I had was whether we should put the table numbers as constants so we don't have to have 27 and 80 hard-coded in a bunch of places.

@knobunc I tried this years ago (https://github.com/openshift/origin/commit/479264578ca5e96c31749e65fbaaeee3a6320721) but [the reviewer didn't like how it turned out](https://github.com/openshift/origin/pull/10190#issuecomment-247021358) :slightly_smiling_face:

PR #411, opened 2022-03-15
Bug 2052332: use pods cache during reattach processing 

msherif1234:
Signed-off-by: Mohamed Mahmoud <mmahmoud@redhat.com>
test results
```
OCP 4.9 image:
 
******************sdn-84zqg*************
Plugin Start: 17:28:15.680695
Plugin Ready: 17:28:20.678844
Time diff(s): 5
******************sdn-j6wpf*************
Plugin Start: 17:28:16.139822
Plugin Ready: 17:29:43.829619
Time diff(s): 87
******************sdn-qnjmf*************
Plugin Start: 17:30:57.958794
Plugin Ready: 17:32:12.226418
Time diff(s): 75
******************sdn-qnqdg*************
Plugin Start: 17:33:07.827914
Plugin Ready: 17:34:27.394670
Time diff(s): 80
******************sdn-qz2gm*************
Plugin Start: 17:30:11.963915
Plugin Ready: 17:30:24.910544
Time diff(s): 13
******************sdn-z87rs*************
Plugin Start: 17:28:31.051204
Plugin Ready: 17:28:56.230275
Time diff(s): 25
 
 
Clean informer image:
******************sdn-75mnp*************
Plugin Start: 17:42:13.157691
Plugin Ready: 17:42:19.850444
Time diff(s): 6
******************sdn-7c5q8*************
Plugin Start: 17:40:03.401974
Plugin Ready: 17:40:38.659059
Time diff(s): 35
******************sdn-bjjl5*************
Plugin Start: 17:42:57.667644
Plugin Ready: 17:43:03.719098
Time diff(s): 6
******************sdn-df5kk*************
Plugin Start: 17:36:11.333512
Plugin Ready: 17:36:41.379830
Time diff(s): 30
******************sdn-hbgmj*************
Plugin Start: 17:38:06.244015
Plugin Ready: 17:38:40.667368
Time diff(s): 34
******************sdn-z2qzk*************
Plugin Start: 17:43:43.855341
Plugin Ready: 17:43:50.133778
Time diff(s): 7
```

### DIFF
--- a/pkg/network/node/pod.go
+++ b/pkg/network/node/pod.go
@@ -67,7 +67,8 @@ type podManager struct {
 
 	// Things only accessed through the processCNIRequests() goroutine
 	// and thus can be set from Start()
-	ipamConfig []byte
+	ipamConfig   []byte
+	reattachPods map[string]*corev1.Pod
 }
 
 // Creates a new live podManager; used by node code0
@@ -85,8 +86,9 @@ func newPodManager(kClient kubernetes.Interface, policy osdnPolicy, overlayMTU u
 // Creates a new basic podManager; used by testcases
 func newDefaultPodManager() *podManager {
 	return &podManager{
-		runningPods: make(map[string]*runningPod),
-		requests:    make(chan *cniserver.PodRequest, 20),
+		runningPods:  make(map[string]*runningPod),
+		requests:     make(chan *cniserver.PodRequest, 20),
+		reattachPods: make(map[string]*corev1.Pod),
 	}
 }
 
@@ -203,6 +205,19 @@ func getPodKey(namespace, name string) string {
 
 func (m *podManager) getPod(request *cniserver.PodRequest) *runningPod {
 	return m.runningPods[getPodKey(request.PodNamespace, request.PodName)]
+}
+
+// Add pods to Pod Manager reattach pod cache
+func (m *podManager) setReattachPodsCache(pods []*corev1.Pod) {
+	for _, pod := range pods {
+		pKey := getPodKey(pod.Namespace, pod.Name)
+		m.reattachPods[pKey] = pod
+	}
+}
+
+// Delete the Pod Manager reattach pods cache
+func (m *podManager) clearReattachPodsCache() {
+	m.reattachPods = make(map[string]*corev1.Pod)
 }
 
 // Add a request to the podManager CNI request queue
@@ -470,7 +485,8 @@ func podIsExited(p *kcontainer.Pod) bool {
 // Set up all networking (host/container veth, OVS flows, IPAM, loopback, etc)
 func (m *podManager) setup(req *cniserver.PodRequest) (cnitypes.Result, *runningPod, error) {
 	defer metrics.PodOperationsLatency.WithLabelValues(metrics.PodOperationSetup).Observe(metrics.SinceInMicroseconds(time.Now()))
-
+	var v1Pod *corev1.Pod
+	var err error
 	// Release any IPAM allocations if the setup failed
 	var success bool
 	defer func() {
@@ -478,10 +494,11 @@ func (m *podManager) setup(req *cniserver.PodRequest) (cnitypes.Result, *running
 			m.ipamDel(req.SandboxID)
 		}
 	}()
-
-	v1Pod, err := m.kClient.CoreV1().Pods(req.PodNamespace).Get(context.TODO(), req.PodName, metav1.GetOptions{})
-	if err != nil {
-		return nil, nil, err
+	pKey := getPodKey(req.PodNamespace, req.PodName)
+	if v1Pod, success = m.reattachPods[pKey]; !success {
+		if v1Pod, err = m.kClient.CoreV1().Pods(req.PodNamespace).Get(context.TODO(), req.PodName, metav1.GetOptions{}); err != nil {
+			return nil, nil, err
+		}
 	}
 
 	var ipamResult cnitypes.Result
--- a/pkg/cmd/openshift-sdn-node/cmd.go
+++ b/pkg/cmd/openshift-sdn-node/cmd.go
@@ -174,8 +174,7 @@ func (sdn *openShiftSDN) start(stopCh <-chan struct{}) error {
 	klog.Infof("Starting node networking (%s)", version.Get().String())
 
 	serviceability.StartProfiler()
-	err := sdn.runSDN()
-	if err != nil {
+	if err := sdn.runSDN(); err != nil {
 		return err
 	}
 	proxyInitChan := make(chan bool)
--- a/pkg/network/node/node.go
+++ b/pkg/network/node/node.go
@@ -13,6 +13,7 @@ import (
 
 	metrics "github.com/openshift/sdn/pkg/network/node/metrics"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	kwait "k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apimachinery/pkg/watch"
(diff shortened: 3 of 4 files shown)


pkg/network/node/pod.go
```diff
@@ -79,6 +86,8 @@ func newPodManager(kClient kubernetes.Interface, policy osdnPolicy, overlayMTU u
 	pm.routableMTU = routableMTU
 	pm.podHandler = pm
 	pm.ovs = ovs
+	pm.podInformers = kinformers.NewSharedInformerFactory(kClient, defaultPodInformerResyncPeriod)
```
squeed, line 89: This informer is just used for pods on this node, right? So you can actually use a FieldSelector to limit the pods that we watch here. That will save a lot of memory.

Here's an example for how to do this: https://github.com/ovn-org/ovn-kubernetes/blob/master/go-controller/pkg/factory/factory.go#L259

pkg/network/node/pod_test.go
```diff
@@ -404,10 +410,14 @@ func TestDirectPodUpdate(t *testing.T) {
...
 	podManager := newDefaultPodManager()
+	kubeClient := fake.NewSimpleClientset()
 	podManager.podHandler = podTester
+	podManager.podInformers = kinformers.NewSharedInformerFactory(kubeClient, defaultPodInformerResyncPeriod)
+	podManager.listerSynced = podManager.podInformers.Core().V1().Pods().Informer().HasSynced
+	ch := make(chan struct{})
 	_, clusterCIDR, _ := net.ParseCIDR("1.2.0.0/16")
 	_, serviceCIDR, _ := net.ParseCIDR("172.30.0.0/16")
-	err = podManager.Start(tmpDir, "1.2.3.0/24", []common.ParsedClusterNetworkEntry{{ClusterCIDR: clusterCIDR, HostSubnetLength: 8}}, serviceCIDR, string(configv1.AWSPlatformType))
+	err = podManager.Start(ch, tmpDir, "1.2.3.0/24", []common.ParsedClusterNetworkEntry{{ClusterCIDR: clusterCIDR, HostSubnetLength: 8}}, serviceCIDR, string(configv1.AWSPlatformType))
```
squeed, line 420: Do you need to add the generated Pod object here to the podManager's fake client?

msherif1234: @squeed can u elaborate here more I am not following what u want to add ?

pkg/network/node/pod.go
```diff
@@ -67,18 +72,27 @@ type podManager struct {
...
 	pm := newDefaultPodManager()
 	pm.kClient = kClient
 	pm.policy = policy
 	pm.overlayMTU = overlayMTU
 	pm.routableMTU = routableMTU
 	pm.podHandler = pm
 	pm.ovs = ovs
+	pm.podInformers = kinformers.NewSharedInformerFactoryWithOptions(kClient, defaultPodInformerResyncPeriod,
+		informers.WithTweakListOptions(func(options *metav1.ListOptions) {
+			options.FieldSelector = fields.OneTermEqualSelector("spec.nodeName", nodeName).String()
```
danwinship, line 93: The process already has an all-pods watcher (in networkpolicy.go) so it would be more efficient to just use the same shared informer in this file (and manually filter by node name in the callback)

msherif1234: this is customized informer to just fill up the cache to speed up the reattach without registered event handler

pkg/network/node/pod.go
```diff
@@ -479,11 +498,16 @@ func (m *podManager) setup(req *cniserver.PodRequest) (cnitypes.Result, *running
 		}
 	}()
 
-	v1Pod, err := m.kClient.CoreV1().Pods(req.PodNamespace).Get(context.TODO(), req.PodName, metav1.GetOptions{})
-	if err != nil {
+	podLister := m.podInformers.Core().V1().Pods().Lister()
+	if err = wait.PollImmediateWithContext(context.Background(), 10*time.Millisecond, 5*time.Second, func(ctx context.Context) (bool, error) {
```
danwinship, line 502: So this is the really ugly part... we're converting the code to using informers/listers... except that semantically, this code can't actually be based on informers, because it needs to guarantee it gets the latest pod state from the apiserver, not "whatever state the informer currently knows about".

And especially, the performance problems this PR is trying to fix only exist in one weird case (upgrade requiring a bridge recreation), but the "awkward" part of this PR (looping while waiting for the informer to catch up) isn't needed in that case (all of the pods to be reattached will already have been cached by the informer), and instead will only get used in the "normal" pod startup case, where this PR doesn't actually help at all anyway. :slightly_frowning_face: 

So, basically the idea of this PR is, if we're doing a bunch of pod setups all at once, it would be a lot better to make a single API call to get all of the pod objects rather than doing a separate call for each one... maybe we should do that explicitly?

1. add a new `podManager.cachePods(pods []*corev1.Pod)` to store an array of pods
2. modify `podManager.setup` to look for the pod it wants in the pod manager's list of cached pods first, and only call `Get()` if the pod isn't there.
3. have the pod reattach code call `pods, err := common.ListPodsInNodeAndNamespace(ctx, node.kClient, node.hostName, metav1.NamespaceAll)` and then `podManager.cachePods(pods)` before it starts reattaching pods, and `podManager.cachePods(nil)` after it finishes.

I think that would end up being simpler/clearer?

danwinship (changes requested):
OK, I looked into the idea of redoing sdn startup so everything waits for informers to sync, and I think it would probably involve too much rewriting. So...

pkg/network/node/pod.go
```diff
@@ -205,6 +207,26 @@ func (m *podManager) getPod(request *cniserver.PodRequest) *runningPod {
...
+	}
+}
+
+// Delete pods from Pod Manager reattach pod cache
+func (m *podManager) delPodsFromCache(pods []*corev1.Pod) {
+	for _, pod := range pods {
+		pKey := getPodKey(pod.Namespace, pod.Name)
+		if _, exists := m.reattachPods[pKey]; exists {
+			delete(m.reattachPods, pKey)
+		}
```
danwinship, line 226:
```suggestion
		delete(m.reattachPods, pKey)
```

pkg/network/node/pod.go
```diff
@@ -67,7 +67,8 @@ type podManager struct {
 
 	// Things only accessed through the processCNIRequests() goroutine
 	// and thus can be set from Start()
-	ipamConfig []byte
+	ipamConfig   []byte
+	reattachPods map[string]*corev1.Pod
```
danwinship, line 71: We should be consistent about the naming; if it's just for reattaching, then `addPodsToCache` / `delPodsFromCache` should reflect that too. OTOH if it's not just for reattaching then this field should have a more generic name.

Making it only for reattaching might make sense... In that case there's no need for `delPodsFromCache` to take an argument, since the only allowed call would be to clear the whole cache. So it could be like `setReattachPodsCache(pods)` / `clearReattachPodsCache()`. Or `startReattachPods(pods)` / `endReattachPods()` or something

pkg/network/node/pod.go
```diff
@@ -205,6 +207,26 @@ func (m *podManager) getPod(request *cniserver.PodRequest) *runningPod {
 	return m.runningPods[getPodKey(request.PodNamespace, request.PodName)]
 }
 
+// Add pods to Pod Manager reattach pod cache
+func (m *podManager) addPodsToCache(pods []*corev1.Pod) {
+	for _, pod := range pods {
+		pKey := getPodKey(pod.Namespace, pod.Name)
+		if _, exists := m.reattachPods[pKey]; !exists {
+			m.reattachPods[pKey] = pod
+		}
```
danwinship, line 216: not sure there's any reason to check for exists here either...

mrobson:
Here are the new test timings with the updated PR - looks good.

`******************sdn-6l8kg*************
Plugin Start: 14:54:40.483687
Plugin Ready: 14:54:46.143635
Time diff(s): 6
******************sdn-9jh7c*************
Plugin Start: 14:53:10.938542
Plugin Ready: 14:53:37.671952
Time diff(s): 27
******************sdn-jjg6v*************
Plugin Start: 14:56:14.477243
Plugin Ready: 14:56:20.545520
Time diff(s): 6
******************sdn-t2p57*************
Plugin Start: 14:57:25.797061
Plugin Ready: 14:57:55.288396
Time diff(s): 30
******************sdn-tbbpj*************
Plugin Start: 14:58:54.099398
Plugin Ready: 14:59:02.019293
Time diff(s): 8
******************sdn-zztb5*************
Plugin Start: 15:00:20.752163
Plugin Ready: 15:00:49.636883
Time diff(s): 29`

PR #414, opened 2022-03-15
Bug 2063971: delete stale UDP conntrack entries for loadbalancer IPs

danwinship:
Depends on https://github.com/openshift/kubernetes/pull/1214


danwinship:
/assign @aojea 
(Rebasing picked up some Windows kube proxy fixes too but they don't matter.)

PR #417, opened 2022-03-25
Bug 2063123: Remove node-tainting for too-small MTU

danwinship:
The "MTU is too small" case never happens in normal operation, and giving openshift-sdn-node write access to all Nodes just for that is excessive. So get rid of the tainting, and if we find the MTU is too small then just (a) emit an Event, and (b) exit with an error (which will prevent the node from becoming Ready anyway).

### DIFF
--- a/pkg/network/node/node.go
+++ b/pkg/network/node/node.go
@@ -2,7 +2,6 @@ package node
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"net"
 	"strings"
@@ -14,15 +13,12 @@ import (
 
 	metrics "github.com/openshift/sdn/pkg/network/node/metrics"
 	corev1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/apimachinery/pkg/util/strategicpatch"
 	kwait "k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/record"
-	"k8s.io/client-go/util/retry"
 	kubeletapi "k8s.io/cri-api/pkg/apis"
 	kruntimeapi "k8s.io/cri-api/pkg/apis/runtime/v1"
 	kapi "k8s.io/kubernetes/pkg/apis/core"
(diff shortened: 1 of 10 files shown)


danwinship:
/assign @jcaamano 
since you've looked at MTU stuff. (also see https://github.com/openshift/cluster-network-operator/pull/1350 after this)

jcaamano (commented):
>The "MTU is too small" case never happens in normal operation, and giving openshift-sdn-node write access to all Nodes just for that is excessive. So get rid of the tainting, and if we find the MTU is too small then just (a) emit an Event, and (b) exit with an error (which will prevent the node from becoming Ready anyway).

@danwinship I am missing where we are emitting the event and exiting. As far as I can see `validateMTU` error is handled by just logging it. Am I missing something?

danwinship:
"exiting" happens if we return an error from `validateMTU()`; previously that just got logged by `Start()`, but now `Start()` returns the error rather than continuing.

I appear to have forgotten to add the code to emit the event :slightly_smiling_face:

danwinship:
> Lines 434 to 436 in [e3389c4](https://github.com/openshift/sdn/commit/e3389c47500656d6604a8f257478a181bc841129)

that's not this PR...

danwinship:
ok, fixed the commit message to not claim we emit an event; I remembered that I decided to take that part out, because the pod is going to keep getting restarted, but we don't want to keep emitting events...

pkg/network/node/node.go
```diff
@@ -296,71 +294,19 @@ func (node *OsdnNode) validateMTU() error {
...
-
-			_, err = node.kClient.CoreV1().Nodes().Patch(context.TODO(), node.hostName, types.StrategicMergePatchType, patchBytes, metav1.PatchOptions{})
-
-			return err
-		})
-
-		if resultErr != nil {
-			return fmt.Errorf("could not update node taints after many retries: %v", resultErr)
-		}
+	if interfaceMTU < int(node.overlayMTU)+50 {
```
msherif1234, line 307: ok this is the fix u described in this PR, why did u change the error code in the above codes to nil ?

danwinship: `validateMTU` has been changed so that it only returns an error if the MTU is definitely wrong (in which case its caller will cause the program to exit). In all of the cases where it runs into a problem trying to figure out if the MTU is right or wrong, it now just logs an error message and then returns `nil`, assuming that _probably_ the MTU is right and it should keep going.

msherif1234: /lgtm

PR #418, opened 2022-03-29
Bug 2062558: egressip: Continue to process other nodes if a node is not ready

tssurya:
If a node is not ready, continue to process other nodes in the
list after marking it offline instead of returning immediately
since if there are other nodes that are offline at the same
time we need to make sure we mark those nodes as offline as
well to be able to do re-assignment of egressIPs.

/cc @danwinship @kyrtapz @flavio-fernandes


flavio-fernandes (commented):
/lgtm

PR #428, opened 2022-05-12
Bug 2016534: Exclude the default drop bit from egress IP VNID

kyrtapz:
This is needed for egress IP traffic that is DNATed to a local IP(ExternalIP/LoadBalancer).
This type of traffic traverses KUBE-FIREWALL chain which drops packets marked with the default drop bit.

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=2082451#c3

### DIFF
--- a/pkg/network/node/egressip.go
+++ b/pkg/network/node/egressip.go
@@ -21,9 +21,10 @@ import (
 )
 
 const (
-	defaultPollInterval = 5 * time.Second
-	repollInterval      = time.Second
-	maxRetries          = 2
+	defaultPollInterval   = 5 * time.Second
+	repollInterval        = time.Second
+	maxRetries            = 2
+	defaultKubeletDropBit = 1 << uint32(15) // https://github.com/kubernetes/kubelet/blob/release-1.24/config/v1beta1/types.go#L555
 )
 
 type egressNode struct {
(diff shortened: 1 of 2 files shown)


pkg/network/node/egressip.go
```diff
@@ -134,6 +135,9 @@ func getMarkForVNID(vnid, masqueradeBit uint32) string {
 	if (vnid & masqueradeBit) != 0 {
 		vnid = (vnid | 0x01000000) ^ masqueradeBit
 	}
+	if (vnid & defaultKubeletDropBit) != 0 {
+		vnid = (vnid | 0x10000000) ^ defaultKubeletDropBit
```
kyrtapz, line 139: What if `masqueradeBit` is equal to `0x10000000`

danwinship:
/lgtm

you should be able to test this (both that it fails without this patch and succeeds with it) by just creating namespaces until you get one with the right bit set. (sdn-controller will get confused if you try to change/set NetNamespace.NetID by hand)

kyrtapz:
Was able to verify it by upgrading from 4.10(where it doesn't work) to this PR(where it works!):
`test6                                              9750882    ["10.0.128.101"]`
4.10 iptables on egress node:
```
Chain OPENSHIFT-MASQUERADE (1 references)
 pkts bytes target     prot opt in     out     source               destination         
    1    60 SNAT       all  --  *      *       10.128.0.0/14        0.0.0.0/0            mark match 0x94c962 to:10.0.128.101
  112 11406 RETURN     all  --  *      *       0.0.0.0/0            0.0.0.0/0            mark match 0x1/0x1
 9282  795K OPENSHIFT-MASQUERADE-2  all  --  *      *       10.128.0.0/14        0.0.0.0/0            /* masquerade pod-to-external traffic */
```
4.11(with this PR) iptables on egress node:
```
Chain OPENSHIFT-MASQUERADE (1 references)
 pkts bytes target     prot opt in     out     source               destination         
    8   480 MASQUERADE  all  --  *      *       10.128.0.0/14        10.128.0.0/14        mark match 0x10944962
    0     0 SNAT       all  --  *      *       10.128.0.0/14        0.0.0.0/0            mark match 0x10944962 to:10.0.128.101
   54  5484 RETURN     all  --  *      *       0.0.0.0/0            0.0.0.0/0            mark match 0x1/0x1
 6493  511K OPENSHIFT-MASQUERADE-2  all  --  *      *       10.128.0.0/14        0.0.0.0/0            /* masquerade pod-to-external traffic */

```

PR #431, opened 2022-05-18
Bug 2087763: Probe failures and pod restarts during 4.7 to 4.8 upgrade

msherif1234:
cherry-pick #411
cherry-pick #400


jcaamano:
I would probably avoid cherry-picking #400 and use the equivalent of `ListPodsInNodeAndNamespace` as it was originally done before #400 given that the associated BZ is not verified.

msherif1234:
> I would probably avoid cherry-picking #400 and use the equivalent of `ListPodsInNodeAndNamespace` as it was originally done before #400 given that the associated BZ is not verified.

do u have concern bringing in #400 it seems like shouldn't cause any harm ?

jcaamano:
@msherif1234 I plan to add

> > I would probably avoid cherry-picking #400 and use the equivalent of `ListPodsInNodeAndNamespace` as it was originally done before #400 given that the associated BZ is not verified.
> 
> do u have concern bringing in #400 it seems like shouldn't cause any harm ? I was thinking of easier way to make backport less painful @danwinship WDYT ?

It shouldn't do any more harm than what is already there, but I plan to add a new PR to that BZ so that the listings are done with `resourceVersion=0`, that way they are served by the api server watch cache. Otherwise they are being served by etcd, which is a more critical component, just as they were before, although now paginated, I guess an improvement.

Conceptually I would not backport including changes of an unverified BZ, but in this specific case I don't think it would do any harm. Your call. If you want to go ahead with it I will LGTM this.

msherif1234:
> @msherif1234 I plan to add
> 
> > > I would probably avoid cherry-picking #400 and use the equivalent of `ListPodsInNodeAndNamespace` as it was originally done before #400 given that the associated BZ is not verified.
> > 
> > 
> > do u have concern bringing in #400 it seems like shouldn't cause any harm ? I was thinking of easier way to make backport less painful @danwinship WDYT ?
> 
> It shouldn't do any more harm than what is already there, but I plan to add a new PR to that BZ so that the listings are done with `resourceVersion=0`, that way they are served by the api server watch cache. Otherwise they are being served by etcd, which is a more critical component, just as they were before, although now paginated, I guess an improvement.
> 
> Conceptually I would not backport including changes of an unverified BZ, but in this specific case I don't think it would do any harm. Your call. If you want to go ahead with it I will LGTM this.
I removed your commit from my PR and called the list api directly

pkg/network/node/node.go
```diff
@@ -450,6 +450,18 @@ func (node *OsdnNode) Start() error {
 // attached to the OVS bridge before restart, and either reattaches or kills each of the
 // corresponding pods.
 func (node *OsdnNode) reattachPods(existingPodSandboxes map[string]*kruntimeapi.PodSandbox, existingOFPodNetworks map[string]podNetworkInfo) error {
+	fieldSelector := fields.Set{"spec.nodeName": node.hostName}.AsSelector()
+	opts := metav1.ListOptions{
+		LabelSelector: labels.Everything().String(),
+		FieldSelector: fieldSelector.String(),
+	}
+	pods, err := node.kClient.CoreV1().Pods(metav1.NamespaceAll).List(context.TODO(), opts)
```
danwinship, line 458: hm... yeah, I know I said "just call that", but I forgot you needed to set up the fieldSelector too...

I think you could just call GetRunningPods(). It does a little bit more filtering than this, but filtering the list down to only pods that are still running is probably a good thing, since pods that aren't `Running` aren't going to have sandboxes to reattach...

PR #432, opened 2022-05-25
Bug 2067865: Bump prometheus/client_golang to 1.11.1

martinkennelly:

go mod edit -replace ..
go mod tidy
go mod vendor


martinkennelly:
Expected to see updates in vendor directory, but none seen. I would like to understand why.

danwinship:
go.mod has a `replace` entry for it:

	github.com/prometheus/client_golang => github.com/prometheus/client_golang v1.7.1

sdn's go.mod is kind of insane... it should not be doing all that replacing, it should be requiring the versions it wants in the first place... This is probably a bad conversion from whathever it was we were using before go.mod (glide?)..

go.mod
```diff
@@ -64,7 +64,7 @@ replace (
 	github.com/peterbourgon/diskv => github.com/peterbourgon/diskv v2.0.1+incompatible
 	github.com/pkg/errors => github.com/pkg/errors v0.9.1
 	github.com/pkg/profile => github.com/pkg/profile v1.3.0
-	github.com/prometheus/client_golang => github.com/prometheus/client_golang v1.7.1
+	github.com/prometheus/client_golang => github.com/prometheus/client_golang v1.11.1
```
danwinship, line 67: just remove this; there's no reason to specify a `replace` if you aren't trying to override the `require` section

PR #434, opened 2022-05-28
DNM testing upstream kube-proxy patch

danwinship:
/cc


danwinship:
FTR: doesn't work right, because gcp-routes's rules end up rewriting the packet before we can drop it. The old `KUBE-MARK-DROP` approach works because we mark it in `nat` before gcp-routes modifies it and the mark is still there after the packet is rewritten...

PR #436, opened 2022-06-06
[release-4.9] Bug 2093978: Remove node-tainting for too-small MTU

danwinship:
The "MTU is too small" case never happens in normal operation, and giving openshift-sdn-node write access to all Nodes just for that is excessive. So get rid of the tainting, and if we find the MTU is too small then just exit with an error (which will prevent the node from becoming Ready anyway).

4.9 backport of #422 / #417


danwinship:
CNO PR: https://github.com/openshift/cluster-network-operator/pull/1476

PR #437, opened 2022-06-06
WIP enable UDP GRO

danwinship:
/cc
testing...


danwinship:
We're going to enable this by default in ovn-kubernetes. I only did the sdn version because ovn-k builds were broken for a few days last week so I couldn't test there. It's not clear if we're going to enable it in openshift-sdn too. (It doesn't seem to help as much in sdn as in ovn-k anyway.)

danwinship:
> (It doesn't seem to help as much in sdn as in ovn-k anyway.)

This probably indicates that there's a bug in the PR... anyway, I'm going to submit something related to this upstream to `containernetworking/plugins` and then if it merges there maybe it can eventually make its way into sdn...

PR #440, opened 2022-06-13
Bug 2089350: kube 1.24 rebase

danwinship:
Pulls in https://github.com/danwinship/kubernetes/commits/sdn-4.11-kubernetes-1.24.0, which I'll merge if this passes tests/review.

Contrary to what I said in https://github.com/openshift/kubernetes/pull/1278, we actually do need to keep the modifications to the upstream userspace proxy, because we still allow people to configure sdn to use the userspace proxy... if they're using multitenant mode. Sigh. Anyway, so that's still needed for now.


danwinship:
cc @bpickard22

danwinship:
/retest-required

i can bring up an AWS cluster with this PR by hand... the problem here seems to be with CI, not with the PR

kyrtapz:
I don't understand why do we need https://github.com/openshift/sdn/pull/440/commits/5c451aa1f1828ab5791f738759fa0124f3bd35da.
The new `configmapslease` uses both the `configmap` and `lease` as per: https://github.com/kubernetes/kubernetes/issues/107454#issue-1098220860

ricky-rav:
As for the leader election changes in 1.24, my understanding is that we are migrating to the new API across three releases, so that version x is compatible with version x+1 and upgrades won't fail.

So for ovn-k rebase, I only did this, which is the required step for 1.24: I only did this https://github.com/ovn-org/ovn-kubernetes/pull/3001/commits/99a90f5ae2e8e9969b89b35226b2ba9fcc6313c0

danwinship:
> As for the leader election changes in 1.24, my understanding is that we are migrating to the new API across three releases, so that version x is compatible with version x+1 and upgrades won't fail.

yeah, I found some stuff about that, but I guess I skimmed it too quickly and misunderstood. (I didn't realize `configmapsleases` meant `configmaps`+`leases`, I thought it was a new incompatible type.)

kyrtapz:
Looks good to me

danwinship:
/retest-required
failures seem unrelated (failing on other PRs too)

PR #450, opened 2022-07-25
Bug 2104953: rebase to 1.24, take 2

danwinship:
Rebased to new `sdn-4.11-kubernetes-1.24.3` branch, which includes a fixed implementation of the "local with fallback" hack used by the router, resolving the "remains available" failures in the original rebase.

The problem was with the `local-with-fallback` hack we added for the router to make it behave more reliably during disruption; the traffic policy code changed a ton between 1.23 and 1.24, and I missed a spot when rewriting the `local-with-fallback` code ([this part](https://github.com/openshift/kubernetes/commit/16c693911171be8bb180607a0c0d7bf219d4a96b#diff-2ccafd63d57233fd0697b9852514604ab880e29aeb073035765ad7541cad1b35L1126)) with the result that local-with-fallback traffic would get DNATted and forwarded to another node without also being SNATted, and so the return traffic would go directly back to the client without being un-DNATted, and so would get dropped.

The use of the local detector to masquerade external traffic but not internal is to make the `HAProxy router should set Forwarded headers appropriately` test in origin pass reliably; without that exception the test sometimes fails because the request appears to come from the node IP rather than the pod IP. (Though arguably we should just not be expecting the source IP to be preserved reliably when using local-with-fallback, and maybe I'll change the test...)

Passing payload e2e-gcp-upgrade run: https://pr-payload-tests.ci.openshift.org/runs/ci/4b896440-0d30-11ed-8d2f-99045a084268-0


danwinship:
(Note that the 4.12 CI payload doesn't include the e2e-gcp-upgrade job so there's no reason to run it here; we have to rely on the fact that the release-4.11 version in #451 passed.)

danwinship:
via cluster-bot:
job [workflow-upgrade openshift-upgrade-gcp 4.11 openshift/sdn#450](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp-modern/1552323790061441024) succeeded

go.mod~
```diff
@@ -0,0 +1,103 @@
+module github.com/openshift/sdn
```
ricky-rav, line 1: This is the backup file created by... emacs? :)

danwinship: bad `git add .` !

PR #454, opened 2022-08-10
OCPBUGS-69: Emit event when egressIP reaches capacity limit

pperiyasamy:
This would emit an event log into k8s when number of egressIP added
to hostsubnet reaches the capability limit.

### DIFF
--- a/pkg/network/common/egressip.go
+++ b/pkg/network/common/egressip.go
@@ -116,8 +116,8 @@ type EgressIPTracker struct {
 
 	watcher EgressIPWatcher
 
-	kubeClient kubernetes.Interface
-
+	kubeClient       kubernetes.Interface
+	localIP          string
 	nodes            map[ktypes.UID]*nodeEgress
 	nodesByNodeIP    map[string]*nodeEgress
 	namespacesByVNID map[uint32]*namespaceEgress
@@ -132,7 +132,7 @@ type EgressIPTracker struct {
 	isFirstEvent      bool
 }
 
-func NewEgressIPTracker(watcher EgressIPWatcher, cloudEgressIP bool, eventRecorder record.EventRecorder) *EgressIPTracker {
+func NewEgressIPTracker(watcher EgressIPWatcher, cloudEgressIP bool, localIP string, eventRecorder record.EventRecorder) *EgressIPTracker {
 	return &EgressIPTracker{
 		watcher: watcher,
 
@@ -145,6 +145,7 @@ func NewEgressIPTracker(watcher EgressIPWatcher, cloudEgressIP bool, eventRecord
 
 		changedEgressIPs:  make(map[*egressIPInfo]bool),
 		changedNamespaces: make(map[*namespaceEgress]bool),
+		localIP:           localIP,
 		recorder:          eventRecorder,
 		start:             time.Now(),
 		isFirstEvent:      true,
@@ -259,7 +260,8 @@ func (eit *EgressIPTracker) handleAddOrUpdateHostSubnet(obj, _ interface{}, even
 			return
 		}
 	}
-	if eit.recorder != nil && len(hs.EgressIPs) > 0 && eit.CloudEgressIP {
+	if eit.localIP != "" && eit.localIP == hs.HostIP &&
+		eit.recorder != nil && len(hs.EgressIPs) > 0 && eit.CloudEgressIP {
 		capacity, err := eit.getCapacity(hs)
 		if err != nil {
 			klog.Errorf("Ignoring invalid HostSubnet %s: %v", HostSubnetToString(hs), err)
--- a/pkg/network/common/egressip.go
+++ b/pkg/network/common/egressip.go
@@ -23,6 +23,7 @@ import (
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/record"
 
 	osdnv1 "github.com/openshift/api/network/v1"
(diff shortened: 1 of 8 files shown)


flavio-fernandes:
cool beans! @danwinship PTAL

pkg/network/common/egressip.go
```diff
@@ -249,6 +252,17 @@ func (eit *EgressIPTracker) handleAddOrUpdateHostSubnet(obj, _ interface{}, even
 			return
 		}
 	}
+	if eit.recorder != nil && len(hs.EgressIPs) > 0 && eit.CloudEgressIP {
+		capacity, err := eit.getCapacity(hs)
+		if err != nil {
+			klog.Errorf("Ignoring invalid HostSubnet %s: %v", HostSubnetToString(hs), err)
+		}
+		if len(hs.EgressIPs) >= capacity {
```
danwinship, line 260: it's awkward that this test runs even if `getCapacity` returns an error... would be safer to have an `else` here

pkg/network/common/egressip.go
```diff
@@ -249,6 +252,17 @@ func (eit *EgressIPTracker) handleAddOrUpdateHostSubnet(obj, _ interface{}, even
...
+			hsRef := &corev1.ObjectReference{Kind: "HostSubnet", Name: hs.Name, UID: hs.ObjectMeta.UID}
+			eit.recorder.Eventf(hsRef, corev1.EventTypeNormal, "HostSubnet",
+				"The Hostsubnet reached its maximum node capacity %d", capacity)
```
danwinship, line 263: Maybe use the Node as the object reference rather than the HostSubnet? The HostSubnet is really just a weird implementation detail...

danwinship, line 263: So it's hard to figure out with the `pkg/network/common/egressip.go` code, but does this get logged by the master, the workers, or both? It should only get logged in one place; either by the master, or by the specific node that has reached capacity.

Also, we need to make sure that it doesn't spam events... we want an event when the node _reaches_ capacity, but we shouldn't log again after that unless it drops below capacity for a while and then hits it again.

Maybe a metric would be better than an Event, and then the admin can decide how to monitor the metric?

danwinship, line 263: This event is not at all clear about what it means. In particular, it doesn't even say that it's talking about egress IPs.

pperiyasamy: @danwinship I guess this code always gets executed in node upon host subnet create and update and this node capacity event is triggered at the last egress ip. now added 5 mins interval gap to avoid spam. is this fine ? 
Moreover it's done here because event recorder is readily there from node code and consumed here  (master doesn't have event handler and it would just ignore this).

pperiyasamy: correct, changed log statement, is that ok ?

pperiyasamy:
I've tested this on an OCP cluster and event trigger happens as expected. But it generates same event from each node. is that ok ?

pperiyasamy:
> I've tested this on an OCP cluster and event trigger happens as expected. But it generates same event from each node. is that ok ?

It is fixed with commit https://github.com/openshift/sdn/pull/454/commits/d02d7f19f2a67c339df8a902a46279d71c265f32 and can see only one event generated from particular node.

flavio-fernandes:
@danwinship PTAL if you can

pperiyasamy:
@danwinship is this good to go in now ?

danwinship:
It really seems like this ought to be a metric indicating the number of egress IPs (or the percentage "full" that the node is or something)... Then rather than getting just a binary "OK" vs "Already Too Late", the admin can monitor when they are _almost_ out of egress IPs and add another node or something like that. Etc.

But I wasn't involved in the design of the feature. If there's a good reason for doing it like this, then OK. I'll approve but someone else should lgtm.

/approve

pperiyasamy:
> It really seems like this ought to be a metric indicating the number of egress IPs (or the percentage "full" that the node is or something)... Then rather than getting just a binary "OK" vs "Already Too Late", the admin can monitor when they are _almost_ out of egress IPs and add another node or something like that. Etc.

Sure, The event is similar to what ovn-k does when egress ip exceeds its limit on node, hence [OCPBUGS-69](https://issues.redhat.com/browse/OCPBUGS-69) wanted similar fix in sdn. yes, metrics may give appropriate warning before egress ips are fully used on a node. This commit https://github.com/openshift/sdn/pull/454/commits/2213eb02063e1ba75262e2898e11036a7cada66e introduces `AssignedEgressIP` and `EgressIPCapacity` metrics which can be useful to calculate percentage.

pkg/network/common/egressip.go
```diff
@@ -1008,14 +1023,23 @@ func (eit *EgressIPTracker) ReallocateEgressIPs() map[string][]string {
...
 	metrics.RecordEgressIPCount(activeEgressIPCount)
+	if eit.localIP != "" && eit.localIP == hs.HostIP && eit.CloudEgressIP {
+		capacity, err := eit.getCapacity(hs)
+		if err != nil {
+			klog.Errorf("Ignoring invalid HostSubnet %s for node metrics: %v", HostSubnetToString(hs), err)
+			return
+		}
+		nodemetrics.EgressIPCapacity.WithLabelValues(hs.Host).Set(float64(capacity))
+		nodemetrics.AssignedEgressIP.WithLabelValues(hs.Host).Set(float64(len(hs.EgressIPs)))
+	}
```
danwinship, line 1042: You don't need to label the metric with the hostname; each openshift-sdn-node is only reporting the metric for its own HostIP, right?

pperiyasamy: That's correct. changed it.

danwinship:
/skip
(I think that will poke prow to notice that all currently-in-use tests have passed)

PR #459, opened 2022-09-26
OCPBUGS-256: fix network policy egress

JacobTanenbaum:
in the case where an egress rule has no namespaceSelector and an empty pod Selector we generate a flow fmt.Sprintf("reg1=%d, ", npns.vnid), that is wrong because reg1 is not set at this point.

the egress flows need to be generated using only podIPs rather then VNIDs for egress flows


danwinship:
/hold
e2e tests are actually running against ovn-kubernetes :slightly_frowning_face:

PR #462, opened 2022-10-04
OCPBUGS-1783: check if there is a connection in conntrack when performing egressIP

JacobTanenbaum:
Before going into the buckets check if there is a conntrack entry already established and follow that before going into the buckets for egressIP. This should prevent routing on a packet by packet basis

### DIFF
--- a/pkg/network/node/ovscontroller.go
+++ b/pkg/network/node/ovscontroller.go
@@ -939,6 +939,7 @@ func (oc *ovsController) ensureTunMAC() error {
 func (oc *ovsController) SetNamespaceEgressNormal(vnid uint32) error {
 	otx := oc.ovs.NewTransaction()
 	otx.DeleteFlows("table=101, reg0=%d", vnid)
+	otx.DeleteFlows("table=125, reg0=%d", vnid)
 	otx.DeleteGroup(vnid)
 	return otx.Commit()
 }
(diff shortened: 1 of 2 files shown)


pkg/network/node/ovscontroller.go
```diff
@@ -977,6 +977,8 @@ func (oc *ovsController) SetNamespaceEgressViaEgressIPs(vnid uint32, egressIPsMe
 		// Namespace wants egressIP, but no node hosts it, so drop
 		otx.AddFlow("table=101, priority=100, reg0=%d, actions=drop", vnid)
 	} else {
+		// check if there is a connection already stored in conntrack
+		otx.AddFlow("table=101, priority=101, reg0=%d, ct_state=+est, actions=move:NXM_NX_CT_NW_DST[]->NXM_NX_TUN_IPV4_DST[],output:vxlan0", vnid)
```
danwinship, line 981: So I had suggested this, but after thinking some more, I think it's wrong; `ct_nw_dst` will be the `nw_dst` of the original connection (ie, the external IP we're using the egress IP to reach), whereas we want the `tunnel_dst` of the original connection, which doesn't seem to be stored in conntrack...

I think we could store it in `ct_mark` though... we'd have to copy it there explicitly when sending the first packet, and then copy it back out when sending later packets.

JacobTanenbaum: done

PR #463, opened 2022-10-05
OCPBUGS-95: Add node egress IP assignment resync

kyrtapz:
Sync node egress IP assignment every time an egress IP is removed from the interface.

Even though the IPs should be reassigned quickly there is still a chance for some disruption to the traffic using egress IP.

/cc @jcaamano @cybertron @danwinship 

### DIFF
--- a/pkg/network/node/egressip.go
+++ b/pkg/network/node/egressip.go
@@ -50,7 +50,9 @@ type egressIPWatcher struct {
 
 	monitorNodesLock sync.Mutex
 	monitorNodes     map[string]*egressNode
-	stop             chan struct{}
+	stopMonitorNodes chan struct{}
+
+	stopIPResync chan struct{}
 
 	testModeChan chan string
 }
@@ -85,34 +87,49 @@ func (eip *egressIPWatcher) Start(osdnInformers osdninformers.SharedInformerFact
 	return nil
 }
 
-func (eip *egressIPWatcher) Synced() {
+func (eip *egressIPWatcher) SyncIPAssignments() error {
 	link, _, err := GetLinkDetails(eip.localIP)
 	if err != nil {
-		// shouldn't happen, but obviously there's nothing to clean up...
-		return
+		return fmt.Errorf("could not get %s link details %v", eip.localIP, err)
 	}
 	label, err := egressIPLabel(link)
 	if err != nil {
-		klog.Errorf("Could not check for stale egress IPs: %v", err)
-		return
+		return fmt.Errorf("invalid egress IP label: %v", err)
 	}
 	addrs, err := netlink.AddrList(link, netlink.FAMILY_V4)
 	if err != nil {
-		klog.Errorf("Could not check for stale egress IPs: %v", err)
-		return
+		return fmt.Errorf("could not check for egress IPs: %v", err)
 	}
 
+	existingIPs := sets.NewString()
 	for _, addr := range addrs {
 		ip := addr.IP.String()
-		if addr.Label == label && eip.iptablesMark[ip] == "" {
-			klog.Infof("Cleaning up stale egress IP %s", addr.IP.String())
-			err = netlink.AddrDel(link, &addr)
-			if err != nil {
-				klog.Errorf("Could not clean up stale egress IP: %v", err)
+		if addr.Label == label {
+			if eip.iptablesMark[ip] == "" {
+				klog.Infof("Cleaning up stale egress IP %s", addr.IP.String())
+				err = netlink.AddrDel(link, &addr)
+				if err != nil {
+					klog.Errorf("Could not clean up stale egress IP: %v", err)
+				}
+			} else {
+				existingIPs.Insert(ip)
 			}
 		}
 	}
+	for ip, mark := range eip.iptablesMark {
+		if !existingIPs.Has(ip) {
+			klog.Infof("Assigning missing egress IP %s", ip)
+			eip.assignEgressIP(ip, mark)
+		}
+	}
+	return nil
+}
 
+func (eip *egressIPWatcher) Synced() {
+	if err := eip.SyncIPAssignments(); err != nil {
+		klog.Errorf("Failed to sync ip assignments: %v", err)
+		return
+	}
 	eip.iptables.SyncEgressIPRules()
 }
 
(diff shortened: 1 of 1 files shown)


pkg/network/node/egressip.go
```diff
@@ -85,34 +88,50 @@ func (eip *egressIPWatcher) Start(osdnInformers osdninformers.SharedInformerFact
...
+		return fmt.Errorf("invalid egress IP label: %v", err)
 	}
 	addrs, err := netlink.AddrList(link, netlink.FAMILY_V4)
 	if err != nil {
-		klog.Errorf("Could not check for stale egress IPs: %v", err)
-		return
+		return fmt.Errorf("could not check for egress IPs: %v", err)
 	}
 
+	existingIPs := make(map[string]struct{})
```
danwinship, line 106: `existingIPs := sets.NewString()`
where `sets` is `"k8s.io/apimachinery/pkg/util/sets"`, which is already being used in this file

pkg/network/node/egressip.go
```diff
@@ -85,34 +88,50 @@ func (eip *egressIPWatcher) Start(osdnInformers osdninformers.SharedInformerFact
 	return nil
 }
 
-func (eip *egressIPWatcher) Synced() {
+func (eip *egressIPWatcher) SyncIPAssignments() error {
 	link, _, err := GetLinkDetails(eip.localIP)
 	if err != nil {
 		// shouldn't happen, but obviously there's nothing to clean up...
-		return
+		return err
```
danwinship, line 95: comment should be updated, since it implies this will be a no-op, but it's not now

pkg/network/node/egressip.go
```diff
@@ -141,13 +160,76 @@ func getMarkForVNID(vnid, masqueradeBit uint32) string {
...
+	}
+
+	syncIPs := func() {
+		eip.tracker.Lock()
+		defer eip.tracker.Unlock()
+
+		eip.SyncIPAssignments()
+	}
+
+	addressSyncTimer := time.NewTicker(defaultIPAssignmentResyncInterval)
```
danwinship, line 181: We don't need the timer. Trust the netlink notifications. (If we can't trust them, then there's no reason to bother using that API at all...)

pkg/network/node/egressip.go
```diff
@@ -141,13 +160,76 @@ func getMarkForVNID(vnid, masqueradeBit uint32) string {
...
 		eip.iptablesMark[egressIP] = mark
 		if err := eip.assignEgressIP(egressIP, mark); err != nil {
 			klog.Errorf("Error assigning Egress IP %q: %v", egressIP, err)
 		}
+
+		if len(eip.iptablesMark) == 1 {
+			eip.stopIPResync = make(chan struct{})
+			go eip.runIPAssignmentResync(eip.stopIPResync)
+		}
+
```
danwinship, line 232: (extra blank line)

pkg/network/node/egressip.go
kyrtapz: I initially included the timer as a last resort but I am ok with relying on netlink events only

kyrtapz:
@danwinship I addressed your comments, please take a look

pkg/network/node/egressip.go
```diff
@@ -85,34 +87,49 @@ func (eip *egressIPWatcher) Start(osdnInformers osdninformers.SharedInformerFact
...
+					klog.Errorf("Could not clean up stale egress IP: %v", err)
+				}
+			} else {
+				existingIPs.Insert(ip)
 			}
 		}
 	}
+	for ip, mark := range eip.iptablesMark {
+		if !existingIPs.Has(ip) {
+			klog.Infof("Assign missing egress IP %s", ip)
```
danwinship, line 121:
```suggestion
			klog.Infof("Assigned missing egress IP %s", ip)
```

pkg/network/node/egressip.go
```diff
@@ -141,13 +158,70 @@ func getMarkForVNID(vnid, masqueradeBit uint32) string {
...
+		eip.tracker.Lock()
+		defer eip.tracker.Unlock()
+
+		eip.SyncIPAssignments()
+	}
+
+	subscribeErr := addrSubscribe()
+	if subscribeErr != nil {
+		klog.Error("Error during netlink subscribe: %v", subscribeErr)
+	}
```
danwinship, line 187: You need to return if there was an error, right? (If it failed on the first attempt, there's no good reason to expect it would succeed on future attempts...)

pkg/network/node/egressip.go
```diff
@@ -141,13 +158,70 @@ func getMarkForVNID(vnid, masqueradeBit uint32) string {
...
+
+	for {
+		select {
+		case a, ok := <-addrChan:
+			if !ok {
+				if subscribeErr = addrSubscribe(); subscribeErr != nil {
```
danwinship, line 193: you should put the `stopCh` clause before the `addrChan` one, or else it's possible the subscribe thread will process `stopCh` being closed before this thread does, and then you'll try to resubscribe with an already-closed stop channel and get spurious errors, etc, etc

kyrtapz: good point! done

pkg/network/node/egressip.go
kyrtapz: changed to `klog.Infof("Assigning missing egress IP %s", ip)` to match: https://github.com/openshift/sdn/blob/76f2671ee199483a86df10471918a6bf7f857b4a/pkg/network/node/egressip.go#L109

pkg/network/node/egressip.go
kyrtapz: I assumed we wanted to have `runIPAssignmentResync` run in the background the whole time SDN is up, if we return the resync will never happen.

I am not sure how probable the netlink subscribe failure is but is the first attempt any different from the subsequent ones?

pkg/network/node/egressip.go
```diff
@@ -141,13 +158,70 @@ func getMarkForVNID(vnid, masqueradeBit uint32) string {
...
+
+	for {
+		select {
+		case <-stopCh:
+			klog.V(5).Infof("Stopping egress IP assignment resync")
+			return
+		case a, ok := <-addrChan:
+			if !ok {
+				if subscribeErr = addrSubscribe(); subscribeErr != nil {
+					klog.Error("Error during netlink re-subscribe due to address channel closing: %v", subscribeErr)
```
flavio-fernandes, line 197: Go compliance shim [284] [][]: invoking real go binary
# github.com/openshift/sdn/pkg/network/node
pkg/network/node/egressip.go:200:3: k8s.io/klog/v2.Error call has possible formatting directive %v
pkg/network/node/egressip.go:211:6: k8s.io/klog/v2.Error call has possible formatting directive %v
FAIL	github.com/openshift/sdn/pkg/network/node [build failed]
Go compliance shim [284] [][]: Exited with: 1

flavio-fernandes (commented):
@kyrtapz  oops  :)

PR #467, opened 2022-10-07
[release-4.10] OCPBUGS-647: Fix DNS endpoint hack to prefer local instead of forcing it

martinkennelly:


melvinjoseph86:
as per the comments added in OCPBUGS-647, i am marking this as verified via pre-merge workflow.
/label qe-approved

PR #469, opened 2022-10-11
[release-4.8] OCPBUGS-2205: Fix DNS endpoint hack to prefer local instead of forcing it

martinkennelly:


melvinjoseph86:
as per the comments added in OCPBUGS-2205, i am marking this as verified via pre-merge workflow.

martinkennelly:
@danwinship Can you take a look and add the required labels?

PR #471, opened 2022-11-03
Bug 2050230: pass ResourceVersion:0 for kube List() calls

jcaamano:
Set ResourceVersion to 0 so that list requests are served from the api server cache instead of etcd.

Keep pagination, as requests may still fallback to etcd if the apiserver cache has not been initialized yet (i.e. apiserver restart).

Ref: https://github.com/kubernetes/enhancements/blob/328e235effab8ba0869586bdb4ee7b10a0ee0801/keps/sig-api-machinery/3157-watch-list/README.md#appendix


jcaamano:
@dcbw PTAL

PR #481, opened 2022-12-12
Update for 4.13 / go 1.19, including gofmt updates

danwinship:
Autogenerated ART PR (#478) is failing because of gofmt


martinkennelly:
Assuming kube-proxy update will come later as per #486

PR #488, opened 2022-12-17
OCPBUGS-998: [release-4.11] NetworkPolicy: Remove stale flows that may exist due to an upgrade

flavio-fernandes:
When upgrading from OCP 4.10, temporary OpenFlow rules became permanent due to a change in the cookies used for them. Because of that, traffic for pods that should have been blocked only while network policies were being applied was never unblocked.

Reported-at: https://issues.redhat.com/browse/OCPBUGS-998
Signed-off-by: Flavio Fernandes <flaviof@redhat.com>

### DIFF
--- a/pkg/network/node/networkpolicy.go
+++ b/pkg/network/node/networkpolicy.go
@@ -166,8 +166,10 @@ func (np *networkPolicyPlugin) Start(node *OsdnNode) error {
 			otx.AddFlow("table=30, priority=100, ip, nw_src=%s, nw_dst=%s, ct_state=-rpl, actions=ct(commit,table=31)", scn.ClusterCIDR.String(), dcn.ClusterCIDR.String())
 		}
 	}
+	np.purgeLegacyFlows(otx)
 
 	if err := otx.Commit(); err != nil {
+		klog.Errorf("Error performing initial OVS flows: %v", err)
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/network/node/networkpolicy.go
```diff
@@ -333,6 +335,17 @@ func (np *networkPolicyPlugin) syncNamespaceImmediately(npns *npNamespace) {
...
 
+// purgeLegacyFlows removes flows that may have been created by previous versions and
+// are no longer applicable. Called once upon networkPolicyPlugin start.
+func (np *networkPolicyPlugin) purgeLegacyFlows(otx ovs.Transaction) {
+	const legacyNetPolIsolationCookie = 1
+
+	// Remove temporary flows added by SetUpPod using cookie from previous release.
+	// Ref: https://issues.redhat.com/browse/OCPBUGS-998
+	otx.DeleteFlows("table=27, cookie=%d/-1, ip", legacyNetPolIsolationCookie)
+	otx.DeleteFlows("table=80, cookie=%d/-1, ip", legacyNetPolIsolationCookie)
```
npinaeva, line 346: That is a nice fix!
Another thought: when these flows are present, it means netpol flows may not yet be created for some pod, so if we just remove them, it will be unprotected before netpols are re-evaluated for that pod. Not sure how bad it is, @danwinship wdyt?

flavio-fernandes: hm... not sure if I follow @npinaeva .  The rules with the new cookie would still be present in that case, even if these ones are removed in that case, no? Being so, I'd think that should be enough to keep the pods protected while netpol is being re-evaluated.

npinaeva: New cookie is set by `SetUpPod`, which I guess is only called on CNI request, so just on restart (upgrade), if `networkPolicyPlugin.Start` is called before the new flows are installed for that pod, we may have no netpol flows and no "drop all" flow at the same time. But I am not sure about what happens on restart and in which order.

danwinship: I don't think it really matters; this is a super-obscure bug that can only happen on one possible upgrade per cluster

flavio-fernandes:
@danwinship @abhat could you add `/label backport-risk-assessed` here, please?

PR #221, opened 2020-11-13
Make binaries not stripped

juanluisvaladas:
By explicitly defining ldflags we remove the -s flag making the binary
produced by the compiler not stripped.

PR #489, opened 2023-01-06
do not strip binary files

JacobTanenbaum:
by removing the "-w" flag in the make file we leave in the DWARF symbol table during compilation leaves in the debug information. This will allow us to get data from coredumps

discussion from the last time that we removed flags that stripped the symbol tables  https://github.com/openshift/sdn/pull/221

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ export GO111MODULE=on
 unexport GOPATH
 
 GO_LD_FLAGS = \
-    -ldflags "-w $(call version-ldflags,$(GO_PACKAGE)/pkg/version) $(GO_LD_EXTRAFLAGS)"
+    -ldflags "$(call version-ldflags,$(GO_PACKAGE)/pkg/version) $(GO_LD_EXTRAFLAGS)"
 
 GO_BUILD_PACKAGES = \
     ./cmd/... \


JacobTanenbaum:
without debug 
15329K  openshift-sdn-cni
44540K  openshift-sdn-controller
54467K  openshift-sdn-node
2084K  host-local
45670K  kube-proxy

160006K total 

with debug 
19731K  openshift-sdn-cni
56150K  openshift-sdn-controller
68691K  openshift-sdn-node
2953K  host-local
57761K  kube-proxy

205286K total  

45280K difference 

28.3 % larger binaries with the debug data

palonsoro:
Thanks for opening this. I was about to open a similar thing, because I tried to get a coredump from 4.9.43 and I was unable.

While having a look at this topic, I have found this commit on build-machinery-go: https://github.com/openshift/build-machinery-go/commit/367a4bdc4bfe5b407b365a693134da66d541f7b3

If I am understanding this correctly, as long as the SDN vendors a build-machinery-go version that includes this commit, binaries should be no longer stripped at all by default.

TL;DR worth checking if just removing the custom `GO_LD_FLAGS` + vendoring latest build-machinery-go does the trick.

JacobTanenbaum:
no longer required

PR #490, opened 2023-01-11
Handle race condition to setup default vnid flows

pperiyasamy:
There is a race condition where EnsureVNIDRules (via setup in pod.go) is invoked followed by AddNetNamespace which causes default vnid ingress and egress flow rules in table 80 and 27 not getting programmed. This fix ensures these flow rules are programmed though AddNetNamespace is processed at later point of time.

### DIFF
--- a/pkg/network/node/networkpolicy.go
+++ b/pkg/network/node/networkpolicy.go
@@ -56,6 +56,13 @@ type networkPolicyPlugin struct {
 	// since pod.Status.PodIP is not set during pod creation, we keep a separate cache of ips already known to sdn,
 	// but not yet known to the apiserver. When pod.Status.PodIP is set, pod can be deleted from this map
 	localPodIPs map[ktypes.UID]string
+	// unProcessedVNIDs holds the vnid entries for which EnsureVNIDRules() is already invoked upon
+	// pod setup even before AddNetNamespace() is invoked. This is possible when WaitAndGetVNID()
+	// could return the vnid in the pod setup thread and then the setup thread continues, before
+	// the AddNetNamespace() call runs in the handler thread. In such scenario populate the vnid
+	// into unProcessedVNIDs in EnsureVNIDRules() method and it can be looked up later in
+	// AddNetNamespace() method so that required nw policy flow rules programmed into OVS.
+	unProcessedVNIDs sets.Int32
 }
 
 // npNamespace tracks NetworkPolicy-related data for a Namespace
@@ -118,6 +125,8 @@ func NewNetworkPolicyPlugin() osdnPolicy {
 		warnedPolicies:  make(map[ktypes.UID]string),
 		skippedPolicies: make(map[ktypes.UID]string),
 		localPodIPs:     make(map[ktypes.UID]string),
+
+		unProcessedVNIDs: sets.NewInt32(),
 	}
 }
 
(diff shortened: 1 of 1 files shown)


pkg/network/node/networkpolicy.go
```diff
@@ -118,6 +122,8 @@ func NewNetworkPolicyPlugin() osdnPolicy {
 		warnedPolicies:  make(map[ktypes.UID]string),
 		skippedPolicies: make(map[ktypes.UID]string),
 		localPodIPs:     make(map[ktypes.UID]string),
+
+		nsToBeCreated: make(map[uint32]interface{}),
```
danwinship, line 126: you should use `sets.Int32`. (Yeah, we're working with `uint32`s, but we know they're only 24 bits max, so they can be safely cast to `int32`.)

also, `nsToBeCreated` isn't a very descriptive name... it needs some combination of a better name, and a good comment explaining how/when/why it's used

pkg/network/node/networkpolicy.go
```diff
@@ -1078,7 +1088,13 @@ func (np *networkPolicyPlugin) refreshNamespaceNetworkPolicies() {
...
 			}
 		}
-		if npns.mustRecalculate && npns.inUse {
+		var npnsNotCreated bool
+		if _, ok := np.nsToBeCreated[npns.vnid]; ok {
+			npnsNotCreated = true
+			npns.inUse = true
+			delete(np.nsToBeCreated, npns.vnid)
+		}
+		if npnsNotCreated || (npns.mustRecalculate && npns.inUse) {
```
danwinship, line 1097: This should all be in `AddNetNamespace`.

Also, need to think about how to make sure `nsToBeCreated` doesnt' leak elements. You should remove elements from it in `DeleteNetNamespace`, and possibly `SyncVNIDRules`

pkg/network/node/networkpolicy.go
pperiyasamy: oh! yes, vnid is 24 bits max as per vxlan header field and using `sets.Int32` now. changed field name into `unProcessedVNIDs` with some more description. is that fine now ?

pkg/network/node/networkpolicy.go
pperiyasamy: correct, added removal code in above mentioned methods to ensure no leaks with `unProcessedVNIDs`.

pkg/network/node/networkpolicy.go
```diff
@@ -279,6 +298,10 @@ func (np *networkPolicyPlugin) DeleteNetNamespace(netns *osdnv1.NetNamespace) {
 	delete(np.namespaces, netns.NetID)
 	npns.gotNetNamespace = false
 
+	if np.unProcessedVNIDs.Has(int32(npns.vnid)) {
+		np.unProcessedVNIDs.Delete(int32(npns.vnid))
+	}
```
danwinship, line 303:
```suggestion
	np.unProcessedVNIDs.Delete(int32(npns.vnid))
```

pkg/network/node/networkpolicy.go
```diff
@@ -552,6 +579,9 @@ func (np *networkPolicyPlugin) SyncVNIDRules() {
 			npns.inUse = false
 			np.syncNamespace(npns)
 		}
+		if np.unProcessedVNIDs.Has(int32(vnid)) {
+			np.unProcessedVNIDs.Delete(int32(vnid))
+		}
```
danwinship, line 584:
```suggestion
		np.unProcessedVNIDs.Delete(int32(vnid))
```

danwinship (commented):
/approve
just some trivial cleanups

PR #494, opened 2023-01-18
[release-4.10] OCPBUGS-5977: Use pagination and set ResourceVersion to 0 when listing resources

jcaamano:
Using the resource List API without pagination overloads etcd
for a high number of resources. Use pagination with the default page
item count, currently 500.

Set ResourceVersion to 0 so that list requests are served from the api
server cache instead of etcd. Requests may fallback to etcd if the apiserver
cache has not been initialized yet (i.e. apiserver restart).

Ref: https://github.com/kubernetes/enhancements/blob/328e235effab8ba0869586bdb4ee7b10a0ee0801/keps/sig-api-machinery/3157-watch-list/README.md#appendix

### DIFF
--- a/pkg/network/common/common.go
+++ b/pkg/network/common/common.go
@@ -145,7 +145,7 @@ func (pcn *ParsedClusterNetwork) CheckHostNetworks(hostIPNets []*net.IPNet) erro
 	return kerrors.NewAggregate(errList)
 }
 
-func (pcn *ParsedClusterNetwork) CheckClusterObjects(subnets []osdnv1.HostSubnet, pods []corev1.Pod, services []corev1.Service) error {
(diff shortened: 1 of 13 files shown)


jcaamano:
/retitle [release-4.10] OCPBUGS-5977: Use pagination and set ResourceVersion to 0 when listing resources

danwinship:
Why are we backporting the pagination fix now but we didn't before?
Are you sure you got all relevant fixes to it? I feel like there were more fixups to it than this...

jcaamano:
@danwinship

All the fixes are associated with the same single bug: https://bugzilla.redhat.com/show_bug.cgi?id=2050230
The first two fixes were done during the 4.11 dev cycle.
The last one was done the 4.12 dev cycle.
And once the bug was verified, we backported the last fix to 4.11. And now all the fixes to 4.10.

PR #496, opened 2023-01-18
iptables: use container iptables, not host's

dcbw:
Using the host's iptables was necessary when the container image might have been running on RHEL7 BYOH, but given that's no longer possible, we can clean this up and just use the container's iptables (since both it and the host will be iptables-nft).


danwinship:
yay!
/lgtm

PR #498, opened 2023-01-25
OCPBUGS-4133: rebase kube to kube 1.25.4

JacobTanenbaum:
Windows needs commit: https://github.com/openshift/kubernetes/commit/e93a90ea9a3c13cf1b1d9f8f94bca18374305bfc in order
to function.

bump kube to the same version as kubelet


aravindhp:
Is there a reason not to be rebase to 1.24.4 given that is where kubelet is at?

aravindhp:
For my understanding, why is this PR needed for Windows? All WMCO cares about is if the fix is in https://github.com/openshift/kubernetes/commits/sdn-4.12-kubernetes-1.25.4. We will move to using that branch in our [sub-modules](https://github.com/openshift/windows-machine-config-operator/blob/c9c49bea1366fbc20d947740ef2f583cac2fdaf4/.gitmodules#L22) to pick up the fix. WMCO does not depend on openshift/sdn in any way. Is this to keep the repo in sync?

dcbw:
/lgtm

Fairly small changeset, though @danwinship should OK the node local detector stuff

PR #509, opened 2023-02-20
Revert "Revert "SDN-3635: OSDN kube 1.26.0 rebase""

msherif1234:
Reverts openshift/sdn#508


danwinship:
/hold
don't just revert the revert. Push it as multiple commits again.

It might also be helpful to push the "drop all remaining references to userspace proxy" part as a separate PR before the rebase, to get that merged and then have fewer things to debug in this one?

msherif1234:
> /hold don't just revert the revert. Push it as multiple commits again.
> 
> It might also be helpful to push the "drop all remaining references to userspace proxy" part as a separate PR before the rebase, to get that merged and then have fewer things to debug in this one?

I want to findout about the CI breakage and so far no sign so this PR was incorrectly reverted , I will break it down as it was orginally and no need to have multiple PR we already burnt enough time here, will just wait for the current run to complete then will fix it

danwinship:
ah, well, I filed #511; if that passes the payload test and this still fails then we could merge that to simplify this PR.

msherif1234:
not need #510 will handle the rebase

PR #510, opened 2023-02-21
SDN-3635: OSDN kube 1.26.0 rebase 2nd attempt

msherif1234:


pkg/cmd/openshift-sdn-node/kube_proxy.go
```diff
@@ -121,6 +118,8 @@ func newProxyServer(config *kubeproxyconfig.KubeProxyConfiguration, client clien
 	klog.V(0).Infof("kube-proxy running in single-stack %s mode", iptInterface.Protocol())
 
 	if proxyMode == proxyModeIPTables {
+		// userspace proxy mode was removed upstream in 1.26,
+		// let's fallback to iptables
```
danwinship, line 122: stray comment

danwinship:
commit message is still wrong too:

> Also, k8s.io/kubernetes/pkg/proxy/userspace was removed from 1.26 code base, so let's keep  "--proxy-mode userspace" for backwards compatibility, but let's make it use iptables instead.

pkg/cmd/openshift-sdn-node/kube_proxy.go
msherif1234: removed

msherif1234:
azure job is passing and no sign of endpoint issue for openshit-console

PR #504, opened 2023-02-11
SDN-3635: OSDN kube 1.26.0 rebase

msherif1234:

PR #511, opened 2023-02-21
Drop support for --proxy-mode=userspace

danwinship:
Split out from #504 / #510, so there's less to test in that commit. Also includes #507 to remove some now-useless code.

/assign @msherif1234 @ricky-rav

### DIFF
--- a/pkg/cmd/openshift-sdn-node/kube_proxy.go
+++ b/pkg/cmd/openshift-sdn-node/kube_proxy.go
@@ -5,7 +5,6 @@ import (
 	"k8s.io/apimachinery/pkg/fields"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	"k8s.io/kubernetes/pkg/features"
-	"net"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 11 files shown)


msherif1234:
lets see what the CI outcomes will look I fired 2 more runs so we can get a quorum

msherif1234:
/hold
need to check manual ci runs results

danwinship:
> lets see what the CI outcomes will look I fired 2 more runs so we can get a quorum

I think those jobs do multiple runs internally... that's why they take so long

pkg/cmd/openshift-sdn-node/kube_proxy.go
```diff
@@ -187,7 +163,7 @@ func newProxyServer(config *kubeproxyconfig.KubeProxyConfiguration, client clien
 		EnableProfiling:    config.EnableProfiling,
 		ConfigSyncPeriod:   config.ConfigSyncPeriod.Duration,
 		HealthzServer:      healthzServer,
-		UseEndpointSlices:  useEndpointSlices,
+		UseEndpointSlices:  true,
```
msherif1234, line 166: we don't need this anymore as well as the check

danwinship: yeah... I was consider that to be part of the "sync to 1.26" commit (which isn't in this PR) rather than the "stop supporting userspace" commit...

danwinship:
obsoleted by #510

PR #517, opened 2023-03-15
[release-4.9] OCPBUGS-10363: Add node egress IP assignment resync

kyrtapz:
Manual back-port of https://github.com/openshift/sdn/pull/501
There were conflicts due to cloud egress IP bits that were added in 4.10.

### DIFF
--- a/pkg/network/node/egressip.go
+++ b/pkg/network/node/egressip.go
@@ -168,11 +168,27 @@ func (eip *egressIPWatcher) runIPAssignmentResync(stopCh <-chan struct{}) {
 		return err
 	}
 
+	// ~1 sec total
+	syncIPsBackOff := utilwait.Backoff{
+		Duration: 100 * time.Millisecond,
+		Factor:   1.25,
+		Steps:    7,
+	}
 	syncIPs := func() {
 		eip.tracker.Lock()
 		defer eip.tracker.Unlock()
 
-		eip.SyncIPAssignments()
+		var lastSyncErr error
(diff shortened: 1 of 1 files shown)


pkg/network/node/egressip.go
```diff
@@ -79,34 +81,49 @@ func (eip *egressIPWatcher) Start(osdnInformers osdninformers.SharedInformerFact
...
-			klog.Infof("Cleaning up stale egress IP %s", addr.IP.String())
-			err = netlink.AddrDel(link, &addr)
-			if err != nil {
-				utilruntime.HandleError(fmt.Errorf("Could not clean up stale egress IP: %v", err))
+		if addr.Label == label {
+			if eip.iptablesMark[ip] == "" {
+				klog.Infof("Cleaning up stale egress IP %s", addr.IP.String())
+				err = netlink.AddrDel(link, &addr)
+				if err != nil {
+					utilruntime.HandleError(fmt.Errorf("Could not clean up stale egress IP: %v", err))
```
danwinship, line 106: we don't like `utilruntime.HandleError` any more anyway (#356), so I'd say just make this use `klog.Error` like the 4.10 version, so there will be fewer conflicts if we need to backport more patches later.

(I think this ended up being the only big difference from the 4.10 version, but if I missed anything else, the same argument may apply there.)

danwinship:
/approve
/label backport-risk-assessed

if possible I'd like QE to test this a bit and give the lgtm since it wasn't a totally clean backport

kyrtapz:
4.9 is out of support, closing.

PR #525, opened 2023-04-12
OCPBUGS-1370: save and delete the old egress network policy

JacobTanenbaum:
When egress network policies are updated they are deleted and recreated. When we are processing the update for egress network policies currently we delete the new version of the network policy as opposed to the old version. If a dnsName is being removed from an egress network policy during an update the dnsName is not in the most recent version of the object and it will not be removed from the egress network policies egressDNS maps.

Plug the leaked data by ensuring that we delete the previous version of the egressNetwork policy and correctly remove the DNS names from the egressDNS object.

### DIFF
--- a/pkg/network/node/egress_network_policy.go
+++ b/pkg/network/node/egress_network_policy.go
@@ -75,10 +75,10 @@ func (plugin *OsdnNode) handleEgressNetworkPolicy(policy *osdnv1.EgressNetworkPo
 	for i, oldPolicy := range policies {
 		if oldPolicy.UID == policy.UID {
 			policies = append(policies[:i], policies[i+1:]...)
+			plugin.egressDNS.Delete(oldPolicy)
 			break
 		}
 	}
-	plugin.egressDNS.Delete(*policy)
 
 	if eventType != watch.Deleted && len(policy.Spec.Egress) > 0 {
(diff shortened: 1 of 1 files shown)


pkg/network/node/egress_network_policy.go
```diff
@@ -71,14 +71,16 @@ func (plugin *OsdnNode) handleEgressNetworkPolicy(policy *osdnv1.EgressNetworkPo
 	plugin.egressPoliciesLock.Lock()
 	defer plugin.egressPoliciesLock.Unlock()
 
+	var oldPolicySave osdnv1.EgressNetworkPolicy
 	policies := plugin.egressPolicies[vnid]
 	for i, oldPolicy := range policies {
 		if oldPolicy.UID == policy.UID {
 			policies = append(policies[:i], policies[i+1:]...)
+			oldPolicySave = oldPolicy
```
dcbw, line 79: @JacobTanenbaum I was thinking, why don't we just put the `plugin.egressDNS.Delete(oldPolicy)` here?

If we can't find the old policy in the list, then oldPolicySave will be empty, and the Delete() will just do nothing anyway.

JacobTanenbaum: done

PR #532, opened 2023-04-26
OCPBUGS-12435: EgressNetworkPolicy DNS resolution does not fall back to TCP

JacobTanenbaum:
DNS resolution has a fixed buffer size of 512 and when getting a DNS response that is to large for the buffer the response is truncated and in general systems fall back to a TCP dns request. Currently we do not fall back to that TCP dns request.

Add the fall back TCP DNS request when the original UDP request returns a truncated response. In addition I do not use the ErrTruncated to test for a truncated message because eventually the library gets rid of that error in favor of checking the Truncated flag in the message.


JacobTanenbaum:
@dcbw @danwinship PTAL when you get a chance

PR #534, opened 2023-05-02
OCPBUGS-13717: Use the ovsver build arg to infer the openvswitch short version number

aleskandro:
This PR proposes to add a build argument to drive the proper openvswitch package to install from the RPM repositories.

This will ease the versions bump and will be helpful for some OKD-related builds when the RPM repositories are not in sync, and some packages version might miss.

cc @vrutkovs @LorbusChris


trozet (approved):
/approve

PR #545, opened 2023-05-22
[release-4.10] OCPBUGS-13919: save and delete the old egress network policy

JacobTanenbaum:
When egress network policies are updated they are deleted and recreated. When we are processing the update for egress network policies currently we delete the new version of the network policy as opposed to the old version. If a dnsName is being removed from an egress network policy during an update the dnsName is not in the most recent version of the object and it will not be removed from the egress network policies egressDNS maps.

Plug the leaked data by ensuring that we delete the previous version of the egressNetwork policy and correctly remove the DNS names from the egressDNS object.

manual cherry-pick of https://github.com/openshift/sdn/pull/544 to move things along faster


dcbw:
AWS flake appears to be:

`: [sig-instrumentation] Prometheus when installed on the cluster should report telemetry if a cloud.openshift.com token is present`

which doesn't appear to be SDN-caused.

/retest-required

PR #549, opened 2023-05-25
OCPBUGS-3176: Enable IP Forwarding if disabled

trozet:
With changes to disable IP forwarding by default with OVNK: https://github.com/openshift/machine-config-operator/pull/3676

OVN to SDN migration is now broken, because pods roll out first that require IP forwarding. IP forwarding for SDN is enabled via MCO node rollout, which happens later.

This change enables IP forwarding in the SDN pod itself if needed.


pliurh:
I tested this PR. It works in SDN migration.

PR #551, opened 2023-05-30
[release-4.12] OCPBUGS-14304: fix possible concurrent map read/write

JacobTanenbaum:
Manual cherry-pick of https://github.com/openshift/sdn/pull/546

there is a potential concurrent map read/write in handleAddOrUpdatePod() in pkg/network/node/networkpolicy.go

handleAddOrUpdatePod() has to get the networkPolicyPlugin lock before calling isOnPodNetwork() because that can call getPodIP() which reads the localPodIPs map. If the lock is not grabbed there could be a concurrent read/write of the map

this patch grabs the lock before calling isOnPodNetwork()


mffiedler:
@huiran0826 @anuragthehatter One more to test in this backport chain.   4.14, 4.13 and 4.11 are already verified.

PR #552, opened 2023-05-30
[release-4.11] OCPBUGS-14305: fix possible concurrent map read/write

JacobTanenbaum:
manual cherry-pick of https://github.com/openshift/sdn/pull/546

there is a potential concurrent map read/write in handleAddOrUpdatePod() in pkg/network/node/networkpolicy.go

handleAddOrUpdatePod() has to get the networkPolicyPlugin lock before calling isOnPodNetwork() because that can call getPodIP() which reads the localPodIPs map. If the lock is not grabbed there could be a concurrent read/write of the map

this patch grabs the lock before calling isOnPodNetwork()


anuragthehatter:
This looks good on 4.11z bot image using same verification steps as in OCPBUGS-13975

/label qe-approved

PR #557, opened 2023-07-10
OCPBUGS-15977: Dockerfile changes to build both rhel8 and rhel9 binaries

abhat:
Since the shim (openshift-sdn) gets copied to the host OS and executed in the host mount namespace by CRIO/Multus it needs to be runtime compatible with the host OS. Running a RHEL9-built shim on a RHEL8 system doesn't work due to different shared library dependencies between the two OS versions.

This wasn't a problem before because CGO_ENABLED=0 which essentially statically linked everything into the binary. But since we actually need CGO_ENABLED=1 (which ART forces on "official" builds anyway) to ensure we use OpenSSL's crypto for FIPS compliance, we run into the OS version problem with our binaries since they are really always built with CGO_ENABLED=1 anyway.

So... build two separate versions of openshift-sdn and osdn-host-local (which is invoked by openshift-sdn shim) in different layers, and copy the shims into a special location where our container startup scripts can find it.


images/sdn/Dockerfile.rhel
```diff
@@ -1,19 +1,33 @@
...
 COPY . .
 RUN make build --warn-undefined-variables
-RUN CGO_ENABLED=0 GO_BUILD_FLAGS="-tags no_openssl" make build GO_BUILD_PACKAGES="github.com/openshift/sdn/cmd/openshift-sdn-cni" --warn-undefined-variables
+RUN CGO_ENABLED=1 make build GO_BUILD_PACKAGES="github.com/openshift/sdn/cmd/openshift-sdn-cni" --warn-undefined-variables
 
 FROM registry.ci.openshift.org/ocp/4.13:cli AS cli
 FROM registry.ci.openshift.org/ocp/4.13:base
 
+mkdir -p /opt/cni/bin/rhel9
 COPY --from=builder /go/src/github.com/openshift/sdn/openshift-sdn-node /usr/bin/
```
mrunalp, line 11: The base image is rhel 8, so should copy the rhel 8 binaries to /usr/bin.

images/sdn/Dockerfile.rhel
```diff
@@ -1,19 +1,33 @@
-FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.19-openshift-4.13 AS builder
+FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.19-openshift-4.13 AS builder
```
mrunalp, line 1: nit: Might be better to explicitly call it rhel9-builder

images/sdn/Dockerfile.rhel
```diff
@@ -1,19 +1,33 @@
...
 
 COPY --from=cli /usr/bin/oc /usr/bin/
 
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.19-openshift-4.13 AS rhel8-builder
+WORKDIR /go/src/github.com/openshift/sdn
+COPY . .
+RUN make build --warn-undefined-variables
+RUN CGO_ENABLED=1 make build GO_BUILD_PACKAGES="github.com/openshift/sdn/cmd/openshift-sdn-cni" --warn-undefined-variables
+
+FROM registry.ci.openshift.org/ocp/4.13:base
```
mrunalp, line 22: I don't think you need this line.

images/sdn/Dockerfile.rhel
```diff
@@ -1,19 +1,33 @@
...
-COPY --from=builder /go/src/github.com/openshift/sdn/openshift-sdn-controller /usr/bin/
-COPY --from=builder /go/src/github.com/openshift/sdn/openshift-sdn-cni /opt/cni/bin/openshift-sdn
-COPY --from=builder /go/src/github.com/openshift/sdn/host-local /usr/bin/cni/osdn-host-local
+mkdir -p /opt/cni/bin/rhel9
+COPY --from=rhel9-builder /go/src/github.com/openshift/sdn/openshift-sdn-cni /opt/cni/bin/rhel9/openshift-sdn
+COPY --from=rhel9-builder /go/src/github.com/openshift/sdn/host-local /opt/cni/bin/rhel9/osdn-host-local
 
 COPY --from=cli /usr/bin/oc /usr/bin/
 
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.19-openshift-4.13 AS rhel8-builder
```
mrunalp, line 16: YOu can move this whole rhel 8 build code steps to after the rhel 9 steps above.

images/sdn/Dockerfile.rhel
```diff
@@ -1,16 +1,27 @@
...
 FROM registry.ci.openshift.org/ocp/4.13:cli AS cli
 FROM registry.ci.openshift.org/ocp/4.13:base
 
-COPY --from=builder /go/src/github.com/openshift/sdn/openshift-sdn-node /usr/bin/
-COPY --from=builder /go/src/github.com/openshift/sdn/openshift-sdn-controller /usr/bin/
-COPY --from=builder /go/src/github.com/openshift/sdn/openshift-sdn-cni /opt/cni/bin/openshift-sdn
-COPY --from=builder /go/src/github.com/openshift/sdn/host-local /usr/bin/cni/osdn-host-local
+mkdir -p /opt/cni/bin/rhel9
+COPY --from=rhel9-builder /go/src/github.com/openshift/sdn/openshift-sdn-cni /opt/cni/bin/rhel9/openshift-sdn
+COPY --from=rhel9-builder /go/src/github.com/openshift/sdn/host-local /opt/cni/bin/rhel9/osdn-host-local
```
dcbw, line 18: @abhat turns out we don't need osdn-host-local to be built for RHEL9 since it's just run from inside the container.

But for RHEL8 we do want to keep it in /usr/bin/cni since that's where it gets execed from.

dcbw:
Follow-up fix: https://github.com/openshift/sdn/pull/558

PR #558, opened 2023-07-12
OCPBUGS-15977: Dockerfile: keep a RHEL-8 CNI shim binary in the default /opt/cni/bin dir

dcbw:
Ensure that we keep a binary at the same place as the older images just in case something expects it there. Really only affects CI due to the way we run our CI upgrade jobs by hacking around with the CNO manifests.


zhaozhanqi:
@mrunalp Could you help take a look this PR to add the label jira/valid-bug

PR #560, opened 2023-07-18
[release-4.9] OCPBUGS-16306: Dockerfile: build both RHEL7 and RHEL8 shims

jcaamano:
Since the shim (openshift-sdn) gets copied to the host OS and
    executed in the host mount namespace by CRIO/Multus it needs
    to be runtime compatible with the host OS. Running a RHEL8-built
    shim on a RHEL7 system doesn't work due to different shared library
    dependencies between the two OS versions.
    
    This wasn't a problem before because CGO_ENABLED=0 which essentially
    statically linked everything into the binary. But since we actually
    need CGO_ENABLED=1 (which ART forces on "official" builds anyway)
    to ensure we use OpenSSL's crypto for FIPS compliance, we run into
    the OS version problem with our binaries since they are really
    always built with CGO_ENABLED=1 anyway.
    
    So... build two separate versions of openshift-sdn and
    osdn-host-local (which is invoked by openshift-sdn shim) in
    different layers, and copy the shims into a special location where
    our container startup scripts can find it.


images/sdn/Dockerfile.rhel
```diff
@@ -1,14 +1,25 @@
-FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.16-openshift-4.9 AS builder
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.16-openshift-4.9 AS rhel8-builder
 WORKDIR /go/src/github.com/openshift/sdn
 COPY . .
 RUN make build --warn-undefined-variables
-RUN CGO_ENABLED=0 GO_BUILD_FLAGS="-tags no_openssl" make build GO_BUILD_PACKAGES="github.com/openshift/sdn/cmd/openshift-sdn-cni" --warn-undefined-variables
+RUN CGO_ENABLED=1 GO_BUILD_FLAGS="-tags no_openssl" make build GO_BUILD_PACKAGES="github.com/openshift/sdn/cmd/openshift-sdn-cni" --warn-undefined-variables
```
dcbw, line 5: @jcaamano lets remove the tags no_openssl while we're here too; ART also kills that at real build time.

jupierce:
@jcaamano - please ensure that a PR is opened against ocp-build-data when the new parent image change merges. e.g. https://github.com/openshift-eng/ocp-build-data/pull/3214

dcbw:
/override ci/prow/e2e-gcp

Faliure is unrelated to this PR

dcbw:
/override ci/prow/e2e-gcp

Faliure is unrelated to this PR

PR #561, opened 2023-07-24
[release-4.8] OCPBUGS-16253: Dockerfile changes to build both rhel7 and rhel8 binaries

jcaamano:
Since the shim (openshift-sdn) gets copied to the host OS and executed in the host mount namespace by CRIO/Multus it needs to be runtime compatible with the host OS. Running a RHEL8-built shim on a RHEL7 system doesn't work due to different shared library dependencies between the two OS versions.

This wasn't a problem before because CGO_ENABLED=0 which essentially statically linked everything into the binary. But since we actually need CGO_ENABLED=1 (which ART forces on "official" builds anyway) to ensure we use OpenSSL's crypto for FIPS compliance, we run into the OS version problem with our binaries since they are really always built with CGO_ENABLED=1 anyway.

So... build two separate versions of openshift-sdn and osdn-host-local (which is invoked by openshift-sdn shim) in different layers, and copy the shims into a special location where our container startup scripts can find it.

(cherry picked from commit 8fe1a62f940b2f8b6a5ac9f1f4ab6a57572ff75a)

### DIFF
--- a/images/sdn/Dockerfile.rhel
+++ b/images/sdn/Dockerfile.rhel
@@ -1,14 +1,25 @@
-FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.16-openshift-4.8 AS builder
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.16-openshift-4.8 AS rhel8-builder
 WORKDIR /go/src/github.com/openshift/sdn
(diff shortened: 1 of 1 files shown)


dcbw:
/override ci/prow/e2e-aws

https://github.com/openshift/origin/pull/27422

PR #569, opened 2023-07-31
Live migration: Add flows that allow openshift-sdn to connect to ovnkube

pliurh:
During live migration, openshift-sdn nodes will be connected to
ovnkube network as hybrid overlay nodes. The VXLAN packets from
ovnkube node have the same dest_MAC as the both the inner and
outer MAC with the MAC address of the physical interface. In order
to forward the packet to the pod, we swap the inner MAC to the
real MAC of the pod in table 10.

### DIFF
--- a/pkg/network/node/ovscontroller.go
+++ b/pkg/network/node/ovscontroller.go
@@ -80,9 +80,11 @@ import (
 //   acceptable incoming tun0 traffic (in_port=2) goes to table 30 (general routing)
 //   traffic from any other port is assumed to be from a container and goes to table 20
 
-// Table 10: VXLAN ingress filtering
+// Table 10: Inbound VXLAN packet handling
 //   per-remote-node rules are filled in by AddHostSubnetRules()
 //   any VXLAN traffic from a non-node IP is dropped
+//   when doing SDN live migration, the node is connected to OVN-K as a hybrid-overlay node,
+//   swap overlay dst MAC for ingress traffic from OVN-K nodes
 
 // Table 20: from OpenShift container
 //   mostly filled in by setupPodFlows
@@ -172,6 +174,7 @@ type ovsController struct {
 	pluginId int
 	localIP  string
 	tunMAC   string
+	localMAC string
 }
 
 const (
@@ -185,8 +188,8 @@ const (
 	ruleVersionTable = 253
 )
 
-func NewOVSController(ovsif ovs.Interface, pluginId int, localIP string) *ovsController {
-	return &ovsController{ovs: ovsif, pluginId: pluginId, localIP: localIP}
+func NewOVSController(ovsif ovs.Interface, pluginId int, localIP string, localMAC string) *ovsController {
+	return &ovsController{ovs: ovsif, pluginId: pluginId, localIP: localIP, localMAC: localMAC}
 }
 
 func (oc *ovsController) getVersionNote() string {
@@ -238,6 +241,13 @@ func (oc *ovsController) SetupOVS(clusterNetworkCIDR []string, serviceNetworkCID
 		return err
 	}
 
+	if oc.localMAC == "" {
+		return fmt.Errorf("failed to find the MAC address for local IP %s", oc.localIP)
+	}
+	if err := oc.ensureTunMAC(); err != nil {
+		return err
+	}
+
 	otx := oc.ovs.NewTransaction()
 
 	// Table 0: initial dispatch based on in_port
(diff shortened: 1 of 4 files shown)


pkg/network/node/ovscontroller.go
```diff
@@ -923,3 +926,17 @@ func (oc *ovsController) SetNamespaceEgressViaEgressIPs(vnid uint32, egressIPsMe
...
+	for _, interf := range interfaces {
+		if addrs, err := interf.Addrs(); err == nil {
+			for _, addr := range addrs {
+				if strings.Contains(addr.String(), localIP) {
+					return interf.HardwareAddr.String()
+				}
+			}
+		}
+	}
+	return ""
```
kyrtapz, line 941: Should this be an error instead? Or do you expect a scenario in which `localMAC` is empty?

pliurh: I add the check at line 272. PTAL

kyrtapz:
@pliurh  UTs are failing 
```
pkg/network/node/ovscontroller_test.go:27:5: oc.localMAC undefined (type *ovsController has no field or method localMAC)
```
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_sdn/569/pull-ci-openshift-sdn-master-unit/1686992266020786176

pkg/network/node/ovscontroller.go
```diff
@@ -267,6 +268,17 @@ func (oc *ovsController) SetupOVS(clusterNetworkCIDR []string, serviceNetworkCID
...
 	// eg, "table=10, priority=100, tun_src=${remote_node_ip}, actions=goto_table:30"
+	// Swap overlay dst MAC for ingress OVN-K hybrid overlay traffic
+	if oc.localMAC == "" {
+		return fmt.Errorf("failed to find the MAC address for local IP %s", oc.localIP)
+	}
+	if err := oc.ensureTunMAC(); err != nil {
+		return err
+	}
+	// set the dst MAC for traffic from OVN-K nodes to local node
+	otx.AddFlow("table=10, priority=210, ip, nw_dst=%s, dl_dst=%s, actions=set_field:%s->eth_dst, resubmit:10", localSubnetGateway, oc.localMAC, oc.tunMAC)
```
danwinship, line 279: I believe if a host-level process or hostnetwork pod on one node sends to a pod IP on another node, that traffic will end up with a source IP of the `localSubnetGateway` on the source node. Adding this rule would break that traffic.

If there isn't anything else you can use to recognize this traffic, then you might need to use conntrack, and set `ct_mark` or `ct_label` in the following rule, so that this rule can then recognize the packet as being a reply to a packet rewritten by that rule...

Alternatively there could be different rules for sdn nodes and ovn-k nodes.

Also, why do you need the `resubmit`? If the answer is "because the rewriting doesn't take effect otherwise", then please put that in the comment.

pkg/network/node/ovscontroller.go
```diff
@@ -267,6 +268,17 @@ func (oc *ovsController) SetupOVS(clusterNetworkCIDR []string, serviceNetworkCID
...
+	// set the dst MAC for traffic from OVN-K nodes to local pods
+	otx.AddFlow("table=10, priority=200, ip, nw_dst=%s, dl_dst=%s, actions=move:nw_dst->dl_dst[0..31], set_field:0a:58:00:00:00:00/ff:ff:00:00:00:00->eth_dst, resubmit:10", localSubnetCIDR, oc.localMAC)
```
danwinship, line 281: use `eth_dst` everywhere rather than saying `dl_dst` (which is an alias) in some places.

(also, same question on the `resubmit`)

pkg/network/node/ovscontroller.go
```diff
@@ -923,3 +935,17 @@ func (oc *ovsController) SetNamespaceEgressViaEgressIPs(vnid uint32, egressIPsMe
 	}
 	return otx.Commit()
 }
+
+func ensureLocalMAC(localIP string) string {
```
danwinship, line 939: We already check the primary NIC in `pkg/network/node.go:validateNodeIP()`, which calls `GetLinkDetails()` but ignores the result other than whether it returns an error. But you could just capture the MAC there as well.

pkg/network/node/ovscontroller.go
```diff
@@ -267,6 +268,17 @@ func (oc *ovsController) SetupOVS(clusterNetworkCIDR []string, serviceNetworkCID
 
 	// Table 10: VXLAN ingress filtering; filled in by AddHostSubnetRules()
 	// eg, "table=10, priority=100, tun_src=${remote_node_ip}, actions=goto_table:30"
+	// Swap overlay dst MAC for ingress OVN-K hybrid overlay traffic
+	if oc.localMAC == "" {
+		return fmt.Errorf("failed to find the MAC address for local IP %s", oc.localIP)
+	}
+	if err := oc.ensureTunMAC(); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
pliurh: fixed

pkg/network/node/ovscontroller.go
pliurh: fixed

pkg/network/node/ovscontroller.go
```diff
@@ -267,6 +268,17 @@ func (oc *ovsController) SetupOVS(clusterNetworkCIDR []string, serviceNetworkCID
 
 	// Table 10: VXLAN ingress filtering; filled in by AddHostSubnetRules()
 	// eg, "table=10, priority=100, tun_src=${remote_node_ip}, actions=goto_table:30"
+	// Swap overlay dst MAC for ingress OVN-K hybrid overlay traffic
```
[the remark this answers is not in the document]
pliurh: fixed

pkg/network/node/ovscontroller.go
pliurh: This rule matches both dst IP and MAC. So, only the packets with dst MAC equal to the node interface MAC will hit this rule. Such ingress traffic can only come from OVN-K nodes. The traffic you mention will use the MAC of the `tun0` interface so that it won't hit this rule.

danwinship:
/lgtm
can merge after 4.15 opens

PR #576, opened 2023-09-19
Collect pod operation latency metrics properly

pperiyasamy:
This fixes pod_operations_latency metric for labels setup and teardown as it incorrectly calculating duration while invoking metric's observe method. It also adds new label called e2esetup in metric which provides latency info about total setup processing time for pods.

### DIFF
--- a/pkg/network/common/cniserver/cniserver.go
+++ b/pkg/network/common/cniserver/cniserver.go
@@ -9,8 +9,10 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/gorilla/mux"
+	"github.com/openshift/sdn/pkg/network/node/metrics"
 	"k8s.io/klog/v2"
 
 	utilwait "k8s.io/apimachinery/pkg/util/wait"
@@ -288,6 +290,7 @@ func cniRequestToPodRequest(r *http.Request) (*PodRequest, error) {
 // Dispatch a pod request to the request handler and return the result to the
 // CNI server client
 func (s *CNIServer) handleCNIRequest(w http.ResponseWriter, r *http.Request) {
+	start := time.Now()
 	req, err := cniRequestToPodRequest(r)
 	if err != nil {
 		http.Error(w, fmt.Sprintf("%v", err), http.StatusBadRequest)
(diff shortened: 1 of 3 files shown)


pkg/network/node/pod.go
```diff
@@ -483,7 +483,8 @@ func podIsExited(p *kcontainer.Pod) bool {
 
 // Set up all networking (host/container veth, OVS flows, IPAM, loopback, etc)
 func (m *podManager) setup(req *cniserver.PodRequest) (cnitypes.Result, *runningPod, error) {
-	defer metrics.PodOperationsLatency.WithLabelValues(metrics.PodOperationSetup).Observe(metrics.SinceInMicroseconds(time.Now()))
+	start := time.Now()
+	defer metrics.PodOperationsLatency.WithLabelValues(metrics.PodOperationSetup).Observe(metrics.SinceInMicroseconds(start))
```
kyrtapz, line 487: The deferred function's arguments are evaluated immediately, so I think this is still wrong since `metrics.SinceInMicroseconds(start)` will likely return ~0.

pkg/network/common/cniserver/cniserver.go
```diff
@@ -293,6 +295,12 @@ func (s *CNIServer) handleCNIRequest(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, fmt.Sprintf("%v", err), http.StatusBadRequest)
 		return
 	}
+	if req.Command == CNI_ADD {
```
dcbw, line 298: @pperiyasamy I would set the start time for *all* CNI operations and then at the end of the function use the req.Command to determine which metric to set. We should have both ADD & DEL metrics for e2e latency.

pkg/network/node/metrics/metrics.go
```diff
@@ -34,6 +34,7 @@ const (
 	OVSOperationFailure = "failure"
 	// Pod Operation types
 	PodOperationSetup    = "setup"
+	PodOperationE2ESetup = "e2esetup"
```
dcbw, line 37: @pperiyasamy Let's do an e2eteardown too

pkg/network/node/pod.go
pperiyasamy: yes @kyrtapz , thanks. now doing it with an inline function.

pkg/network/common/cniserver/cniserver.go
pperiyasamy: sure @dcbw, now adding metric labels for both delete as well as update (though update doesn't happen from CNI).

dcbw:
@pperiyasamy One last thing: need to record metrics in `update()` since you added a metric for it now.

PR #578, opened 2023-10-04
OCPBUGS-18785: Controller: add required flag for node name

martinkennelly:
Currently, we are retrieving the node/host name using the 'os' pkg. This works fine the vast majority of the time.

If we get the hostname using the 'os' pkg and if a nodes hostname is not correctly set for whatever reason, one or more nodes, may end up with the same hostname (e.g. localhost.localdomain).

This can cause multiple sdn controllers to be active because we use the hostname as the ID for the leader SDN controller.

PR https://github.com/openshift/cluster-network-operator/pull/2047 needs to get in first to update the sdn controller manifests and pass the flag to the controller binary as we now depend on it being set. 

/hold


martinkennelly:
job [test e2e-serial 4.14,openshift/sdn#578,openshift/cluster-network-operator#2047 aws](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-aws-modern/1709934056583467008) succeeded

job [test e2e-serial 4.14,openshift/sdn#578,openshift/cluster-network-operator#2047 aws](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-aws-modern/1709933978598772736) succeeded

PR #579, opened 2023-10-16
OCPBUGS-16788: Bump containernetworking/plugins

ricky-rav:
The purpose of this bump is to conform with CIS benchmarksChange CISv1.4 pdf ("1.1.9 Ensure that the Container Network Interface file permissions are set to 600 or more restrictive"). 

The permissions of files in the `/var/lib/cni/networks/openshift-sdn` folder for all sdn pods should be set to 600.


pkg/cmd/openshift-sdn-cni/openshift-sdn.go
```diff
@@ -304,6 +304,13 @@ func (p *cniPlugin) CmdDel(args *skel.CmdArgs) error {
 	return err
 }
 
+func (p *cniPlugin) CmdCheck(args *skel.CmdArgs) error {
```
ricky-rav, line 307: Here I included the same behaviour (and comment) we have in ovnk:
https://github.com/ovn-org/ovn-kubernetes/blob/5a08eb28c1618ce9748073b42bd9ef6c795e7a98/go-controller/pkg/cni/cni.go#L241

danwinship:
There's a bunch of new code here that might be doing subtly different things than the old code did...

I hate to suggest this after you did all this work, but given that openshift-sdn is deprecated, and that rebasing to current CNI does not get us any customer-requested improvements, I think it would be simpler to just backport the fix we need to the version of CNI we import here, apply it to `vendor/` directly, and run `make update-deps-overrides` to record the local patch.

ricky-rav:
Closing this in favor of https://github.com/openshift/sdn/pull/584

PR #580, opened 2023-10-25
SDN-4123: rebase sdn to kube 1.28.3

jluhrsen:


pkg/cmd/openshift-sdn-node/proxy.go
```diff
@@ -23,7 +24,7 @@ import (
 func readProxyConfig(filename string) (*kubeproxyconfig.KubeProxyConfiguration, error) {
 	o := kubeproxyoptions.NewOptions()
 	o.ConfigFile = filename
-	if err := o.Complete(); err != nil {
+	if err := o.Complete(new(pflag.FlagSet)); err != nil {
```
danwinship, line 27: you need to pass down `cmd.Flags()` from above

jluhrsen: let me know if what I did is not right

jluhrsen:
> OK, end result is good, but the path there is wrong. Your first commit doesn't actually do anything; it's still vendoring the same 1.27 k8s code, it's just telling go.mod to _pretend_ that it's 1.28.
> 
> Basically all the `go.mod`, `go.sum` and `vendor/` changes should be in the first commit, and then the second commit would be just the code fixups. I'm not sure which commit the "go 1.19" bump makes more sense in... could go either way.

Gotcha, there is one problem where I can't get vendor/ files to be updated without first fixing the conntrack import,
so I had to include that in the first commit to revendor things, but I think it's close to what you are saying now.


> But actually, I think 1.19 is wrong anyway... `.ci-operator.yaml` (which is auto-updated by ART) says to use golang 1.20.

i used 1.20 now.
 
> I'm going to merge the o/k PR now so you should be able to rebase with those fixes and using the real o/k branch in go.mod.

thank you

go.mod
```diff
@@ -1,104 +1,106 @@
...
-	k8s.io/kubernetes v1.27.1 // actually branch sdn-4.14-kubernetes-1.27.1 in openshift/kubernetes
-	k8s.io/utils v0.0.0-20230209194617-a36077c30491
+	k8s.io/cri-api v1.28.3
+	k8s.io/klog/v2 v2.100.1
+	k8s.io/kubernetes v1.28.3 // actual branch sdn-4.15-kubernetes-1.28.3 in openshift/kubernetes
+	k8s.io/utils v0.0.0-20230726121419-3b25d923346b
 	sigs.k8s.io/yaml v1.3.0
 )
 
+require github.com/spf13/pflag v1.0.5
```
danwinship, line 35: you ought to be able to put that above with all the others

jluhrsen: done

danwinship:
/lgtm
/retest-required
/retitle SDN-4123: rebase sdn to kube 1.28.3

PR #588, opened 2023-11-20
OCPBUGS-22517: CVE 2023 45142 openshift-sdn: opentelemetry: DoS vulnerability in otelhttp [openshift-4]

flavio-fernandes:
Upgrading **go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp** from **v0.35.1** to **v0.46.1**


danwinship:
>     $ go mod why go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp
>     github.com/openshift/sdn/pkg/network/node
>     k8s.io/kubernetes/pkg/kubelet/cri/remote
>     k8s.io/component-base/tracing
>     go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp

The bug is in the server-side otelhttp code, but in this chain of events, sdn is only creating a client. Thus it is not vulnerable to the CVE. Thus, we should not merge this very large PR which has a very large number of changes we couldn't possibly hope to review.

PR #598, opened 2023-12-19
NO-JIRA: Move sdn to RHEL9 base image

jcaamano:


images/sdn/Dockerfile.rhel
```diff
@@ -10,19 +10,19 @@ COPY . .
...
-
-RUN mkdir -p /opt/cni/bin/rhel9
-COPY --from=rhel9-builder /go/src/github.com/openshift/sdn/openshift-sdn-cni /opt/cni/bin/rhel9/openshift-sdn
+ARG ovsver=3.1
 
 RUN mkdir -p /opt/cni/bin/rhel8
-COPY --from=rhel8-builder /go/src/github.com/openshift/sdn/openshift-sdn-node /usr/bin/
-COPY --from=rhel8-builder /go/src/github.com/openshift/sdn/openshift-sdn-controller /usr/bin/
-COPY --from=rhel8-builder /go/src/github.com/openshift/sdn/openshift-sdn-cni /opt/cni/bin/openshift-sdn
 COPY --from=rhel8-builder /go/src/github.com/openshift/sdn/openshift-sdn-cni /opt/cni/bin/rhel8/openshift-sdn
```
danwinship, line 18: If this image is going to support both RHEL 8 and RHEL 9, then I think it needs to continue using RHEL 8 (and, eg, OVS 2.13) as its base.

If we're not supporting RHEL8 in 4.16 then you can just remove the RHEL8 build stuff.

jcaamano: @danwinship The idea of moving to RHEL9 is that kube-proxy image is built on that as base. So the ART automation gets confused and moves back & forth
https://github.com/openshift/sdn/pull/599
https://github.com/openshift/sdn/pull/597

danwinship: Well, then maybe the kube-proxy image needs to be built against RHEL 8 too?

To be clear, I don't know if we actually need any RHEL 8 support in this image any more. I'm just saying, if we _do_ need to support RHEL 8, then we need to use a RHEL 8 base, and if we _don't_ need to support RHEL 8, then you should remove all the RHEL 8 bits.

jcaamano: Yeah, I ask ART about the two things.

jcaamano: @danwinship BTW, in OVNK we moved forward to RHEL9 based images while still supporting RHEL8 with a pretty similar situation in regards to OVS. What would be the specific concern here?

danwinship: The specific concern is that in the past, we have not guaranteed that you can use a RHEL N ovs/iptables/etc client against a RHEL N-1 kernel. E.g., the RHEL 9 OVS client may be built in a way such that it assumes the kernel supports some API that wasn't actually there in RHEL 8. There was a whole thing about this when OCP went from RHEL 7 to RHEL 8.

Maybe the OVS team has said that RHEL 9 clients on a RHEL 8 kernel _is_ supported; I don't know...

jcaamano:
Holding off moving to RHEL9 until we have more feedback form OVS team

jcaamano:
According to OVS team we can use RHEL9 ovs 3.X ofctl/vcctl clients against RHEL 8 ovs 2.X servers.

jcaamano:
This PR sync the release repo https://github.com/openshift/release/pull/48077

danwinship:
/retitle NO-JIRA: Move sdn to RHEL9 base image

jcaamano:
/retitle OCPBUGS-28579: Move sdn to RHEL9 base image

PR #604, opened 2024-01-12
OCPBUGS-23666: Add Snyk file to exclude directories on scan

pperiyasamy:
This commits makes .synk file to exclude vendor and images directories from vulnerability scan.


danwinship:
> This PR creates .synk file

It should be `.snyk` shouldn't it?

pperiyasamy:
> > This PR creates .synk file
> 
> It should be `.snyk` shouldn't it?

yes @danwinship , you are right. corrected it. I've also updated the .snyk file based on the discussion [here](https://redhat-internal.slack.com/archives/C06A6KFFBRQ/p1705580799265369) with security team.

danwinship:
ok, I have no idea what this does but
/lgtm

PR #608, opened 2024-01-25
Remove .ci-operator.yaml

jcaamano:
The .ci-operator.yaml to customize the build of the root CI pipeline image. ART automation keeps trying to update this file between RHEL 8 & 9 just because we use those different versions to build the sdn and kube proxy images. AFAIK we don't use the root image in SDN CI so just remove it to avoid the hassle.

Also remove deps.diff which looks like was merged unintentionally.

### DIFF
--- a/.ci-operator.yaml
+++ b/.ci-operator.yaml
@@ -1,4 +1,0 @@
-build_root_image:
-  name: release
-  namespace: openshift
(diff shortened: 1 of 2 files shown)


jcaamano:
So it looks like ci-operator requires a root image even if it is not used https://github.com/openshift/release/pull/48063

PR #618, opened 2024-02-02
SDN-4404: Kube 1.29 bump

bpickard22:
Bumps k8s deps to 1.29

WIP: Replacing old / deprecated code


pkg/cmd/openshift-sdn-node/kube_proxy.go
```diff
@@ -142,6 +142,7 @@ func newProxyServer(config *kubeproxyconfig.KubeProxyConfiguration, client clien
 			recorder,
 			healthzServer,
 			config.NodePortAddresses,
+			false, //@DanWinship can you take a look here to make sure im using the right value for initOnly
```
danwinship, line 145: yes.
(If we weren't about to kill openshift-sdn off, then we could take advantage of the new `--init-only` feature, but it's not worth it at this point.)

pkg/cmd/openshift-sdn-node/proxy.go
```diff
@@ -135,6 +136,10 @@ func (d *sdnLocalDetector) IfNotLocal() []string {
 	return []string{"!", "-i", sdnnode.Tun0}
 }
 
+// func (d *sdnLocalDetector) IfLocalNFT() []string {
+//	return []string{}
+//}
```
danwinship, line 141: yeah, you'll need to add `IfLocalNFT` and `IfNotLocalNFT` to obey the interface but they can just return `nil` because they'll never get called

pkg/network/node/node.go
```diff
@@ -439,7 +439,7 @@ func (node *OsdnNode) killFailedPods(failed map[string]*kruntimeapi.PodSandbox)
 // See https://github.com/containernetworking/cni/issues/89
 func (node *OsdnNode) UpdatePod(pod corev1.Pod) error {
 	filter := &kruntimeapi.PodSandboxFilter{
-		LabelSelector: map[string]string{ktypes.KubernetesPodUIDLabel: string(pod.UID)},
+		LabelSelector: map[string]string{ktypes.KubernetesPodUIDLabel: string(pod.UID)}, //this const was removed from ktypes and am unsure what to replace it with
```
danwinship, line 442: Possibly something in `kruntimeapi` since this is a `kruntimeapi.PodSandboxFilter`?

One trick that is sometimes useful is to check out a copy of the N-1 version of kubernetes/kubernetes, and find places using that symbol, and then check out the current version of kubernetes, and see what that code is doing now. (_Hopefully_ there's just a different define, but possibly there was some sort of larger API change...)

vendor/github.com/containernetworking/plugins/plugins/ipam/host-local/backend/disk/backend.go
```diff
@@ -56,7 +56,7 @@ func New(network, dataDir string) (*Store, error) {
 
 func (s *Store) Reserve(id string, ip net.IP, rangeID string) (bool, error) {
 	fname := filepath.Join(s.dataDir, ip.String())
-	f, err := os.OpenFile(fname, os.O_RDWR|os.O_EXCL|os.O_CREATE, 0600)
+	f, err := os.OpenFile(fname, os.O_RDWR|os.O_EXCL|os.O_CREATE, 0644)
```
danwinship, line 59: note that you're accidentally reverting the local patch to this file (which will then cause `verify-deps` to fail)

danwinship (commented):
(you should have your fixes in a separate commit from the rebase itself, even though that means the rebase commit won't compile)

pkg/network/node/node.go
bpickard22: yeah was definetly part of a larger api change as the SwapBehavior types (where that value was a part of) no longer exists (seems like what happened was the label keys that were being used have replaced the old swap behavior types) 

we should be able to replace this with the string value of the old constant right?

vendor/github.com/containernetworking/plugins/plugins/ipam/host-local/backend/disk/backend.go
bpickard22: hmm wonder how that happened, should i just ignore this change?

pkg/network/node/node.go
danwinship: Well, the problem is that if the behavior on the other side of the API has changed, then the old constant value may no longer have the effect we wanted it to have.

What we're trying to do here, overall, is get the sandboxID of a particular pod, which we are currently identifying by UID. If we can't fix the `PodSandboxFilter` to do what we want, you could also just modify `getPodSandboxID` to not use a filter, and scan the returned list of pods itself. (This function is the only caller of `getPodSandboxID`.)

vendor/github.com/containernetworking/plugins/plugins/ipam/host-local/backend/disk/backend.go
danwinship: Just make sure to undo this change (`git checkout vendor/github.com/containernetworking/plugins/plugins/ipam/host-local/backend/disk/backend.go`) after running `go mod vendor`. (Or I think if you use `make update-deps` to update the vendoring it will re-apply the patch for you?)

(What's going on here is that we needed to fix a CVE, but actually updating to the version of `containernetworking/plugins` that had the fix would have pulled in dozens of other dep updates, etc. So we just patched the fix into the vendor tree by hand, resulting in a vendor tree that no longer actually matches what `go.sum` thinks it should, but `make verify-deps` has a hack to deal with this (`./deps.diff`).)

danwinship:
This looks ready to merge once https://github.com/openshift/kubernetes/pull/1877 is ready

danwinship:
Once this passes CI with the updated vendoring I'll merge the o/k commit

bpickard22:
> Once this passes CI with the updated vendoring I'll merge the o/k commit

fixing the vendoring now

go.mod
```diff
@@ -175,34 +177,34 @@ replace (
...
+	k8s.io/csi-translation-lib => github.com/openshift/kubernetes/staging/src/k8s.io/csi-translation-lib v0.0.0-20240104165947-948d02fb3a9b
+	k8s.io/kms => github.com/openshift/kubernetes/staging/src/k8s.io/kms v0.0.0-20240104165947-948d02fb3a9b
+	k8s.io/kube-aggregator => github.com/openshift/kubernetes/staging/src/k8s.io/kube-aggregator v0.0.0-20240104165947-948d02fb3a9b
+	k8s.io/kube-controller-manager => github.com/openshift/kubernetes/staging/src/k8s.io/kube-controller-manager v0.0.0-20240104165947-948d02fb3a9b
+	k8s.io/kube-proxy => github.com/openshift/kubernetes/staging/src/k8s.io/kube-proxy v0.0.0-20240104165947-948d02fb3a9b
+	k8s.io/kube-scheduler => github.com/openshift/kubernetes/staging/src/k8s.io/kube-scheduler v0.0.0-20240104165947-948d02fb3a9b
+	k8s.io/kubectl => github.com/openshift/kubernetes/staging/src/k8s.io/kubectl v0.0.0-20240104165947-948d02fb3a9b
+	k8s.io/kubelet => github.com/openshift/kubernetes/staging/src/k8s.io/kubelet v0.0.0-20240104165947-948d02fb3a9b
+	k8s.io/kubernetes => github.com/bpickard22/kubernetes v1.24.0-alpha.3.0.20240201203439-45de81136a16
```
danwinship, line 201: `45de81136a16` is the `sdn-4.16-kubernetes-1.29.0` branch _before_ your PR. You need to use `9d8de2b40ed`

go.mod
```diff
@@ -175,34 +177,34 @@ replace (
...
+	k8s.io/csi-translation-lib => github.com/openshift/kubernetes/staging/src/k8s.io/csi-translation-lib v0.0.0-20240104165947-948d02fb3a9b
+	k8s.io/kms => github.com/openshift/kubernetes/staging/src/k8s.io/kms v0.0.0-20240104165947-948d02fb3a9b
+	k8s.io/kube-aggregator => github.com/openshift/kubernetes/staging/src/k8s.io/kube-aggregator v0.0.0-20240104165947-948d02fb3a9b
+	k8s.io/kube-controller-manager => github.com/openshift/kubernetes/staging/src/k8s.io/kube-controller-manager v0.0.0-20240104165947-948d02fb3a9b
+	k8s.io/kube-proxy => github.com/openshift/kubernetes/staging/src/k8s.io/kube-proxy v0.0.0-20240104165947-948d02fb3a9b
+	k8s.io/kube-scheduler => github.com/openshift/kubernetes/staging/src/k8s.io/kube-scheduler v0.0.0-20240104165947-948d02fb3a9b
+	k8s.io/kubectl => github.com/openshift/kubernetes/staging/src/k8s.io/kubectl v0.0.0-20240104165947-948d02fb3a9b
+	k8s.io/kubelet => github.com/openshift/kubernetes/staging/src/k8s.io/kubelet v0.0.0-20240104165947-948d02fb3a9b
+	k8s.io/kubernetes => github.com/bpickard22/kubernetes v1.24.0-alpha.3.0.20240417213555-9d8de2b40ed0
```
danwinship, line 201: for e2e purposes that's fine, but in the final commit you should update all of the other k8s.io rewrites to the same SHA too

deps.diff
```diff
@@ -1,9 +0,0 @@
-diff --no-dereference -N -r current/vendor/github.com/containernetworking/plugins/plugins/ipam/host-local/backend/disk/backend.go updated/vendor/github.com/containernetworking/plugins/plugins/ipam/host-local/backend/disk/backend.go
-59c59
-< 	f, err := os.OpenFile(fname, os.O_RDWR|os.O_EXCL|os.O_CREATE, 0600)
----
-> 	f, err := os.OpenFile(fname, os.O_RDWR|os.O_EXCL|os.O_CREATE, 0644)
-77c77
-< 	err = ioutil.WriteFile(ipfile, []byte(ip.String()), 0600)
----
-> 	err = ioutil.WriteFile(ipfile, []byte(ip.String()), 0644)
```
danwinship, line 9: This file got deleted; it's needed to preserve a local patch for a CVE. I think if you use `make update-deps` rather than `go mod vendor` it will do the right magic automatically. (Do that after un-deleting this file.)

danwinship:
merged the kubernetes PR

PR #628, opened 2024-06-04
OCPBUGS-20736: update x/net to v0.17.0

JacobTanenbaum:
update x/net to v0.17.0 to fix  CVE-2023-39325

in order to update the package the go version also needed to be updated


JacobTanenbaum:
is not required

PR #630, opened 2024-07-25
SDN-4925: Revendor k/k to upstream 1.30.3 (for kube-proxy)

danwinship:
This will let us build kube-proxy 1.30. It breaks the sdn build, but we shouldn't be building that any more anyway.


danwinship:
/override ci/prow/unit
/override ci/prow/verify

not expected to pass because they only check the sdn code, which no longer compiles... if we were going to keep building kube-proxy from here I'd fix this, but the plan is to migrate it somewhere else for 4.18...

PR #635, opened 2024-09-20
OCPBUGS-43344: OCPBUGS-42244: Implement namespaceSelectors on ingress with all pods IPs

pliurh:
During SDN live migration, the VNID from ovn-k nodes cannot be used distinguish namespaces. So generate per-pod-IP match rules instead.

### DIFF
--- a/pkg/network/node/networkpolicy.go
+++ b/pkg/network/node/networkpolicy.go
@@ -3,6 +3,8 @@ package node
 import (
 	"context"
 	"fmt"
+	"math"
+	"os"
 	"reflect"
 	"sort"
 	"strings"
@@ -35,6 +37,8 @@ const HostNetworkNamespace = "openshift-host-network"
 // only used by network policy functions
 const NetPolIsolationCookie = 1147582955
 
+const migrationEnvVar = "NODE_CNI"
+
 type networkPolicyPlugin struct {
 	node   *OsdnNode
 	vnids  *nodeVNIDMap
@@ -63,6 +67,9 @@ type networkPolicyPlugin struct {
 	// into unProcessedVNIDs in EnsureVNIDRules() method and it can be looked up later in
 	// AddNetNamespace() method so that required nw policy flow rules programmed into OVS.
 	unProcessedVNIDs sets.Int32
+
+	// indicate if the node is running in SDN live migration mode
+	inMigrationMode bool
 }
 
 // npNamespace tracks NetworkPolicy-related data for a Namespace
@@ -116,16 +123,17 @@ const (
 )
 
 func NewNetworkPolicyPlugin() osdnPolicy {
+	_, inMigration := os.LookupEnv(migrationEnvVar)
 	return &networkPolicyPlugin{
 		namespaces:       make(map[uint32]*npNamespace),
 		namespacesByName: make(map[string]*npNamespace),
 
 		nsMatchCache: make(map[string]*npCacheEntry),
 
-		warnedPolicies:  make(map[ktypes.UID]string),
-		skippedPolicies: make(map[ktypes.UID]string),
-		localPodIPs:     make(map[ktypes.UID]string),
-
+		warnedPolicies:   make(map[ktypes.UID]string),
+		skippedPolicies:  make(map[ktypes.UID]string),
+		localPodIPs:      make(map[ktypes.UID]string),
+		inMigrationMode:  inMigration,
 		unProcessedVNIDs: sets.NewInt32(),
 	}
 }
@@ -775,6 +783,12 @@ func (np *networkPolicyPlugin) parsePeerFlows(npns *npNamespace, npp *npPolicy, 
 			if dir == ingressFlow && (len(peer.PodSelector.MatchLabels) == 0 && len(peer.PodSelector.MatchExpressions) == 0) {
 				// The PodSelector is empty, meaning it selects all pods in this namespace
 				peerFlows = append(peerFlows, fmt.Sprintf("reg0=%d, ", npns.vnid))
+				if np.inMigrationMode {
+					npp.watchesOwnPods = true
+					for _, ip := range np.selectPods(npns, peer.PodSelector) {
+						peerFlows = append(peerFlows, fmt.Sprintf("ip, nw_src=%s, ", ip))
+					}
+				}
 			} else {
 				npp.watchesOwnPods = true
 				for _, ip := range np.selectPods(npns, peer.PodSelector) {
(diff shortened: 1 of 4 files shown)


pliurh:
@danwinship PTAL

pkg/network/node/networkpolicy.go
```diff
@@ -789,15 +769,15 @@ func (np *networkPolicyPlugin) parsePeerFlows(npns *npNamespace, npp *npPolicy,
 			if len(peer.NamespaceSelector.MatchLabels) == 0 && len(peer.NamespaceSelector.MatchExpressions) == 0 {
 				// The NamespaceSelector is empty, meaning it selects all namespaces
 				peerFlows = append(peerFlows, "")
-			} else if dir == ingressFlow {
-				// We can implement namespaceSelectors on ingress by just
-				// checking the source VNID...
-				npp.watchesNamespaces = true
-				peerFlows = append(peerFlows, np.selectNamespaces(peer.NamespaceSelector, dir)...)
```
danwinship, line 796: Do we not have some way of distinguishing that we are in "migration mode"?

It's _much_ less efficient to use per-IP matches rather than VNID matches. This will break some very large clusters.

martinkennelly: @danwinship even if we can detect migration and only implement this per IP matches during that time period, do you also foresee issues with large clusters? 
I guess so..

pliurh: > Do we not have some way of distinguishing that we are in "migration mode"?

No. We could add an environment variable for that if needed.

But currently per-IP is also used by egress policies, would it get this scale problem?

danwinship: > even if we can detect migration and only implement this per IP matches during that time period, do you also foresee issues with large clusters?

Yes. In theory, some large clusters may have NP-related performance problems during migration that they did not have before migration.

I don't have any good estimates of how large a cluster needs to be to hit this problem though; SDN has always had this particular optimization, so we don't have any evidence of how different customers' clusters would behave without it...

> But currently per-IP is also used by egress policies, would it get this scale problem?

Yes, but egress policies have always worked that way, so people would have already avoided creating egress policies that caused problems in their cluster.

pliurh:
Relies on https://github.com/openshift/cluster-network-operator/pull/2532

pkg/network/node/networkpolicy.go
pliurh: I updated the patch. Now, the per-podIP rules will be added after the VNI rules and only when running in migration mode. Also, I take the host-network namespace into account. PTAL

pkg/network/node/networkpolicy.go
```diff
@@ -35,6 +37,8 @@ const HostNetworkNamespace = "openshift-host-network"
 // only used by network policy functions
 const NetPolIsolationCookie = 1147582955
 
+const migrtionEnvVar = "NODE_CNI"
```
danwinship:
```suggestion
const migrationEnvVar = "NODE_CNI"
```

pkg/network/node/networkpolicy.go
```diff
@@ -794,6 +798,35 @@ func (np *networkPolicyPlugin) parsePeerFlows(npns *npNamespace, npp *npPolicy,
 				// checking the source VNID...
 				npp.watchesNamespaces = true
 				peerFlows = append(peerFlows, np.selectNamespaces(peer.NamespaceSelector, dir)...)
+				if _, ok := os.LookupEnv(migrtionEnvVar); ok {
```
danwinship: do the `LookupEnv` in `NewNetworkPolicyPlugin` and stash the result in a field on `networkPolicyPlugin`

pkg/network/node/networkpolicy.go
```diff
@@ -794,6 +798,35 @@ func (np *networkPolicyPlugin) parsePeerFlows(npns *npNamespace, npp *npPolicy,
...
+					// In SDN live migration mode, we can't use source VNID to
+					// distinguish namespaces for traffic from OVN nodes.
+					// So instead we pretend the rule was a combined
+					// namespaceSelector+podSelector rule with a match-all
+					// podSelector, and generate per-pod-IP match rules.
+					npp.watchesAllPods = true
+					peerFlows = append(peerFlows, np.selectPodsFromNamespaces(peer.NamespaceSelector, &metav1.LabelSelector{}, dir)...)
+
+					// If the host network namespace is selected, Add rules for
+					// the OVN mp0 IP of each node.
```
danwinship, line 821: SDN treats all non-pod-cidr traffic as "host network", so we just need to accept anything that's not in `npp.node.clusterCIDRs`.

The "not" makes this difficult though... the easiest way to deal with this would probably be to update [this rule](https://github.com/openshift/sdn/blob/d515d15883394318af2c55e75e17291a65d023b2/pkg/network/node/ovscontroller.go#L258) to add `load:1->NXM_NX_REG3[]` to the actions, to tag all nw_src=clusterCIDR with reg3=1 (doesn't need to be conditional on migration). Then here you can just check `reg3=0` here to match everything that didn't match that (ie, non-clusterCIDR-source traffic).

Oh! And either way, you need to bump `ruleVersion` in ovscontroller.go (which will require updates to the unit tests as well).

pliurh: But the IP of the mp0 interface is in the range of the clusterCIDR. It's the second IP of the node subnet.

pliurh: I added unit test and bumped the `ruleVersion`. PTAL

deps.diff
```diff
@@ -1,9 +0,0 @@
-diff --no-dereference -N -r current/vendor/github.com/containernetworking/plugins/plugins/ipam/host-local/backend/disk/backend.go updated/vendor/github.com/containernetworking/plugins/plugins/ipam/host-local/backend/disk/backend.go
```
danwinship: We're carrying a patch against upstream `containernetworking/plugins` (or at least, against the version of that module that we're vendoring). So this file needs to be kept, and you need to manually undo the changes that "go mod vendor" tries to make to `vendor/github.com/containernetworking/plugins/`.

(Maybe I should hack this to work differently because it's constantly a pain to deal with...)

pkg/network/node/networkpolicy.go
```diff
@@ -794,6 +802,35 @@ func (np *networkPolicyPlugin) parsePeerFlows(npns *npNamespace, npp *npPolicy,
...
+					sel, _ := metav1.LabelSelectorAsSelector(peer.NamespaceSelector)
+					if _, ok := np.nsMatchCache[sel.String()].matches[HostNetworkNamespace]; ok {
+						hsns, err := np.node.osdnInformers.Network().V1().HostSubnets().Lister().List(labels.Everything())
```
danwinship: ok, since this is only for migration mode, and we shouldn't be adding or removing nodes while migrating, you could just precompute the list of mp0-related flows in `Start()` rather than regenerating it every time.

Alternatively... OVS lets you match IPs against _arbitrary_ bitmasks, not just proper CIDR expressions. If `node.networkInfo.ClusterNetworks` is `[{ ClusterCIDR: "10.128.0.0/14", HostSubnetLength: 23 }]`, then the mp0 IPs are exactly the ones that have "`10.128`" in bits 0-14 and "`2`" in bits 23-32 (and anything in bits 15-22). So that's `nw_src=10.128.0.2/0xfffc1f` (where that mask value is `uint32(math.MaxUint32 << (32 - cidrMaskLen)) | (math.MaxUint32 >> hostSubnetLen)`)

Right?

pliurh: Done. PTAL

PR #637, opened 2024-10-23
OCPBUGS-43193: CVE-2024-21626: Bump runc from v1.1.10 to v1.1.12

k37y:
- Bump `github.com/opencontainers/runc` from `v1.1.10` to `v1.1.12`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/fsnotify/fsnotify v1.7.0
 	github.com/gorilla/mux v1.8.0
 	github.com/miekg/dns v1.1.35
-	github.com/opencontainers/runc v1.1.10
+	github.com/opencontainers/runc v1.1.12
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -8,7 +8,7 @@ require (
 	github.com/fsnotify/fsnotify v1.7.0
 	github.com/gorilla/mux v1.8.0
 	github.com/miekg/dns v1.1.35
-	github.com/opencontainers/runc v1.1.10
+	github.com/opencontainers/runc v1.1.12
```
danwinship, line 11: note that this has no effect because of a `replace` rule later in the file:

```
github.com/opencontainers/runc => github.com/opencontainers/runc v1.0.0-rc95
```

Removing that to allow the dep to be updated results in a very large change, which we don't want in openshift-sdn unless absolutely necessary.

In this case, sdn isn't using the functions that the CVE covers anyway, so there's no need to update it.

PR #638, opened 2024-10-23
OCPBUGS-43764: Revendor to a patched k/k with our prefer-local-DNS hack

danwinship:
Apparently IBM was relying on this behavior in the standalone kube-proxy image.

(This also updates from k8s v1.30.3 to v1.30.5, which pulls in some irrelevant changes, and probably I should have just stuck with 1.30.3 to keep it simple but whatever. Anyway, the relevant change is in `vendor/k8s.io/kubernetes/pkg/proxy/iptables/proxier.go`.)

The `replace` stuff in `go.mod` is just taken from the 4.16 branch. I had simplified things when rebasing to 1.30 for 4.17 because we weren't going to be keeping local patches any more, but now we are again, so...


danwinship:
/override ci/prow/unit
/override ci/prow/verify

openshift/sdn CI is broken on master

PR #643, opened 2024-12-09
OCPBUGS-45806: Stop checking ruleVersion [4.16]

danwinship:
At some point things seem to have broken such that doing a "full setup" ends up breaking the node until you restart all the pods.

Since all upgradeable-from versions have basically equivalent OVS flows, just remove the ruleVersion check, so that we never have to do a "full setup" for that reason.

In particular, although I thought we needed a bump from ruleVersion 13 to 14 for the OVN migration fix, it was actually unnecessary; we already rewrite all NetworkPolicy rules on restart anyway (and even if we didn't, we need to rewrite them when you switch between regular mode and migration mode, not when you first upgrade to the release that supports migration).

/assign @pliurh

### DIFF
--- a/pkg/network/node/ovscontroller.go
+++ b/pkg/network/node/ovscontroller.go
@@ -182,9 +182,6 @@ const (
 	Tun0   = "tun0"
 	Vxlan0 = "vxlan0"
 
-	// rule versioning; increment each time flow rules change
-	ruleVersion = 14
-
 	ruleVersionTable = 253
 )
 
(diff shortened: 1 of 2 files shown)


pliurh:
During live migration, nodes won't be restarted when we redeploy SDN in the migration mode. However, the mtu migration followed by that will reboot all the nodes. So we should be fine without the ruleVersion.

/lgtm

danwinship:
> During live migration, nodes won't be restarted when we redeploy SDN in the migration mode. However, the mtu migration followed by that will reboot all the nodes. So we should be fine without the ruleVersion.

We should be fine even without a reboot: it's only the NetworkPolicy-related rules that need to be rewritten when you start a migration, but that happens any time you restart the openshift-sdn process anyway. I was just wrong before about bumping the ruleVersion.

PR #647, opened 2024-12-19
OCPBUGS-46536: Bump openvswitch

k37y:
- Bump openvswitch from `2.13` to `2.17` to mitigate CVE-2019-25076

### DIFF
--- a/images/sdn/Dockerfile.rhel
+++ b/images/sdn/Dockerfile.rhel
@@ -13,7 +13,7 @@ RUN CGO_ENABLED=1 make build GO_BUILD_PACKAGES="github.com/openshift/sdn/cmd/ope
 FROM registry.ci.openshift.org/ocp/4.14:cli AS cli
 FROM registry.ci.openshift.org/ocp/4.14:base
 
-ARG ovsver=2.13
+ARG ovsver=2.17
 
 RUN mkdir -p /opt/cni/bin/rhel9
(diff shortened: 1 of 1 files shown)


danwinship:
> Bump openvswitch from `2.13` to `2.17` to mitigate [CVE-2019-25076](https://github.com/advisories/GHSA-8ch8-jcmf-3gj4?rgh-link-date=2024-12-19T13%3A51%3A51Z)

Just to be clear: this is not actually about mitigating the CVE; the CVE was mitigated years ago by updating the openvswitch RPM in the RHCOS image (and it would only have been exploitable under ovn-kubernetes anyway). This change is just about preventing the sdn image from being flagged by automated CVE checkers.

danwinship:
To fix the `jira/invalid-bug` label, you need to do what it says [in this comment](https://github.com/openshift/sdn/pull/647#issuecomment-2575035889).

For the release note, pick "Rebase" for the type, and for the release note text, just note that we updated the image from ovs 2.13 to 2.17 for maintenance reasons but we do not expect any behavioral changes as a result.

For the "dependent bug", you can use OCPBUGS-28580, which is the bug that changed the OVS RPM version in 4.15. (So on 46536, scroll to "Issue Links", click the "+" icon on the right, then choose "depends on" and enter OCPBUGS-28580.)

PR #648, opened 2024-12-29
OCPBUGS-44674: Handle `openshift-host-network` namespace as special when it modifies

k37y:
- Add condition in `NetworkPolicy` `updateMatchCache` to handle `HostNetworkNamespace` changes

### DIFF
--- a/pkg/network/node/networkpolicy.go
+++ b/pkg/network/node/networkpolicy.go
@@ -616,16 +616,9 @@ func (np *networkPolicyPlugin) selectNamespacesInternal(selector labels.Selector
 	match := np.nsMatchCache[cacheKey]
 	if match == nil {
 		match = &npCacheEntry{selector: selector, matches: make(map[string]uint32)}
-		for vnid, npns := range np.namespaces {
+		for _, npns := range np.namespaces {
 			if npns.gotNamespace && selector.Matches(labels.Set(npns.labels)) {
-				// handle host network namespace as special and classify it as vnid 0 for
-				// network policy purposes, so it can ride upon the handling of default
-				// namespace for host network traffic.
-				if npns.name == HostNetworkNamespace {
-					match.matches[npns.name] = 0
-				} else {
-					match.matches[npns.name] = vnid
-				}
+				match.matches[npns.name] = npns.GetMatchVNID()
 			}
 		}
 		np.nsMatchCache[cacheKey] = match
@@ -636,7 +629,7 @@ func (np *networkPolicyPlugin) selectNamespacesInternal(selector labels.Selector
 func (np *networkPolicyPlugin) updateMatchCache(npns *npNamespace) {
 	for _, match := range np.nsMatchCache {
 		if npns.gotNamespace && npns.gotNetNamespace && match.selector.Matches(labels.Set(npns.labels)) {
-			match.matches[npns.name] = npns.vnid
+			match.matches[npns.name] = npns.GetMatchVNID()
 		} else {
 			delete(match.matches, npns.name)
 		}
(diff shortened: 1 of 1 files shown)


pkg/network/node/networkpolicy.go
```diff
@@ -614,7 +614,11 @@ func (np *networkPolicyPlugin) selectNamespacesInternal(selector labels.Selector
 func (np *networkPolicyPlugin) updateMatchCache(npns *npNamespace) {
 	for _, match := range np.nsMatchCache {
 		if npns.gotNamespace && npns.gotNetNamespace && match.selector.Matches(labels.Set(npns.labels)) {
-			match.matches[npns.name] = npns.vnid
+			if npns.name == HostNetworkNamespace {
+				match.matches[npns.name] = 0
+			} else {
+				match.matches[npns.name] = npns.vnid
+			}
```
danwinship: Ah! Yes, this seems right, but can you turn this into a helper function so we don't have the same logic in both `selectNamespacesInternal` and `updateMatchCache`? (Maybe `npns.GetMatchVNID()`?)

danwinship:
openshift-sdn master is dead now (and in particular, there is no CI for openshift-sdn master any more). SDN bugs should be filed against 4.16 first, and then backported from there as needed. You can click the "Edit" button by the bug summary, and then that will let you change the base branch from `openshift:master` to `openshift:release-4.16`. (And you'll need to update the jira issue as well. You'll also need to create a dummy 4.17 version of the jira issue to make the 4.16 one "depends on", in order to get the openshift bots to be happy.)

pkg/network/node/networkpolicy.go
k37y: Thanks, @danwinship. The suggested changes have been added.

k37y:
@danwinship, I have made the below changes to the Jira:

Target release: `4.16.0`
Release note: Handle the openshift-host-network namespace as special and set the VNID as zero when it modifies.
Release note type: `Bug Fix`

And cloned `4.17.0` targeted Jira and closed it as `Done`.

pkg/network/node/networkpolicy.go
```diff
@@ -1190,3 +1186,10 @@ func (np *networkPolicyPlugin) refreshPodNetworkPolicies(pod *corev1.Pod) bool {
 func getPodFullName(pod *corev1.Pod) string {
 	return fmt.Sprintf("%s/%s", pod.Namespace, pod.Name)
 }
+
+func (npns *npNamespace) GetMatchVNID(vnid uint32) uint32 {
+	if npns.name == HostNetworkNamespace {
+		return 0
+	}
+	return vnid
```
danwinship: you can just return `npns.vnid` rather than taking a `vnid` argument. (And then just update the loop in `selectNamespacesInternal` to be `for _, npns := range np.namespaces`

pkg/network/node/networkpolicy.go
```diff
@@ -621,11 +621,7 @@ func (np *networkPolicyPlugin) selectNamespacesInternal(selector labels.Selector
 				// handle host network namespace as special and classify it as vnid 0 for
 				// network policy purposes, so it can ride upon the handling of default
 				// namespace for host network traffic.
-				if npns.name == HostNetworkNamespace {
-					match.matches[npns.name] = 0
-				} else {
-					match.matches[npns.name] = vnid
-				}
+				match.matches[npns.name] = npns.GetMatchVNID(vnid)
```
danwinship: You should move the comment over to the `GetMatchVNID` function too

danwinship (commented):
just a few minor things

pkg/network/node/networkpolicy.go
k37y: I was thinking about that, but then I decided not to touch anything else. Sure, I will add that change.

PR #649, opened 2025-02-05
OCPBUGS-46493: Raise the minSyncPeriod for NetworkPolicies to 10s during migration

danwinship:
Normally in SDN, NetworkPolicies that match an entire namespace are implemented by matching the VNID, but this doesn't work during sdn->ovn migration, so they get rewritten to use pod IPs instead. The code for doing this is not really optimized, and particularly during migration (when every pod in the cluster gets deleted from its original node and then recreated on another) it results in the NP code doing a huge amount of work (and, seemingly, generating new garbage faster than the garbage collector can collect it).

This tries to mitigate the problem by only doing resyncs once every 10 seconds in migration mode (as opposed to once a second in non-migration mode). Hopefully this is enough to get the existing customers through migration. (If this does work I have other ideas, but they require rewriting a large chunk of code.)

(See also #650 for a more complicated, but maybe better, approach.)

/assign @pliurh

### DIFF
--- a/pkg/network/node/networkpolicy.go
+++ b/pkg/network/node/networkpolicy.go
@@ -189,10 +189,23 @@ func (np *networkPolicyPlugin) Start(node *OsdnNode) error {
 		return err
 	}
 
-	// Rate-limit calls to np.syncFlows to 1-per-second after the 2nd call within 1
-	// second. The maxInterval (time.Hour) is irrelevant here because we always call
-	// np.runner.Run() if there is syncing to be done.
-	np.runner = async.NewBoundedFrequencyRunner("NetworkPolicy", np.syncFlows, time.Second, time.Hour, 2)
+	// Rate-limit calls to np.syncFlows to once per minSyncPeriod. The maxInterval
+	// (time.Hour) is irrelevant here because we always call np.runner.Run() if there
+	// is syncing to be done.
+	minSyncPeriod := time.Second
+	if np.inMigrationMode {
+		// The existing code can't keep up with the churn rate in large clusters
+		// when in migration mode.
+		allSubnets, err := np.node.osdnClient.NetworkV1().HostSubnets().List(context.TODO(), metav1.ListOptions{})
+		if err != nil {
+			klog.Warningf("Unexpected error trying to list HostSubnets (%v); assuming large cluster", err)
+			minSyncPeriod = 10 * time.Second
+		} else if len(allSubnets.Items) > 100 {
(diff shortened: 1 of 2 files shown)


pkg/network/node/networkpolicy.go
```diff
@@ -189,10 +189,16 @@ func (np *networkPolicyPlugin) Start(node *OsdnNode) error {
...
-	// np.runner.Run() if there is syncing to be done.
-	np.runner = async.NewBoundedFrequencyRunner("NetworkPolicy", np.syncFlows, time.Second, time.Hour, 2)
+	// Rate-limit calls to np.syncFlows to once per minSyncPeriod. The maxInterval
+	// (time.Hour) is irrelevant here because we always call np.runner.Run() if there
+	// is syncing to be done.
+	minSyncPeriod := time.Second
+	if np.inMigrationMode {
+		// The existing code can't keep up with the churn rate in large clusters
+		// when in migration mode.
+		minSyncPeriod = 10 * time.Second
```
npinaeva: is there a chance of wrongly allowing or denying existing connections for 10s because of this change?

danwinship: So of course, in the affected clusters, this is basically already happening because of slow syncs.

If someone modifies the NetworkPolicies or Pod/Namespace labels, then it will take us up to 10s to catch up to that modification. But there is no reason why someone would need to do that in the middle of a migration, so this doesn't seem like a huge problem.

It does mean that pods will be fully-isolated for up to 10s on startup. It would be good to avoid that, but the goal of this particular PR was to be simple... (I couldn't even come up with a good way to make the behavior be conditional on "it's a big cluster", because this code runs before the informer caches have fully synced, and there's no way to change the parameters of the BoundedFrequencyRunner after creating it.)

npinaeva: My concern is that smaller clusters that didn't have this extra delay will now get it, trying to figure out what the result of this could be.
Pods that are migrating to the new node will be ignored for 10s I guess, so if a migrated pod tries to immediately reach a non-migrated pod after its startup, this could fail?

danwinship: yes, though presumably it will retry. But you'll get additional delay, and spurious error messages in your pod logs.

danwinship:
OK, we tried to test this with random minSyncPeriod, but it's hard to test because only _some_ of the pods in any run will have the memory usage problems (and it takes a _really_ long time to do a test run). So rather than delay this further, we're just going to stick with 10s, which we know works. (Though in the latest version, the 10s now only applies to clusters with at least 100 nodes; smaller clusters still use the old value.)

PR #650, opened 2025-02-05
OCPBUGS-46493: Optimize namespaceSelector NetworkPolicies under migration mode

danwinship:
Normally in SDN, NetworkPolicies that match an entire namespace are implemented by matching the VNID, but this doesn't work during sdn->ovn migration, so they get rewritten to use pod IPs instead. The code for doing this is not really optimized, and particularly during migration (when every pod in the cluster gets deleted from its original node and then recreated on another) it results in the NP code doing a huge amount of work (and, seemingly, generating new garbage faster than the garbage collector can collect it).

In particular, currently we end up recalculating all "whole namespace" policies whenever any pod in any namespace changes. This changes things so that we only trigger recalculation when a pod changes in a namespace whose namespaceSelector actually matches that pod.

I think this will miss updates if labels on Namespaces change, but that's fine; the new code is only used in migration mode, and we can't just document that you're not allowed to change the labels on Namespaces during migration.

(Alternative to #649.)

/assign @pliurh

### DIFF
--- a/pkg/network/node/networkpolicy.go
+++ b/pkg/network/node/networkpolicy.go
@@ -99,6 +99,7 @@ type npPolicy struct {
 	watchesNamespaces bool
 	watchesAllPods    bool
 	watchesOwnPods    bool
+	watchesSomePods   sets.Set[uint32]
 
 	selectedIPs   []string
 	selectsAllIPs bool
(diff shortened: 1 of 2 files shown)


pkg/network/node/networkpolicy.go
```diff
@@ -1162,7 +1163,7 @@ func (np *networkPolicyPlugin) refreshPodNetworkPolicies(pod *corev1.Pod) bool {
 	podNs := np.namespacesByName[pod.Namespace]
 	for _, npns := range np.namespaces {
 		for _, npp := range npns.policies {
-			if (npp.watchesOwnPods && npns == podNs) || npp.watchesAllPods {
+			if (npp.watchesOwnPods && npns == podNs) || npp.watchesSomePods.Has(npns.vnid) || npp.watchesAllPods {
```
npinaeva: I may not fully understand these structs, but shouldn't this condition be
`npp.watchesSomePods.Has(podNs.vnid)`?

danwinship: ah, yes, good catch...

I guess the unit test didn't catch this because this code only runs when a pod is added or deleted, but we mostly only do that at the start of the test...

danwinship:
fixed the bug you pointed out... I tried to add another unit test to catch it, but it still gets the right answer either way because of the way the policies are set up in the test. :slightly_frowning_face: . Anyway, doesn't seem worth it to try to tweak the test further.

pliurh:
/lgtm

Do we need to adopt both this and #649? Or this one itself is sufficient for the bug?

danwinship:
Either one is hopefully sufficient but neither is tested yet... we're waiting on QE

danwinship:
OK, based on @liqcui 's testing, this PR _does not_ improve memory usage, but #649 does, so we'll go with that.

